PR #20, opened 2021-09-24
UX cleanup: quality of life improvements

mvaivre:
Quality of life update:

- Dark theme
- Smooth route transitions
- Better error handling (import words)
- Fix display of TX sent to self
- Splash screen
- Show version number in setting
- Show current account in sidebar
- Various UI improvements (amounts etc.)
- Set global content-security-policy


src/components/Modal.tsx
```diff
@@ -30,14 +30,23 @@ export const Modal: React.FC<{ title: string; onClose: () => void; focusMode?: b
...
-    // Prevent body scroll on mount
     document.body.style.overflow = 'hidden'
     return () => {
       document.body.style.overflow = 'unset'
     }
   }, [])
 
+  // Handle escape key press
+  useEffect(() => {
+    document.addEventListener('keydown', onClose, false)
```
tdroxler, line 43: don't you need to filter the `escape` key?

tdroxler (approved):
Reviewed and manually tested all features, LGTM :nail_care:

polarker:
6. I don't see the log during the process of `Import Wallet`, but I could see them on the loading page. Not a big deal though

7. yes, should be able to delete without password, since one could delete the file anyway without a password, even if it's inside leveldb.

PR #39, opened 2021-11-01
Words check: alphabetically ordered list

mvaivre:
Makes the process of checking the words a bit less painful, while still forcing the user to take enough time to do it properly.

### DIFF
--- a/src/pages/WalletManagement/CheckWordsPage.tsx
+++ b/src/pages/WalletManagement/CheckWordsPage.tsx
@@ -23,16 +23,7 @@ import { AlertTriangle, ThumbsUp } from 'lucide-react'
 const Storage = getStorage()
 
 const getAlphabeticallyOrderedList = (arr: string[]) => {
-  const newArr = arr.slice()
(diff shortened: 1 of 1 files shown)


src/pages/WalletManagement/CheckWordsPage.tsx
```diff
@@ -22,13 +22,17 @@ import { AlertTriangle, ThumbsUp } from 'lucide-react'
...
 
-const getShuffledArr = (arr: string[]) => {
+const getAlphabeticallyOrderedList = (arr: string[]) => {
   const newArr = arr.slice()
-  for (let i = newArr.length - 1; i > 0; i--) {
-    const rand = Math.floor(Math.random() * (i + 1))
-    ;[newArr[i], newArr[rand]] = [newArr[rand], newArr[i]]
-  }
-  return newArr
+  return newArr.sort(function (a, b) {
```
polarker, line 27: It seems `newArr.sort()` is enough ?

polarker (approved):
A minor comment, otherwise LGTM

PR #46, opened 2021-11-02
Network switch and network configuration presets

mvaivre:
Introduces a network switch, along with testnet and mainnet presets configurations:

- The new network switch now allows "normal" users to easily switch from testnet and mainnet networks, avoid mistakes.
- Advanced users can manually override those presets and set any endpoints they want (for local dev etc.)

A new `ExpandableSection` component has been created to hide secondary / advanced information.


mvaivre:
This will need to be updated once the mainnet URLs are known.

PR #47, opened 2021-11-05
fix: set the app window icon on Linux

0xjac:
Electron should use the executable icon as the default application
window. However this seems broken on Linux. This might be related to
https://github.com/alephium/alephium-wallet/pull/41

This sets the alephium icon for the application to be displayed by the
wm on the window decoration, on windows switchers, etc.

Notice the difference before (yellow line) and after on rofi
2021-11-05-170151_3840x2081_scrot
:


polarker (approved):
Thanks for the fix ❤️ Bounty added.

PR #55, opened 2021-11-22
Fix local node connection issues + better errors

### LINKED ISSUE #53: Customized nodeHost does not work

mvaivre:
- Auto append http protocol to custom node URL
- Return human-readable errors
- Don't throw error on selfClique call issues (previous pseudo health-check)

Fixes #53

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -74,27 +74,19 @@ const App = () => {
   // Create client
   useEffect(() => {
     const getClient = async () => {
-      try {
-        setClientIsLoading(true)
-        // Get clients
-        const clientResp = await createClient(settings)
-        setClient(clientResp)
+      setClientIsLoading(true)
+      // Get clients
+      const clientResp = await createClient(settings)
+      setClient(clientResp)
 
-        console.log('Clients initialized.')
+      console.log('Clients initialized.')
 
-        setSnackbarMessage({
-          text: `Connected to Alephium's Node "${settings.nodeHost}"!`,
-          type: 'info',
-          duration: 4000
-        })
-        setClientIsLoading(false)
-      } catch (e) {
-        setSnackbarMessage({
-          text: 'Unable to initialize the client, please check your network settings.',
-          type: 'alert'
-        })
-        setClientIsLoading(false)
-      }
+      setSnackbarMessage({
+        text: `Alephium's Node URL: "${settings.nodeHost}"`,
+        type: 'info',
+        duration: 4000
+      })
+      setClientIsLoading(false)
 
       // Save settings
       saveSettings(settings)
--- a/src/pages/SettingsPage.tsx
+++ b/src/pages/SettingsPage.tsx
@@ -203,8 +203,8 @@ const ClientSettings = () => {
   const editSettings = (v: Partial<Settings>) => {
     const newSettings = { ...tempSettings, ...v }
 
-    // Check if we need to append the http:// protocol if an IP is used
-    if (v.nodeHost?.match(/^(?:[0-9]{1,3}\.){3}[0-9]{1,3}[:[0-9]*]?$/)) {
+    // Check if we need to append the http:// protocol if an IP or localhost is used
+    if (v.nodeHost?.match(/^(?:(?:[0-9]{1,3}\.){3}[0-9]{1,3}|localhost)(?::[0-9]*)?$/)) {
       newSettings.nodeHost = `http://${v.nodeHost}`
     }
 
--- a/src/pages/SettingsPage.tsx
+++ b/src/pages/SettingsPage.tsx
@@ -203,6 +203,11 @@ const ClientSettings = () => {
   const editSettings = (v: Partial<Settings>) => {
     const newSettings = { ...tempSettings, ...v }
 
+    // Check if we need to append the http:// protocol if an IP is used
+    if (v.nodeHost?.match(/^(?:[0-9]{1,3}\.){3}[0-9]{1,3}[:[0-9]*]?$/)) {
+      newSettings.nodeHost = `http://${v.nodeHost}`
+    }
+
     overrideSelectionIfMatchesPreset(newSettings)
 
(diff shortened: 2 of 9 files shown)


polarker:
@mvaivre did you test it with host or ip different than `127.0.0.1` and `localhost` ?

mvaivre:
> @mvaivre did you test it with host or ip different than `127.0.0.1` and `localhost` ?

Yes I did. But for now I auto-append "http://" only in front of IPs (IPv4), to avoid miss-configurations. 
Should we do the same for `localhost`?

We'll test the full stack tomorrow morning with @tdroxler.

mvaivre:
Note: @tdroxler has CORS issues when communicating with the local node when not using `localhost` or `127.0.0.1`. This would need to be investigated.

src/App.tsx
```diff
@@ -74,27 +74,19 @@ const App = () => {
...
-          type: 'alert'
-        })
-        setClientIsLoading(false)
-      }
+      setClientIsLoading(true)
+      // Get clients
+      const clientResp = await createClient(settings)
+      setClient(clientResp)
+
+      console.log('Clients initialized.')
```
nop33, line 82: Why plural, since it's one client?

src/pages/SettingsPage.tsx
```diff
@@ -203,6 +203,11 @@ const ClientSettings = () => {
   const editSettings = (v: Partial<Settings>) => {
     const newSettings = { ...tempSettings, ...v }
 
+    // Check if we need to append the http:// protocol if an IP or is used
```
nop33, line 206: > if an IP or is used

❓

src/utils/api.ts
```diff
@@ -1,7 +1,24 @@
...
+
+  if (isHTTPError(e)) {
+    returnedError = e.error.detail
+  } else if (stringifiedError) {
+    returnedError = stringifiedError
+  } else {
+    returnedError = defaultErrorMessage
+  }
+
+  return returnedError
```
nop33, line 23: We could simplify the code to this:

```js
export const getHumanReadableError = (e: unknown, defaultErrorMessage: string) => {
  const stringifiedError = tryGetString(e)

  if (isHTTPError(e)) {
    return e.error.detail
  } else if (stringifiedError) {
    return stringifiedError
  }

  return defaultErrorMessage
}
```

or if you feel fancy, this:

```js
export const getHumanReadableError = (e: unknown, defaultErrorMessage: string) => {
  const stringifiedError = tryGetString(e)
  
  return isHTTPError(e) ? e.error.detail : stringifiedError || defaultErrorMessage
}
```

No strong preference, I leave it up to you :)

src/utils/types.ts
```diff
@@ -0,0 +1,5 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const tryGetString = (obj: any): string | null => {
+  if (typeof obj.toString === 'function') return obj.toString()
+  else return null
+}
```
nop33, line 5: I personally find the following version simpler, but again, no strong opinion here:

```js
export const tryGetString = (obj: any): string | null => (
  typeof obj.toString === 'function' ? obj.toString() : null
)
```

nop33 (changes requested):

src/utils/api.ts
mvaivre: Let's go with the ternary version (I hesitated, went for the explicit version, but let's change my mind ;)). I'll make the change.

src/utils/types.ts
nop33: Yeah, it takes a while to get used to the "no return" statements. Maybe this is simpler then:

```js
export const tryGetString = (obj: any): string | null => {
  return typeof obj.toString === 'function' ? obj.toString() : null
}
```

but anyway, let's not waste more time on this :)

src/App.tsx
mvaivre: The main backend "client" is actually an aggregation of the cliqueClient (node) and the ExplorerClient (explorer-backend).

src/utils/types.ts
mvaivre: It's good to agree on common patterns. Why not go for your last proposition which seems like a good compromise

src/pages/SettingsPage.tsx
mvaivre: "or localhost". Fixed.

src/App.tsx
```diff
@@ -74,27 +74,19 @@ const App = () => {
...
-
-        console.log('Clients initialized.')
-
-        setSnackbarMessage({
-          text: `Connected to Alephium's Node "${settings.nodeHost}"!`,
-          type: 'info',
-          duration: 4000
-        })
-        setClientIsLoading(false)
-      } catch (e) {
```
nop33, line 91: How do we handle wrong network settings now that you removed the `catch`?

mvaivre: It's the outcome of the whole discussion we had on Slack [here](https://alephium.slack.com/archives/C026MHDJD2N/p1637071135044300) and [here](https://alephium.slack.com/archives/C026MHDJD2N/p1637139773058300) and described in the linked task [here](https://github.com/alephium/alephium-wallet/issues/53#issuecomment-971402782).

Returning selfClique errors could be misleading (it has been, for some users), as we previously considered it as a pseudo-health check. 

- For now, we decided to display an error if an TX cannot be send successfully, indicating a bad configuration or network issue.
- When the user update the settings, we simply print a "Settings saved" message.

(BTW, this shows that brainstorming on Slack is probably not the best solution, as the detailed outcomes are easily lost.)

mvaivre:
@nop33 is your change request still on? 
If not, I believe we could merge this one and go ahead a release a new version.

src/App.tsx
nop33: sorry, completely forgot about this. Resolving.

nop33 (approved):

PR #63, opened 2021-12-01
Use softprops' release action & generate checksums in the release

capito27:
Update the github action for release to use softprops' release action and compute checksums for every binary. 
Sample release available here : https://github.com/capito27/alephium-wallet/releases/tag/v9.9.12

NB : the worflow had to be slighly redesigned to work with softprops' action


polarker (approved):
LGTM

PR #65, opened 2021-12-02
Use genesis timestamp to identify and label genesis TXs

### LINKED ISSUE #54: Genesis token allocation are marked as `Mining Reward`
We could check the timestamp when there's no inputs and mark something else or maybe check height == 0.

mvaivre:
Following @tdroxler proposal, we use the genesis timestamp to correctly label genesis TXs.

Fixes #54

### DIFF
--- a/src/pages/Wallet/WalletHomePage.tsx
+++ b/src/pages/Wallet/WalletHomePage.tsx
@@ -199,25 +199,23 @@ const IOList = ({
   isOut: boolean
   transaction: Transaction
 }) => {
-  const io = (isOut ? transaction.outputs : transaction.inputs) as Array<Output | Input>
+  const io = (isOut ? transaction.outputs : transaction.inputs) as Array<Output | Input> | undefined
   const genesisTimestamp = 1231006505000
 
   if (io && io.length > 0) {
-    if (io.every((o) => o.address === currentAddress)) {
-      return <Address key={currentAddress} hash={currentAddress} />
-    } else {
-      return (
-        <>
-          {_(io.filter((o) => o.address !== currentAddress))
-            .map((v) => v.address)
-            .uniq()
-            .value()
-            .map((v) => (
-              <Address key={v} hash={v || ''} />
-            ))}
-        </>
-      )
-    }
+    return io.every((o) => o.address === currentAddress) ? (
+      <Address key={currentAddress} hash={currentAddress} />
+    ) : (
+      <>
+        {_(io.filter((o) => o.address !== currentAddress))
+          .map((v) => v.address)
+          .uniq()
+          .value()
+          .map((v) => (
+            <Address key={v} hash={v || ''} />
+          ))}
+      </>
+    )
   } else if (transaction.timestamp === genesisTimestamp) {
     return <TXSpecialTypeLabel>Genesis TX</TXSpecialTypeLabel>
   } else {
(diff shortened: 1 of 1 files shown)


tdroxler (approved):
Idea LGTM, but let's wait @nop33 review for the typescript part :)

src/pages/Wallet/WalletHomePage.tsx
```diff
@@ -206,6 +190,41 @@ const WalletHomePage = () => {
...
+  transaction
+}: {
+  currentAddress: string
+  isOut: boolean
+  transaction: Transaction
+}) => {
+  const io = (isOut ? transaction.outputs : transaction.inputs) as Array<Output | Input>
+  const genesisTimestamp = 1231006505000
+
+  if (io && io.length > 0) {
```
nop33, line 205: Based on line `202`, the `io` variable is either an `Array` of `Output`s or an `Array` of `Input`s, or an empty `Array`. None of this values is a falsy value so I think we don't need to check `if (io && `. Is there a case where `transaction.outputs` AND `transaction.inputs` are [both undefined](https://github.com/alephium/alephium-js/blob/master/api/api-explorer.ts#L83-L84)?

src/pages/Wallet/WalletHomePage.tsx
```diff
@@ -206,6 +190,41 @@ const WalletHomePage = () => {
...
+            .map((v) => v.address)
+            .uniq()
+            .value()
+            .map((v) => (
+              <Address key={v} hash={v || ''} />
+            ))}
+        </>
+      )
+    }
+  } else if (transaction.timestamp === genesisTimestamp) {
```
nop33, line 221: The multiple `if` statements confused me a lot that I had to open the editor to realize that this code is still reachable. To improve code readability, I'd propose to replace these `if` sections with this:

```js
  if (io && io.length > 0) {
    return io.every((o) => o.address === currentAddress) ? (
      <Address key={currentAddress} hash={currentAddress} />
    ) : (
      <>
        {_(io.filter((o) => o.address !== currentAddress))
          .map((v) => v.address)
          .uniq()
          .value()
          .map((v) => (
            <Address key={v} hash={v || ''} />
          ))}
      </>
    )
  } else if (transaction.timestamp === genesisTimestamp) {
    return <TXSpecialTypeLabel>Genesis TX</TXSpecialTypeLabel>
  } else {
    return <TXSpecialTypeLabel>Mining Rewards</TXSpecialTypeLabel>
  }
```

Not a strong opinion, just an idea.

src/pages/Wallet/WalletHomePage.tsx
mvaivre: My casting is a bit "optimistic": if we look at our API typings, I should cast `as Array<Output | Input> | undefined` and keep the `io &&` part. @tdroxler Do you know a case where bot outputs and inputs would be undefined?

src/pages/Wallet/WalletHomePage.tsx
mvaivre: It's very similar (question of habits I guess), I'm fine with using the ternary version ;)

src/pages/Wallet/WalletHomePage.tsx
tdroxler: mmmm no I don't see it as a valid case BUT it could be that the explorer-backend wrongly sync and suddenly return no outputs nor inputs. So maybe it's safer to handle that case?
Now that both of you are type lovers: that's why some libraries are proposing some `NonEmptyArray` type, where you are 100% sure you get at least one element, but we don't have this for now.

mvaivre: Thanks @tdroxler! As we don't have this for now, I'm for keeping it safe and continue checking for `io` definition.

nop33 (approved):

PR #67, opened 2021-12-03
Aggregate pending TXs per network

### LINKED ISSUE #48: Aggregate pending TXs per network
Previous pending TX are still displayed if the user switch from one network to another.
Either find a better way to fetch pending queries (best option, check with @tdroxler) or store pending TXs per network config.

mvaivre:
Fixes #48 while we wait for a proper way to fetch TX in mempool (per address)

### DIFF
--- a/src/pages/Wallet/WalletHomePage.tsx
+++ b/src/pages/Wallet/WalletHomePage.tsx
@@ -22,7 +22,7 @@ import { useHistory } from 'react-router-dom'
 import { Input, Output, Transaction } from 'alephium-js/dist/api/api-explorer'
 import { Send, QrCode, RefreshCw, Lock, LucideProps, Settings as SettingsIcon } from 'lucide-react'
 import { abbreviateAmount, calAmountDelta } from '../../utils/numbers'
-import { loadSettingsOrDefault } from '../../utils/clients'
+import { loadSettingsOrDefault, useCurrentNetwork } from '../../utils/clients'
 import AmountBadge from '../../components/Badge'
 import _ from 'lodash'
 import dayjs from 'dayjs'
@@ -44,6 +44,7 @@ dayjs.extend(relativeTime)
 
 const WalletHomePage = () => {
   const history = useHistory()
+  const currentNetwork = useCurrentNetwork()
   const { wallet, setSnackbarMessage, client, setWallet, currentUsername } = useContext(GlobalContext)
   const [balance, setBalance] = useState<bigint | undefined>(undefined)
   const { pendingTxList, loadedTxList, setLoadedTxList } = useContext(WalletContext)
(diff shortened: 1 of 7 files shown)


src/pages/Wallet/WalletHomePage.tsx
```diff
@@ -130,17 +131,19 @@ const WalletHomePage = () => {
   }, [fetchData])
 
   // Polling (when pending tx)
-  useInterval(fetchData, 2000, pendingTxList.length === 0)
+  useInterval(fetchData, 2000, pendingTxList[currentNetwork]?.length === 0)
 
   if (!wallet) return null
 
+  const pendingTxLength = (pendingTxList[currentNetwork] || []).length
```
nop33, line 138: How about instead of `pendingTxLength`, we create `pendingTxs` so we don't have to rewrite it in line `187`?
```js
const pendingTxs = pendingTxList[currentNetwork] || []
```

Then, we can also create a variable:
```js
const showSpinner = isLoading || pendingTxs.length > 0
```

and use it in lines `146` and `184`.

nop33 (changes requested):
I went through the code quickly and this is my feedback. I need to checkout the branch and test it out to fully understand what's happening. I'll consult issue #48 as well.

src/pages/Wallet/WalletHomePage.tsx
mvaivre: Definitely. Thanks for your suggestion!

nop33 (commented):
On a second iteration over these changes I observed that the `pendingTxList` of the `WalletContext` used to hold **a list** of pending transactions. However, it now holds **multiple lists** of pending transactions, one for each network. To avoid ambiguity when reading the code for the first time (of after a long time) and comparing `pendingTxList` with `loadedTxList` (which is an array and not an object), I would rename `pendingTxList` to `networkPendingTxLists`.

It also improves readability when later on we spot `networkPendingTxLists[currentNetwork]` instead of `pendingTxList[currentNetwork]`.

Moreover, doesn't `loadedTxList` suffer from the same problem as `pendingTxList` when switching between networks?

src/pages/Wallet/WalletRootPage.tsx
```diff
@@ -16,36 +17,33 @@ export interface SimpleTx {
...
+  const [loadedTxList, setLoadedTxList] = useState<WalletContextType['loadedTxList']>([])
+  const [pendingTxList, setPendingTxList] = useState<WalletContextType['pendingTxList']>({})
 
   const addPendingTx = (tx: SimpleTx) => {
-    setPendingTxList((prev) => {
-      if (prev && tx) {
-        return [...prev, tx]
-      } else return prev
-    })
+    setPendingTxList((prev) => (tx ? { ...prev, [currentNetwork]: [...(prev[currentNetwork] || []), tx] } : prev))
```
nop33, line 46: Shoudn't we avoid calling `setPendingTxList` altogether if `tx` is falsy?

nop33:
@mvaivre not a very strong opinion regarding my comment above. Feel free to merge if you don't agree with it.

nop33 (approved):

mvaivre:
@nop33 Renaming done, rebased. 
And as for your question about the `loadedTxList`: no, in that case, a network switch triggers a new fetch which in turn updates the TX list.

PR #68, opened 2021-12-06
Cleanup (part 1)

nop33:
I will create multiple PRs for the cleanup phase, so that they are easier reviewable. This is the first part.


src/utils/bip39.ts
```diff
@@ -1,2 +1,18 @@
...
+//
+// The library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the library. If not, see <http://www.gnu.org/licenses/>.
+
 export const bip39Words =
```
polarker, line 17: Is it a good practice to store such a long string in memory ?

nop33: The string seems to be ~13KB
```
> new Blob([bip39words]).size
13115
```

or in Node:

```
> Buffer.byteLength(bip39words).length
13115
```

Not sure what other approach there is in order to pass this list of words into the `Tags` component of the `@yaireo/tagify/dist/react.tagify` library so that when importing a wallet, a list of suggested words pop up when the user types their words.

mvaivre:
Based on my previous review and the fact that you keep loading the font in the CSS to avoid the flickering: LGTM!

mvaivre (approved):

PR #70, opened 2021-12-07
Cleanup (part 2)

nop33:
- Update imports order 
  - Separating external from internal modules with an empty line so that it is easier to scan the imports to find a module. Modules are also grouped based on their type (components, context, utils, styles). Image imports are separated by an empty line.
- Simplify how we load settings
   - The default value can be returned from the settings loading function directly if no settings have been found in storage
- Separate settings to their own module


nop33:
@mvaivre can I merge this? I'm happy to rebase it and fix any conflicts after you merge https://github.com/alephium/alephium-wallet/pull/67

mvaivre (approved):

PR #78, opened 2021-12-14
Cleanup (part 5)

nop33:
Keeping code DRY, adding better error handling, improving readability.

Check commits to see an easier to understand overview of changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,16 @@ npm run electron-pack
 
 ## Development
 
+To launch the app as a web app, run:
+
 ```shell
 npm start
+```
+
+To launch it as an electron app, run:
+
+```shell
+npm run electron-dev
 ```
 
 ## Test
--- a/src/components/ActionButton.tsx
+++ b/src/components/ActionButton.tsx
@@ -1,0 +1,84 @@
+import styled, { useTheme } from 'styled-components'
+import { LucideProps } from 'lucide-react'
+import { useHistory } from 'react-router-dom'
+
+const ActionButton = ({
+  Icon,
+  label,
+  link,
+  onClick
+}: {
+  Icon: (props: LucideProps) => JSX.Element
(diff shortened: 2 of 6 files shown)


src/utils/settings.ts
```diff
@@ -37,7 +37,10 @@ export const networkEndpoints: Record<Exclude<NetworkType, 'custom'>, Settings>
   }
 }
 
-export const walletIdleForTooLongThreshold = 3 * 60 * 1000 // 3 minutes
+const lockWalletAfter = 3 * 60 * 1000 // 3 minutes
+const lockWalletAfterDev = 9999999999
```
mvaivre, line 41: In general, I like to have development conditions close to the final release version, so that I'm doing proper QA while developing. 
But if you find the auto-lock too annoying, we should add a test for it to make sure it's not broken and we don't see it.

nop33: Actually, every time I change something I need to refresh the app, so it locks me out anyway... I'll remove this commit.

nop33: done

nop33: but I will add tests eventually as well ;)

mvaivre: Yeah, the inability to interact with the app after a change is really annoying - this I agree. It's linked to the CSP, not allowing React to inject new code, even while developing.

If you are motivated, it would be very nice to find how to :
1) Either adapt the CSP to keep it safe but allow code injection for a local source (self).
2) Or remove the rule, but only in dev mode.

nop33: Yes, I noticed the CSP issue. For now I just added `unsafe-inline` without committing it. I'll look for a proper solution.

mvaivre: Until then, I think we can merge this one!

PR #82, opened 2021-12-15
Fix dev hot reloading issue with CSP

nop33:
Closes #81

## Build
```
$ npm run build
$ cat build/index.html
...
<meta http-equiv="Content-Security-Policy" content="script-src 'self'"/>
...
```

## Dev
```
npm electron-dev
```

image

PR #83, opened 2021-12-15
Revert "Fix dev hot reloading issue with CSP"

mvaivre:
I went ahead and merged #82 (couldn't wait to rebase on this one ;)), to then realise that I have the following error:

```The Content-Security-Policy directive name '%REACT_APP_CSP%' contains one or more invalid characters. Only ASCII alphanumeric characters or dashes '-' are allowed in directive names.```

We could either revert directly, or fix this in the PR. As you prefer @nop33

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "author": "Alephium dev <dev@alephium.org>",
   "main": "public/electron.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


mvaivre:
Issue seems to be gone after having emptied the cache. Please double check that everything looks good on your side @nop33 (both dev and release version). If everything's fine, we could close this one.

nop33:
I checked out `master`, built the macOS app, and launched it. All seem to be good.

I also run the electron dev server, no errors there either.

PR #85, opened 2021-12-16
Simplify Paragraph component

mvaivre:
A short cleaning PR. We can now use 1 `Paragraph` component with simple props (`centered`, `secondary`) instead of 3 different components.

Side-note: I cherry-picked a2323e25a0e6ded4a5fc6df435f2820a4129debf as the gas-price feature will have to wait for the OpenAPI to be updated (`gasPrice` format is incorrect.)


nop33 (approved):
image

PR #89, opened 2021-12-21
Refactor settings storage, improve settings modal and add customisable lock time

### LINKED ISSUE #87: Make wallet lock time configurable from the settings page
@polarker forwarded on Slack this request from a user who performs many transactions and the current (3min) lock time is to short for him resulting in frequent wallet locks.

mvaivre:
- Settings are now managed as a centralised / structured state.

```typescript {
{
  general: {
    theme: ThemeType
    walletLockTimeInMinutes: number | null
  }
  network: {
    nodeHost: string
    explorerApiHost: string
    explorerUrl: string
  }
}
```

- The settings modal is cleaner, better organised. 
- Auto-lock time can now be set manually (or deactivated)
- Cleanup: split inputs component, refine styling.

Fixes #87


src/components/Inputs/InlineLabelValueInput.tsx
```diff
@@ -0,0 +1,52 @@
+import { ReactNode } from 'react'
```
nop33, line 1: Missing comments block

src/components/Inputs/TextAreaTags.tsx
```diff
@@ -0,0 +1,102 @@
...
+  .tagify__input:empty::before {
+    // Placeholder
+    color: ${({ theme }) => theme.font.secondary};
+  }
+`
+
+const defaultStyle = (isValid?: boolean) => {
+  return css`
+    background-image: none;
+    height: 46px;
```
nop33, line 63: Let's use the `var(--inputHeight)` since we have it now.

src/components/Inputs/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+export const inputPlaceHolderVariants: Variants = {
+  up: { y: -35, x: -5, scale: 0.8 },
+  down: { y: 0, scale: 1 }
+}
+
+export const inputDefaultStyle = (isValid?: boolean) => {
+  return css`
+    background-image: none;
+    height: 46px;
```
nop33, line 41: Same here.

src/hooks/useIdleForTooLong.ts
```diff
@@ -14,13 +14,18 @@
 // You should have received a copy of the GNU Lesser General Public License
 // along with the library. If not, see <http://www.gnu.org/licenses/>.
 
-import { useEffect, useState, useCallback } from 'react'
+import { useEffect, useState, useCallback, useContext } from 'react'
+import { GlobalContext } from '../App'
```
nop33, line 18: Should move 1 line down

src/pages/Settings/GeneralSettingsSection.tsx
```diff
@@ -0,0 +1,42 @@
+import { useContext } from 'react'
+import { GlobalContext } from '../../App'
```
nop33, line 2: Needs empty line above

src/pages/Settings/GeneralSettingsSection.tsx
```diff
@@ -0,0 +1,42 @@
+import { useContext } from 'react'
```
nop33, line 1: Missing comments block

src/pages/Settings/GeneralSettingsSection.tsx
```diff
@@ -0,0 +1,42 @@
...
+      general: { walletLockTimeInMinutes }
+    },
+    updateSettings
+  } = useContext(GlobalContext)
+
+  return (
+    <>
+      <KeyValueInput
+        label="Lock time"
+        description="Duration in minutes after which the wallet will be locked automatically."
```
nop33, line 19: I think this is a bit misleading. We should clarify that the wallet will only lock if it is inactive/idle for X minutes, otherwise we risk people disabling it completely if they think that it will ALWAYS lock after X minutes. How about:

> Duration in minutes after which an idle wallet will lock automatically.

src/pages/Settings/GeneralSettingsSection.tsx
```diff
@@ -0,0 +1,42 @@
...
+            placeholder={walletLockTimeInMinutes ? 'Minutes' : 'Off'}
+            type="number"
+            step={1}
+            min={1}
+          />
+        }
+      />
+      <KeyValueInput
+        label="Theme"
+        description="Select the theme and please your eyes."
```
nop33, line 35: 😆 👍

nop33 (changes requested):
I went through the diff quickly and these are the things I noticed. I need to check out the code and do a proper functional test.

nop33:
Nice job with the visibility of the theme switch. Question: Wouldn't it be nice if we implement the same fade-in/out animation (background color and text color) that we have also in the explorer? Probably something like this should do it:

```css
* {
  transition: color 0.2s ease;
  transition: background-color 0.2s ease;
}
```

nop33 (approved):
I tested the settings and they seem to work, good job! 👏

src/components/Inputs/Input.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { motion } from 'framer-motion'
+import styled from 'styled-components'
+import classNames from 'classnames'
+import { useState } from 'react'
+import { Check } from 'lucide-react'
+
+import { sectionChildrenVariants } from '../PageComponents/PageContainers'
+import { InputProps, InputErrorMessage, inputPlaceHolderVariants, inputDefaultStyle } from './index'
+
+export const Input = ({ placeholder, error, isValid, disabled, onChange, value, ...props }: InputProps) => {
```
nop33, line 26: I think this should now be a default export.

nop33 (changes requested):

src/components/Inputs/Select.tsx
```diff
@@ -0,0 +1,184 @@
...
+  disabled?: boolean
+  controlledValue?: SelectOption<T>
+  options: SelectOption<T>[]
+  title?: string
+  id: string
+  onValueChange: (value: SelectOption<T> | undefined) => void
+  className?: string
+}
+
+export function Select<T>({
```
nop33, line 43: This should be a default export as well.

src/components/Inputs/TextArea.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+import { sectionChildrenVariants } from '../PageComponents/PageContainers'
+import { InputValidIconContainer, InputLabel, InputErrorMessage, TextAreaProps, inputDefaultStyle } from './index'
+
+const placeHolderVariants: Variants = {
+  up: { y: -35, x: -5, scale: 0.8 },
+  down: { y: 0, scale: 1 }
+}
+
+export const TextArea = ({ placeholder, error, isValid, disabled, onChange, value, ...props }: TextAreaProps) => {
```
nop33, line 31: Default export

src/components/Inputs/TextAreaTags.tsx
```diff
@@ -0,0 +1,69 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with the library. If not, see <http://www.gnu.org/licenses/>.
+
+import { motion } from 'framer-motion'
+import styled from 'styled-components'
+import Tags from '@yaireo/tagify/dist/react.tagify'
+
+import { inputDefaultStyle } from './index'
+
+export const TextAreaTags = (props: React.ComponentProps<typeof Tags>) => {
```
nop33, line 23: Default export

nop33 (changes requested):

src/hooks/useIdleForTooLong.ts
```diff
@@ -14,13 +14,18 @@
...
 
+import { GlobalContext } from '../App'
 import { useInterval } from '../utils/hooks'
-import { walletIdleForTooLongThreshold } from '../utils/settings'
 
 const useIdleForTooLong = (onIdleForTooLong: () => void) => {
   const [lastInteractionTime, setLastInteractionTime] = useState(new Date().getTime())
+  const {
+    settings: {
+      general: { walletLockTimeInMinutes }
```
nop33, line 26: Conceptually, this hook is agnostic regarding the action that will take place once the timeout is reached. We are passing a callback to it (that will lock the wallet) so that we don't need to pass the `wallet` variable and `setWallet` function as arguments. I don't like that we load the `walletLockTimeInMinutes` in this hook, cause now this hook is coupled with the "locking" feature. I would recommend that we pass a second argument to this hook, called `timeout`, and load the settings in the place where the hook is used.

What do you think?

mvaivre: Technically speaking, what I did here is simply a variable rename. The actual correlation is pretty much the same as before, except semantically now have the term "lock" in the new variable.

I didn't say anything about the hook during the first review - I'm not against it - but as of today I don't see a case where we would use it for something else than locking the wallet. I mean, we could do other actions when locking the wallet, but I guess that the whole logic associated to the idle threshold being reached should probably be written in a single place for clarity sake.

So anyway: TLDR, the coupling doesn't bother me, having a generic hook might be a tiny bit over engineered, but now that we it, we should keep it generic. I'll make the change.

nop33: I agree that a hook is probably overengineering, I just wanted to extract the logic out of the App component but neither a new component was an option (since I wasn't gonna render anything), nor a utility function (since I wanted to use state). Was there another alternative other than a hook?

mvaivre: Making it generic is the "overengineering" part, but I mean... not a big deal at all, it took me 10 secs.

The problem with hooks is the inability to make it conditional. In our case, if the user doesn't define a lock time, ideally we shouldn't execute everything that's in the hook, but it's not always easy nor optimal to do so. See https://github.com/alephium/alephium-wallet/pull/89/commits/1e6200064c2de1b62d0d462b3752678b79b6559e

It's easier to skip executing logic with a renderless component, but less clean / more weird IMO. So yeah. Don't really know a perfect way.

src/components/Inputs/TextAreaTags.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <StyledTags
+        {...props}
+        settings={{
+          enforceWhitelist: true,
+          delimiters: ' ',
+          maxTags: 24,
+          duplicates: true,
+          dropdown: {
+            enabled: 1, // show suggestion after 1 typed character
+            fuzzySearch: false, // match only suggestions that starts with the typed characters
```
nop33, line 37: > match only suggestions that ~starts~ start with the typed characters

😅

nop33 (approved):

src/components/Inputs/TextAreaTags.tsx
mvaivre: haha it's not even my typo. I stole it.

PR #95, opened 2021-12-23
Tests and code improvements

nop33:
- Added some more tests in the wallet page
- Added tests for number utilities
- Replaced QR code library with a better one
- Improved code readability
- Improved accuracy of qALPH calculation


mvaivre (approved):
LGTM

PR #96, opened 2021-12-24
Ability to set gas amount and gas price when creating a transaction + display expected fee

mvaivre:
**NOTE: This has been tested on the testnet directly, and gas settings seem not be taken into account (default value are shown in the TX). Need to do some E2E testing with the latest version of the node.**

- Add ability to set gas amount and gas price when creating a transaction (advanced settings)
- Display "expected fee" (should we call it "max fee"? @tdroxler @polarker)

### DIFF
--- a/src/pages/Wallet/SendPage.tsx
+++ b/src/pages/Wallet/SendPage.tsx
@@ -337,7 +337,7 @@ const onAmountInputValueChange = ({
 }
 
 const getExpectedFee = (gasAmount: string, gasPriceInALPH: string) => {
-  return abbreviateAmount(BigInt(gasAmount) * convertToQALPH(gasPriceInALPH), true)
+  return abbreviateAmount(BigInt(gasAmount) * convertToQALPH(gasPriceInALPH), false, 6)
 }
 
 const HeaderContent = styled(Section)`
--- a/src/pages/Wallet/SendPage.tsx
+++ b/src/pages/Wallet/SendPage.tsx
@@ -97,7 +97,7 @@ const SendPage = () => {
           fullAmount,
           undefined,
           parseInt(gasAmount),
-          gasPriceInALPH
+          convertToQALPH(gasPriceInALPH).toString()
         )
 
         const { txId, unsignedTx } = txCreateResp.data
--- a/src/pages/Wallet/SendPage.tsx
+++ b/src/pages/Wallet/SendPage.tsx
@@ -44,15 +44,13 @@ const SendPage = () => {
   const { client, wallet, setSnackbarMessage } = useGlobalContext()
   const { addPendingTx } = useTransactionsContext()
   const { setModalTitle, onModalClose, setOnModalClose } = useModalContext()
-
   const initialOnModalClose = useRef(onModalClose)
-  const minimalGasPriceInALPH = abbreviateAmount(MINIMAL_GAS_PRICE)
-
-  const [address, setAddress] = useState('')
-  const [amount, setAmount] = useState('')
-  const [gasAmount, setGasAmount] = useState<string>(MINIMAL_GAS_AMOUNT.toString())
-  const [gasPriceInALPH, setGasPriceInALPH] = useState<string>(minimalGasPriceInALPH)
-
+  const [transactionData, setTransactionData] = useState<TransactionData>({
+    address: '',
+    amount: '',
+    gasAmount: MINIMAL_GAS_AMOUNT.toString(),
+    gasPrice: abbreviateAmount(MINIMAL_GAS_PRICE)
+  })
   const [isSending, setIsSending] = useState(false)
   const [step, setStep] = useState<StepIndex>(1)
 
@@ -69,11 +67,8 @@ const SendPage = () => {
     }
   }, [setStep, setModalTitle, setOnModalClose, step])
 
-  const verifyTransactionContent = (address: string, amount: string, gasAmount: string, gasPrice: string) => {
-    setAddress(address)
-    setAmount(amount)
-    setGasAmount(gasAmount)
-    setGasPriceInALPH(gasPrice)
+  const verifyTransactionContent = (transactionData: TransactionData) => {
+    setTransactionData(transactionData)
     setStep(2)
   }
 
@@ -82,7 +77,8 @@ const SendPage = () => {
   }
 
   const handleSend = async () => {
-    const isDataComplete = address && amount && gasPriceInALPH && gasAmount
+    const { address, amount, gasAmount, gasPrice } = transactionData
+    const isDataComplete = address && amount && gasPrice && gasAmount
 
     if (wallet && client && isDataComplete) {
       setIsSending(true)
@@ -97,7 +93,7 @@ const SendPage = () => {
           fullAmount,
           undefined,
           parseInt(gasAmount),
-          convertToQALPH(gasPriceInALPH).toString()
+          convertToQALPH(gasPrice).toString()
         )
 
         const { txId, unsignedTx } = txCreateResp.data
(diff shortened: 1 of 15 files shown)


src/utils/constants.ts
```diff
@@ -0,0 +1,22 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { BILLION } from './numbers'
+
+export const minimalGasAmount = 20000
```
nop33, line 21: Let's make all constants with capital letters, like the `BILLION`:
- `MINIMAL_GAS_AMOUNT`
- `MINIMAL_GAS_PRICE`

nop33 (changes requested):

src/pages/Wallet/SendPage.tsx
```diff
@@ -17,22 +17,63 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+  }
+
+  if (amountNumber < minAmount) {
+    errorStateSetter(errorMessage)
+  } else {
+    if (currentErrorState) errorStateSetter('')
+  }
+
+  stateSetter(amount)
+}
```
nop33, line 71: I see that you are trying to keep the code DRY with this function, but from my side, I don't see a lot of added value. I feel like we unnecessarily increase the number of lines and the complexity of the code. Even if it looks a bit less DRY, I would simplify the code and reduce the total number of lines of the file. Lemme know what you think.

image

src/pages/Wallet/SendPage.tsx
```diff
@@ -17,22 +17,63 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+
+const getExpectedFee = (gasAmount: string, gasPriceInALPH: string) => {
```
nop33, line 73: I would move this utility function at the end of the file (before the styling part) so that the first JS code the reader reads after the imports is the actual component, and not utility functions.

src/pages/Wallet/SendPage.tsx
```diff
@@ -126,55 +206,99 @@ const SendPage = () => {
...
+          </Section>
+          <ExpandableSection sectionTitle="Advanced settings">
+            <Input
+              id="gas-amount"
+              placeholder="Gas amount"
+              value={gasAmount}
+              onChange={handleGasAmountChange}
+              type="number"
+              error={gasAmountError}
+            />
```
nop33, line 244: Shouldn't the `gas-amount` number input also have a `min` attribute, like the `gas-price` below?

nop33 (changes requested):

nop33:
@mvaivre regarding the conflicts, you want to keep these changes:

```js
import { ChangeEvent, useState } from 'react'
```

and

```js
import { minimalGasAmount, minimalGasPrice } from '../../utils/constants'
import { abbreviateAmount, convertToQALPH } from '../../utils/numbers'
```

src/pages/Wallet/SendPage.tsx
mvaivre: Mhh on this one I think that I prefer my version. The increased number of line is mostly caused by the declaration of the function's params and their typings, not by some logic that is hard to comprehend. 

I actually think that the condensed callbacks make it easier to understand the component logic & make it easier to identify and update the error messages if needed (readability can be improved further if we "hide" the `onAmountInputValueChange` at the bottom of the file.)

src/pages/Wallet/SendPage.tsx
mvaivre: Agreed. I'll move the `onAmountInputValueChange` as well, as explained in the other comment.

src/pages/Wallet/SendPage.tsx
nop33: Alright, agreed. Let's move the declaration of the `onAmountInputValueChange` utility to the end of file (before styling) so that it's not the first thing we read when opening the file (I find it nicer to read the component code directly and not implementation details).

src/pages/Wallet/SendPage.tsx
nop33: Oops, spoke before reading this XD

nop33:
Thanks for your changes and your feedback @mvaivre! I don't have any other changes to request. I will check out the branch now and test it locally. If I don't find any problems, I'll approve the PR!

polarker:
Just to confirm: the smallest gas price is `ALPH.nanoAlph(100)` on full node

mvaivre:
> Just to confirm: the smallest gas price is `ALPH.nanoAlph(100)` on full node

Isn't it 1 nanoAlph? @polarker  => https://github.com/alephium/alephium/blob/master/protocol/src/main/scala/org/alephium/protocol/model/package.scala#L31

polarker:
> > Just to confirm: the smallest gas price is `ALPH.nanoAlph(100)` on full node
> 
> Isn't it 1 nanoAlph? @polarker => https://github.com/alephium/alephium/blob/master/protocol/src/main/scala/org/alephium/protocol/model/package.scala#L31

That's the minimal for consensus. The app layer minimal is 100 nanoALPH for the first year to prevent DoS attacks

mvaivre:
> > > Just to confirm: the smallest gas price is `ALPH.nanoAlph(100)` on full node
> > 
> > 
> > Isn't it 1 nanoAlph? @polarker => https://github.com/alephium/alephium/blob/master/protocol/src/main/scala/org/alephium/protocol/model/package.scala#L31
> 
> That's the minimal for consensus. The app layer minimal is 100 nanoALPH for the first year to prevent DoS attacks

Thanks for the clarification! Is it written somewhere in the code atm? Would it make sense to centralise those constants in `alephium-js` as a single source of truth for the app layer?

polarker:
> > > > Just to confirm: the smallest gas price is `ALPH.nanoAlph(100)` on full node
> > > 
> > > 
> > > Isn't it 1 nanoAlph? @polarker => https://github.com/alephium/alephium/blob/master/protocol/src/main/scala/org/alephium/protocol/model/package.scala#L31
> > 
> > 
> > That's the minimal for consensus. The app layer minimal is 100 nanoALPH for the first year to prevent DoS attacks
> 
> Thanks for the clarification! Is it written somewhere in the code atm? Would it make sense to centralise those constants in `alephium-js` as a single source of truth for the app layer?

The logic is a bit involved: https://github.com/alephium/alephium/blob/master/flow/src/main/scala/org/alephium/flow/handler/TxHandler.scala#L76-L89

nop33:
I am using my local set up (node version `1.1.13`, explorer backend `master`). I created a transaction with the following configuration:

image

The data submitted to the API are the following:
```
amount: 100000000000000000000
gasPrice: 400000000000
gasAmount: 20002
```

The result was this:

image

This does not seem to be an error from the wallet side. It's either on the node side or the explorer side. We should investigate.

nop33:
@mvaivre I approved the changes, but can you review my commits as well? I can't assign you as a reviewer on your own PR. Thanks

nop33:
@tdroxler @polarker, please have a look at my comment and screenshots above. Is the explorer displayed transaction fee correct? I find it weird that the displayed value is not exactly the same as the one in the wallet.

mvaivre:
@nop33 Maybe a precision issue linked to our previous investigations on the wallet? We should def centralise our amount display methods in alephium-js asap, now that we've cleaned them up.

nop33:
> @nop33 Maybe a precision issue linked to our previous investigations on the wallet? 

@mvaivre you mean in the explorer? I haven't investigated yet, gotta check.

> We should def centralise our amount display methods in alephium-js asap, now that we've cleaned them up.

Totally agreed. I'll create an issue.

mvaivre:
> > @nop33 Maybe a precision issue linked to our previous investigations on the wallet?
> 
> @mvaivre you mean in the explorer? I haven't investigated yet, gotta check.

I mean, the improvements we've made on the Wallet should be reflected on the explorer :) The recent edge cases we've found with abbreviating the amount are probably existing in the explorer rn (hence the little precision issue you've found there). So yeah, will be good to centralise our methods in alephium-js.

mvaivre:
LGTM! Was funny to work around JS limitations with numbers huh? 🤯

nop33:
> I mean, the improvements we've made on the Wallet should be reflected on the explorer :) The recent edge cases we've found with abbreviating the amount are probably existing in the explorer rn (hence the little precision issue you've found there). So yeah, will be good to centralise our methods in alephium-js.

After my investigation, I discovered that you're right. The `abbreviateAmount` implementation of the explorer causes the precision issue. I'll start working on centralizing them on alephium-js immediately. We'll need to update all 3 projects.

PR #98, opened 2022-01-05
Final cleanup 🎉

nop33:


src/components/Badge.tsx
```diff
@@ -30,14 +30,12 @@ interface BadgeProps {
   prefix?: string
 }
 
-let Badge = ({ content, className, amount, prefix }: BadgeProps) => {
-  return (
```
nop33, line 34: I know you prefer to use the `return` keyword instead of the shorter arrow function syntax, so I only removed it from "dumb" components (dump == they do not have any logic, they simply return JSX) and kept it in utility and other functions. Let me know if you have a problem with that.

src/components/FloatingLogo.tsx
```diff
@@ -21,15 +21,16 @@ import styled from 'styled-components'
...
 import { deviceBreakPoints } from '../style/globalStyles'
 
-const FloatingLogo = styled(AlephiumLogoSVG)`
+const FloatingLogo = styled(AlephiumLogoSVG)<{ position?: 'top' | 'bottom' }>`
   position: absolute;
-  top: 50px;
   left: var(--spacing-5);
-  width: 60px;
-  height: 60px;
+  width: 35px;
```
nop33, line 27: Despite the change in width, the actual size of the logo didn't change. This simply fixes alignment issues.

src/contexts/global.tsx
```diff
@@ -88,7 +88,10 @@ export const GlobalContextProvider: FC<{ overrideContextValue?: PartialDeep<Glob
     return updatedSettings
   }
 
-  const lockWallet = () => setWallet(undefined)
+  const lockWallet = () => {
+    setCurrentUsername('')
```
nop33, line 92: After locking the wallet and opening the accounts settings, the list
would still display a "current" account. This ensures resetting
the current account when locking the wallet.

src/pages/WalletManagement/WalletWordsPage.tsx
```diff
@@ -34,10 +36,17 @@ import { useStepsContext } from '../../contexts/steps'
 import { useWalletContext } from '../../contexts/wallet'
 
 const WalletWordsPage = () => {
-  const { mnemonic, plainWallet } = useWalletContext()
+  const { mnemonic, plainWallet, setPlainWallet, setMnemonic } = useWalletContext()
   const { onButtonBack, onButtonNext } = useStepsContext()
   const { setSnackbarMessage } = useGlobalContext()
 
+  // Init wallet
```
nop33, line 43: I moved the actual wallet generation AFTER the account creation step. If the user never ends up creating an account, there's no need creating a wallet and setting it to our app's state. This also helps to clean up our routes component, so that it's only responsible for the routes.

src/components/Badge.tsx
mvaivre: Perfectly fine!

src/components/Amount.tsx
```diff
@@ -23,12 +23,7 @@ interface AmountProps {
...
 
-const Amount = ({ value, className }: AmountProps) => {
-  if (value !== undefined) {
-    return <span className={className}>{abbreviateAmount(value)}א</span>
-  } else {
-    return <span className={className}>- א</span>
-  }
-}
+const Amount = ({ value, className }: AmountProps) =>
+  value ? <span className={className}>{abbreviateAmount(value)}א</span> : <span className={className}>- א</span>
```
mvaivre, line 27: Careful, if `value` === 0, then `value` is falsy.

mvaivre (changes requested):
First quick review, didn't test on my computer yet. There's one comment about a potential issue (Amount component), the rest looks good :)

src/components/Amount.tsx
```diff
@@ -23,7 +23,8 @@ interface AmountProps {
   className?: string
 }
 
-const Amount = ({ value, className }: AmountProps) =>
-  value ? <span className={className}>{abbreviateAmount(value)}א</span> : <span className={className}>- א</span>
+const Amount = ({ value, className }: AmountProps) => (
+  <span className={className}>{value !== undefined ? abbreviateAmount(value) : '-'} א</span>
```
nop33, line 27: This adds a space between the value and the symbol. Is that OK?

image

mvaivre: Yes, makes sense.

mvaivre (approved):

nop33:
Rebased to `master` and fixed conflicts.

Issue #101, opened 2022-01-07
Sweep to the same address to reduce the number of UTXOs


polarker:
If the user wants to make a large tx using many UTXOs, the tx cannot be built due the limits on tx inputs and gas. We'd suggest users to merge the existing UTXO with `sweep-address` endpiont.

polarker:
@mvaivre This is high priority. Please consider to support this with high priority. It will be later if more users have this issue. I already got 2.

As we are discussed, we will use `UTXO consolidation` for the naming

polarker:
How should this high priority be reflected in your project management?

nop33:
I'll start working on this tmr and treat it as highest priority

mvaivre:
Yes, we're on it, it was discussed yesterday  [here](https://alephium.slack.com/archives/C026MHDJD2N/p1641559534009700?thread_ts=1641559534.009700&cid=C026MHDJD2N)

nop33:
@polarker @mvaivre I guess we can close this one now that the feature is implemented, correct?

polarker:
LGTM

PR #105, opened 2022-01-12
Implement UTXO consolidation modal

nop33:
Related to #101

Here's a beta version of this feature. Important things to note:
- In the video below I am *forcing* the consolidation pop-up to open. It is not opened because I am receiving an error from the API. I still haven't figured out how to accomplish that.
- We need to release a new version of alephium-js and update the dependency here.

https://user-images.githubusercontent.com/1579899/149185108-584f2e55-ac9b-4e07-94f0-ee68aea7a65d.mp4

### DIFF
--- a/src/pages/Wallet/SendPage.tsx
+++ b/src/pages/Wallet/SendPage.tsx
@@ -53,10 +53,13 @@ const SendPage = () => {
     gasAmount: MINIMAL_GAS_AMOUNT.toString(),
     gasPrice: abbreviateAmount(MINIMAL_GAS_PRICE)
   })
-  const [isSending, setIsSending] = useState(false)
+  const [isLoading, setIsLoading] = useState(false)
   const [step, setStep] = useState<StepIndex>(1)
   const [isConsolidateUTXOsModalVisible, setIsConsolidateUTXOsModalVisible] = useState(false)
   const [isConsolidating, setIsConsolidating] = useState(false)
+  const [consolidationRequired, setConsolidationRequired] = useState(false)
+  const [builtTxId, setBuiltTxId] = useState('')
+  const [builtUnsignedTx, setBuiltUnsignedTx] = useState('')
 
   useEffect(() => {
     if (step === 1) {
@@ -77,15 +80,16 @@ const SendPage = () => {
   }
 
   const confirmPassword = () => {
+    if (consolidationRequired) setIsConsolidateUTXOsModalVisible(false)
     setStep(3)
   }
 
-  const handleSend = async () => {
+  const buildTransaction = async () => {
     const { address, amount, gasAmount, gasPrice } = transactionData
     const isDataComplete = address && amount && gasPrice && gasAmount
 
     if (wallet && client && isDataComplete) {
-      setIsSending(true)
+      setIsLoading(true)
 
       const fullAmount = convertToQALPH(amount).toString()
 
@@ -99,11 +103,39 @@ const SendPage = () => {
           parseInt(gasAmount),
           convertToQALPH(gasPrice).toString()
         )
+        setBuiltTxId(txCreateResp.data.txId)
+        setBuiltUnsignedTx(txCreateResp.data.unsignedTx)
+        setStep(3)
+      } catch (e) {
+        // TODO: When API error codes are available, replace this substring check with a proper error code check
+        if (isHTTPError(e) && e.error?.detail && e.error.detail.includes('consolidating')) {
+          setIsConsolidateUTXOsModalVisible(true)
+          setConsolidationRequired(true)
+        } else {
+          setSnackbarMessage({
+            text: getHumanReadableError(e, 'Error while building the transaction'),
+            type: 'alert',
+            duration: 5000
+          })
+        }
+      }
 
-        const { txId, unsignedTx } = txCreateResp.data
-        const txSendResp = await signAndSendTransaction(txId, unsignedTx, address, wallet.privateKey)
+      setIsLoading(false)
+    }
+  }
+
+  const handleSend = async () => {
+    const { address, amount } = transactionData
+
+    if (builtTxId && builtUnsignedTx && wallet && address) {
+      setIsLoading(true)
+
+      try {
+        const txSendResp = await signAndSendTransaction(builtTxId, builtUnsignedTx, address, wallet.privateKey)
 
         if (txSendResp) {
+          const fullAmount = convertToQALPH(amount).toString()
+
           addPendingTx({
             txId: txSendResp.data.txId,
             toAddress: address,
@@ -128,7 +160,7 @@ const SendPage = () => {
         }
       }
 
-      setIsSending(false)
+      setIsLoading(false)
     }
   }
 
@@ -159,7 +191,6 @@ const SendPage = () => {
           type: 'info',
           duration: 15000
         })
-        setIsConsolidateUTXOsModalVisible(false)
         initialOnModalClose.current()
       } catch (e) {
         setSnackbarMessage({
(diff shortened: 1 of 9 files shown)


nop33:
The error logic is implemented. I just don't know how to test it.

mvaivre:
Two minor comments. Looks good!

nop33:
Where are the comments? 😅

mvaivre:
> Where are the comments? 😅

Arh, github and its review flow :P There.

src/pages/Wallet/SendPage.tsx
```diff
@@ -290,6 +351,40 @@ const CheckTransactionContent = ({ data, onSend }: CheckTransactionContentProps)
   )
 }
 
+interface ConsolidateUTXOsModalProps {
+  onConsolidateClick: () => void
+  onClose: () => void
+  isConsolidating: boolean
+}
+
+const ConsolidateUTXOsModal = ({ onConsolidateClick, onClose, isConsolidating }: ConsolidateUTXOsModalProps) => {
```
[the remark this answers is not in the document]
nop33: done

src/pages/Wallet/WalletHomePage.tsx
```diff
@@ -180,7 +181,21 @@ const WalletHomePage = () => {
         <WalletActions>
           <ActionsTitle>Quick actions</ActionsTitle>
           <ActionButton Icon={QrCode} label="Show address" link="/wallet/address" />
-          <ActionButton Icon={Send} label="Send token" link="/wallet/send" />
+          {!thereArePendingConsilidationTxs ? (
```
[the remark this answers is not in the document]
nop33: done

src/pages/Wallet/SendPage.tsx
```diff
@@ -97,32 +101,83 @@ const SendPage = () => {
...
+          })
+        }
 
-        const signature = client.clique.transactionSign(txId, wallet.privateKey)
+        setSnackbarMessage({ text: 'Transaction sent!', type: 'success' })
+        history.push('/wallet')
+      } catch (e) {
+        // TODO: When API error codes are available, replace this substring check with a proper error code check
+        if (isHTTPError(e) && e.error?.detail && e.error.detail.includes('consolidating')) {
+          setIsConsolidateUTXOsModalVisible(true)
```
polarker, line 121: When UTXO modal is visible, will it be sticky to top? I mean we can not bring the main window of wallet to the front

src/pages/Wallet/SendPage.tsx
```diff
@@ -97,32 +101,83 @@ const SendPage = () => {
...
+        )
+        for (const { txId, unsignedTx } of txCreateResp.data.unsignedTxs) {
+          const txSendResp = await signAndSendTransaction(txId, unsignedTx, wallet.address, wallet.privateKey)
+          if (txSendResp) {
+            addPendingTx({
+              txId: txSendResp.data.txId,
+              toAddress: wallet.address,
+              timestamp: new Date().getTime(),
+              amount: '',
+              type: 'consolidation'
```
polarker, line 152: It's better to define `'consolidation'` as constants, so it's easier to modify in the future, or to avoid typos

src/pages/Wallet/SendPage.tsx
```diff
@@ -97,32 +101,83 @@ const SendPage = () => {
...
         })
       }
-
-      setIsSending(false)
+      setIsConsolidating(false)
     }
   }
 
+  const signAndSendTransaction = async (txId: string, unsignedTx: string, toAddress: string, privateKey: string) => {
+    if (!client) return
```
polarker, line 176: Should we throw an error if `client` is not defined? Or we could assume that `client` is define when this function is called

polarker (commented):
Several minor comments

src/pages/Wallet/SendPage.tsx
nop33: Yes, it is sticky to the front. To close it (in order to access the rest of the wallet), you either have to click "Consilidate" or the "X" button.

tdroxler:
In the flow in the video, should we call the `build` endpoint when clicking `Check`? Like this we'll get the error earlier. Then the `Send` is only signing + `submit`.

mvaivre:
> In the flow in the video, should we call the `build` endpoint when clicking `Check`? Like this we'll get the error earlier. Then the `Send` is only signing + `submit`.

That's a very good idea!

nop33:
> In the flow in the video, should we call the build endpoint when clicking Check? Like this we'll get the error earlier. Then the Send is only signing + submit.

I agree! Should we require the password of the user before consolidating?

mvaivre:
> > In the flow in the video, should we call the build endpoint when clicking Check? Like this we'll get the error earlier. Then the Send is only signing + submit.
> 
> I agree! Should we require the password of the user before consolidating?

We're sending a TX, so yes - good point - I think we should.

src/pages/Wallet/SendPage.tsx
nop33: We can assume it's defined, because without a `client`, the user wouldn't be able to login to the wallet.

tdroxler:
> > In the flow in the video, should we call the build endpoint when clicking Check? Like this we'll get the error earlier. Then the Send is only signing + submit.
> 
> I agree! Should we require the password of the user before consolidating?

yes agree with Mika, it's still a send, so password require

nop33:
Here is the new flow!

https://user-images.githubusercontent.com/1579899/149290218-f66b42e3-32bc-4c55-832f-1fb9291799db.mp4

nop33:
The only part of the code that has not been tested is the following. I'll try again to produce this error from the wallet flow (not through Swagger)

```js
} catch (e) {
        // TODO: When API error codes are available, replace this substring check with a proper error code check
        if (isHTTPError(e) && e.error?.detail && e.error.detail.includes('consolidating')) {
         ...
        } else {
          setSnackbarMessage({
            text: getHumanReadableError(e, 'Error while building the transaction'),
            type: 'alert',
            duration: 5000
          })
        }
      }
```

src/pages/Wallet/SendPage.tsx
nop33: done

tdroxler:
@nop33 that's already better, but I was even thinking to raise the error when clicking `Check`? because I'm wondering if it's useful to show the recap while we could know that it will fail?

nop33:
> @nop33 that's already better, but I was even thinking to raise the error when clicking Check? because I'm wondering if it's useful to show the recap while we could know that it will fail?

Ah ok! I misunderstood!

nop33:
THERE WE GO!!!

https://user-images.githubusercontent.com/1579899/149296181-6e114c72-82b8-4fe9-90f9-6b9401145be9.mp4

src/utils/constants.ts
```diff
@@ -20,3 +20,8 @@ import { BILLION } from './numbers'
 
 export const MINIMAL_GAS_AMOUNT = 20000
 export const MINIMAL_GAS_PRICE = BigInt(BILLION * 100) // 100 nanoALPH for the first year to prevent DoS attacks
+
+export const txTypes = {
```
mvaivre, line 24: I'm not sure that the `constants.ts` file is the best place to define types.
Should we init a `src/types` folder? WDYT @nop33

nop33: It's not TypeScript types. It's JS string constants.

src/pages/Wallet/SendPage.tsx
nop33: @polarker actually, TypeScript does not allow us to do typos:

image

As for easier to modify in the future, we can easily do a global search+replace of the strings.

mvaivre (approved):

PR #107, opened 2022-01-13
Refactor DOM structure, allow to use scroll hook to fade header color

mvaivre:
Here we go @nop33 

Not 100% happy, but it's a working quick fix, better than what we had before.
Please take some time to try navigatin / scrolling in every part of the UI to make sure there are no side effect.


nop33 (approved):
Looks like everything work like a charm 🌟

Issue #116, opened 2022-01-18
Find a clearer way to display amounts


nop33:
I had this discussion with @mvaivre the other day:

Something that has been bugging me a lot and I see other people having the same issue: I find it hard to distinguish the abrreviated amounts.
1.000K ℵ
1.123 ℵ
1.002K ℵ
1.002 ℵ
It all looks too similar, especially when they are aligned on the right, as it is the case in the wallet (and also explorer).

Using 3 decimal points is triggering the "thousands" pattern in my brain and I always think (in the example above) that 1.002 ℵ is 1 thousand and 2 ℵ.

![image](https://user-images.githubusercontent.com/1579899/149897626-ef9e633b-2a6e-4d3b-bf6c-dcd7e1b25b85.png)

Things we could test out:
1. Show only 1 or 2 decimals when using K, M etc, so that the above would be 8.6K ℵ or 8.68K ℵ
2. Abbreviate the 0s in general when they are not needed so that the above becomes 1 ℵ
3. Showing only 2 decimals instead of 3, which is a much more common pattern when talking about currencies (cents), unless the amount is less than 0.01 (like consolidation transaction fees), in which case we display the full precision (or just 3 decimals, like now).
4. Do not use the letter K so the above becomes 8681.23 ℵ (or even better, 8'681.23 ℵ), and only start using M and B when we surpass a million.

I personally like the combination of options 3 and 4.

polarker:
How does Metamask do it?

nop33:
| Metamask | BRD | Argent |
| ---------- | -----| -------|
| Metamask seems to display full precision, abbreviating unnecessary 0s, with no decimals when not needed. I don't know how to test to see how they display values > 999 and > 999999. | Bread wallet seems to be using full precision. I also don't know how to test large amounts. | Argent seems to use max 4 decimals, abbreviating 0s when possible and not using the "K" letter to signify thousands (but a "," instead, which I find confusing) |
| ![image](https://user-images.githubusercontent.com/1579899/149902212-f000e8c9-1741-499d-a0d0-63386aaee4a5.png) | ![image](https://user-images.githubusercontent.com/1579899/149903330-ce39acd4-b541-4286-ab41-13c18c11d97a.png) | ![image](https://user-images.githubusercontent.com/1579899/149902873-ca7b1482-31bd-4ace-8cb2-c0393abb7bcf.png) |

mvaivre:
Thanks for summarising the discussion in this Issue @nop33.

[...]

PR #117, opened 2022-01-18
Fix amount display

nop33:
Related to #116

### DIFF
--- a/src/components/Amount.tsx
+++ b/src/components/Amount.tsx
@@ -59,7 +59,7 @@ const Amount = ({ value, className }: AmountProps) => {
 }
 
 const Decimals = styled.span`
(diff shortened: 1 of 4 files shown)


src/components/Amount.tsx
```diff
@@ -30,12 +32,34 @@ const Amount = ({ value, className }: AmountProps) => {
...
+      ) : (
+        '-'
+      )}
+      {' ℵ'}
+    </span>
   )
 }
 
+const Decimals = styled.span`
+  /* font-size: 0.85em; */
```
nop33, line 62: @mvaivre, I just commented it out so that we can experiment with this later on without reinventing the wheel that I just invented 😅

PR #118, opened 2022-01-25
Derive new addresses

nop33:
There is audio as well 🗣️ 😅 

https://user-images.githubusercontent.com/1579899/150992482-e073d97b-395d-4348-a27e-73e9326de512.mp4


src/components/Modal.tsx
```diff
@@ -144,9 +144,26 @@ const CloseButton = styled(Button)`
 `
 
 const ModalContent = styled.div`
-  overflow-y: auto;
-  overflow-x: hidden;
+  /* overflow-y: auto;
+  overflow-x: hidden; */
```
nop33, line 148: This was commented out so that the color picker popup can be properly displayed. Not sure if it has any other implications and wanted to ask if it's ok to remove it before I do.

src/pages/Wallet/NewAddressPage.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+  const generateNewAddress = useCallback(
+    (group?: number) => {
+      if (wallet?.seed) {
+        const data = deriveNewAddressData(wallet.seed, group, undefined, currentAddressIndexes)
+        setNewAddressData(data)
+        setNewAddressGroup(group || addressToGroup(data.address, TOTAL_NUMBER_OF_GROUPS))
+      }
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
nop33, line 55: I specifically **DONT** want `currentAddressIndexes` to be listed here...

mvaivre:
Question about the testing process: I guess I don't have to signal layout issues (when resizing) or visual details yet, correct? Or do you need me to list what I see?

nop33:
Please, go ahead and list everything you don't like. I know that the table is not responsive (if that's what you are mainly referring to), but I don't know if we want it to be. Anyway, let's start the conversations here, it's a good place.

src/utils/addresses.ts
```diff
@@ -16,10 +16,54 @@ You should have received a copy of the GNU Lesser General Public License
...
+
 export const checkAddressValidity = (address: string) => {
   const match = address.match(/^[1-9A-HJ-NP-Za-km-z]+$/)
 
   if (match === null) return false
 
   return match[0] === address && address
 }
+
+export const loadStoredAddressesInfoOfAccount = (username: string) => {
```
mvaivre, line 34: Maybe we could define (cast) the expected return type for a more type strict approach?

src/pages/Wallet/AddressesPage.tsx
```diff
@@ -0,0 +1,169 @@
...
+  tokens: number
+  group: number
+  amount: bigint
+}
+
+const AddressesPage = () => {
+  const [isGenerateNewAddressModalOpen, setIsGenerateNewAddressModalOpen] = useState(false)
+  const { addressesInfo, wallet, client } = useGlobalContext()
+  const [addressesTableData, setAddressesTableData] = useState<AddressesTableData[]>([])
+  const headers = [
```
mvaivre, line 57: This is fully static and hence could be defined outside of the component.
I wonder if there's a performance impact? If not, I'm also fine with packing it inside there. What is your opinion?

Also, we could define the type to avoid mistakes:  `headers: TableProps['headers'] = ...`

mvaivre (commented):
First part of my review. Great job!! 
There's quite a lot left to check, so I need a little break ;)

src/components/Inputs/ColoredLabel.tsx
```diff
@@ -0,0 +1,81 @@
...
+interface ColoredLabelProps {
+  value: ColoredLabelValue
+  className?: string
+  onChange: ({ title, color }: ColoredLabelValue) => void
+  disabled?: boolean
+  placeholder?: string
+  id?: string
+}
+
+let ColoredLabel = ({ placeholder, disabled, onChange, value, className, id }: ColoredLabelProps) => {
```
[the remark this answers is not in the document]
nop33: > I'd rename that to ColoredLabelInput, as I had to read the code to understand that it's not just the label we're talking about (+ I like to have descriptive names when searching for files in VSCode :)).

👍 

> We're not coherent in our use of Styled Components. Here you declare a variable that is wrapped later on. In the other approach, we use constants only (in this case, we would have a InputContainer carrying the styles.

The first reason why I prefer my approach is, as you said, one less "wrapper" element.

Second reason is that the styles of my React component will always be able to be extended by another component. What I mean:

```tsx
// ComponentName.tsx
let ComponentName = ({ className }: { className?: string }) => {
  return <div className={className}>hello</div>
}

ComponentName = styled(ComponentName)`
  color: red;
`

export default ComponentName
```

```tsx
// Page.tsx
import ComponentName from './ComponentName'

const Page = () => {
  return <ComponentNameStyled />
}

const ComponentNameStyled = styled(ComponentName)`
  color: blue; // The color of the "hello" word will now be blue.
`
```

The reason this is possible is that my approach forces me to pass and attach `className` attribute. I was basically inspired by their official docs regarding the [TypeScript caveat](https://styled-components.com/docs/api#caveat-with-classname).

With the first approach, if we forget to pass the `className`, our component won't be re-styled. I had this problem a few times and it took me a while to figure out what is going on:

```tsx
// ComponentName.tsx
const ComponentName = () => {
  return <ComponentNameWrapper>hello<ComponentNameWrapper>
}

const ComponentNameWrapper = styled.div`
  color: red;
`

export default ComponentName
```

```tsx
// Page.tsx
import ComponentName from './ComponentName'

const Page = () => {
  return <ComponentNameStyled />
}

const ComponentNameStyled = styled(ComponentName)`
  color: blue;  // Color will still be red, since `className` was not passed.
`
```

nop33:
> This will cause problem when scrolling is necessary in modal with more content.
> Maybe just allowing overflow-x could work?

There seems to be no easy way for a child element to escape the `overflow: hidden` of the parent. I tried to play a bit with `oveflow-x` and overflow-y` with no success in solving this problem (either the color popup will be cut out or the content will overflow. https://stackoverflow.com/a/10446410/1644591

A quick fix would be to use the Twitter-style component instead of the Block one (see [docs](https://casesandberg.github.io/react-color/)):

image

Another quick fix is to position the picker so that it doesn't go outside the modal (and hide the "triangle"):

image

nop33:
Another approach would be to display it on the side (but this requires a bit of arbitrary positioning values)

image

mvaivre:
Yes, I thought about option 1. Best compromise IMO.

src/components/PageComponents/PageContainers.tsx
```diff
@@ -147,3 +147,32 @@ export const FooterActionsContainer = styled(Section)`
   margin-top: var(--spacing-5);
   width: 100%;
 `
+
+export const ContainerNextToSidebar: FC = ({ children, ...props }) => (
```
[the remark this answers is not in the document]
nop33: I like `MainContent`, since it is a `main` HTML element as well 💪

src/components/Inputs/ColoredLabel.tsx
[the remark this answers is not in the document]
mvaivre: Yes, it's an interesting point. For now, I selectively added the `className` prop when needed.
If we decide to go with "your" strategy, we should.. well, adapt all our components for consistency sake ⭐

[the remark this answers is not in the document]
nop33: I agree, it makes sense to follow one strategy. What are the arguments against "my" approach? Not sure I understand your argument of _"the other one is more immutable and systematic"_.

Also, whichever approach we decide upon, we don't need to change the whole code base in one go. We can change it whenever we face a component that violates our decision while working on new features. Otherwise it'll be a very boring task 😅

src/pages/Wallet/AddressesPage.tsx
nop33: Good catch. I suppose that every time the component rerenders, a new location in memory is being used to define `const headers`. I would guess that the garbage collector would clean the other one, but in any case I agree with you that it's better to move it out of the component.

src/components/Amount.tsx
```diff
@@ -59,7 +64,7 @@ const Amount = ({ value, className }: AmountProps) => {
 }
 
 const Decimals = styled.span`
-  /* font-size: 0.85em; */
+  opacity: 0.6;
```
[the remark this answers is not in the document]
nop33: Done!

src/components/Inputs/Input.tsx
```diff
@@ -35,7 +35,7 @@ import {
 const Input = ({ placeholder, error, isValid, disabled, onChange, value, ...props }: InputProps) => {
   const [canBeAnimated, setCanBeAnimated] = useState(false)
 
-  const className = classNames({
+  const booleanClassName = classNames({
```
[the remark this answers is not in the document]
nop33: Nice one, done.

src/components/Table.tsx
```diff
@@ -0,0 +1,100 @@
...
+)
+
+Table = styled(Table)`
+  border: 1px solid ${({ theme }) => theme.border.primary};
+  border-radius: var(--radius);
+  background-color: ${({ theme }) => theme.bg.secondary};
+  display: table;
+`
+
+interface TableCellProps {
```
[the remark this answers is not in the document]
nop33: Done.

src/utils/addresses.ts
nop33: good one, done

src/pages/Wallet/AddressesPage.tsx
nop33: Updated.

src/components/Button.tsx
```diff
@@ -30,6 +30,7 @@ interface ButtonProps extends HTMLMotionProps<'button'> {
   transparent?: boolean
   squared?: boolean
   submit?: boolean
+  narrow?: boolean
```
mvaivre, line 33: It's reducing the height no? Let's maybe use `short` instead?

src/pages/Wallet/NewAddressPage.tsx
mvaivre: In that case, maybe use a `ref` (`useRef`) to store the `initialAddressIndexes`, that you would then use in `deriveNewAddressData` ? This would avoid overriding the eslint rule while also avoiding refreshing the address index list.

src/utils/addresses.ts
```diff
@@ -16,10 +16,54 @@ You should have received a copy of the GNU Lesser General Public License
...
+    })
+  } else if (
+    existingAddressInfo.label !== label ||
+    existingAddressInfo.color !== color ||
+    existingAddressInfo.isMain !== isMain
+  ) {
+    existingAddressInfo.label = label
+    existingAddressInfo.color = color
+    existingAddressInfo.isMain = isMain
+  }
```
mvaivre, line 67: The "else if" condition can be simplified greatly with `Object.assign(target, source)` to automatically merge changes, as long as the object as a depth of 1. 

Something like:
```
else {
 newAddressInfo = Object.assign({label, color, isMain}, existingAddressInfo)
}
```

mvaivre (commented):
Round 2.

src/components/Inputs/ColoredLabel.tsx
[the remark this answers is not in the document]
mvaivre: Let's do that then! When we come across new components or components that need to be updated, we switch to the "wrapping" strategy.

src/utils/addresses.ts
nop33: Done!

src/pages/Wallet/AddressesPage.tsx
```diff
@@ -0,0 +1,169 @@
...
+        tableData.push({
+          hash: address,
+          isMain,
+          label: {
+            text: label,
+            color
+          },
+          lastUsed: '-', // TODO
+          transactions: txNumber,
+          tokens: 0, // TODO
```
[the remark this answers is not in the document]
nop33: Fixed!

src/components/Button.tsx
mvaivre: Yes I saw that... then `short` is really not that bad no? 
As you wish, it was just a suggestion, I let you decide.

src/components/Inputs/ColoredLabel.tsx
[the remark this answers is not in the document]
mvaivre: As in "your" approach (option 1)

nop33:
@mvaivre I believe I've addressed all your comments! Thanks so much for the valuable feedback, learnt a lot!

polarker:
A general question to make sure I am following the approach: if we 1) derive several new addresses; 2) remove the wallet; 3) restore the wallet, we will need to derive the list of addresses again right?

nop33:
> A general question to make sure I am following the approach: if we 1) derive several new addresses; 2) remove the wallet; 3) restore the wallet, we will need to derive the list of addresses again right?

Yes. We'll have no way of knowing however which addresses need to be derived. We'd need to implement some kind of "discovery".

mvaivre:
@polarker Yes. We've discussed that with @tdroxler already, we need to design our own address discovery strategy. The one [described in BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery) recommends a gap of 20 (ie. stop scanning after 20 addresses without any txs). As we have multiple groups, using 20 generally won't be enough to actually discover all used addresses.

Maybe you already have some thoughts on that topic Cheng? What would you recommend?

mvaivre (commented):
LGTM! Great job 🚀 

It's a first PR, so I don't dive into layout stuff for now. We'll keep that for the next one :)

polarker:
> keep

Yes, I know this issue and I guess Thomas will work on the discover mechanism. I was just to keep track of the high-level approach here

polarker:
One feature I am not sure if it's planned or not: when restoring or creating a wallet, provide users the option to get 4 addresses with one in each group

mvaivre:
> One feature I am not sure if it's planned or not: when restoring or creating a wallet, provide users the option to get 4 addresses with one in each group

[If you check the mockups](https://www.figma.com/proto/wDWcIxJIq5aBTLwMPGb1Js/Alephium-Desktop-Wallet---DeFi-evolution?page-id=0%3A1&node-id=5%3A30&viewport=268%2C48%2C0.3&scaling=min-zoom&starting-point-node-id=1%3A2), you'll see that "action shortcuts" are planned to greatly ease some technical workflow, like consolidating UTXOs manually or as you just mentioned: creating 1 address for each group in one go.

polarker:
> > One feature I am not sure if it's planned or not: when restoring or creating a wallet, provide users the option to get 4 addresses with one in each group
> 
> [If you check the mockups](https://www.figma.com/proto/wDWcIxJIq5aBTLwMPGb1Js/Alephium-Desktop-Wallet---DeFi-evolution?page-id=0%3A1&node-id=5%3A30&viewport=268%2C48%2C0.3&scaling=min-zoom&starting-point-node-id=1%3A2), you'll see that some "action shortcuts" are planned to greatly ease some technical workflow, like consolidating UTXOs manually or as you just mentioned: creating 1 address for each group in one go.

Very cool! Generating mining addresses mostly happen when the wallet is created. We could support it when the wallet is ready, but I am not sure what is the best way to generate if for example I already have 1 address in the wallet. In this case, do we generate 3 new addresses, or we want to generate 4 new addresses?

The more complicated case is that I might have 2 addresses in the wallet and they can be in the same or different groups.

tdroxler:
sorry I'm late to the party, just noticed that PR :sweat_smile: The video looks super nice, great job @nop33 
Yes @polarker for address discovery, we were wondering if we should simply go with 20 addresses gap per group, so we'll have to check 80 addresses? anyway we'll make that gap limit configurable, I don't remember in which wallet, but Mika shown me one were we could change it.

mvaivre:
> > > One feature I am not sure if it's planned or not: when restoring or creating a wallet, provide users the option to get 4 addresses with one in each group
> > 
> > 
> > [If you check the mockups](https://www.figma.com/proto/wDWcIxJIq5aBTLwMPGb1Js/Alephium-Desktop-Wallet---DeFi-evolution?page-id=0%3A1&node-id=5%3A30&viewport=268%2C48%2C0.3&scaling=min-zoom&starting-point-node-id=1%3A2), you'll see that some "action shortcuts" are planned to greatly ease some technical workflow, like consolidating UTXOs manually or as you just mentioned: creating 1 address for each group in one go.
> 
> Very cool! Generating mining addresses mostly happen when the wallet is created. We could support it when the wallet is ready, but I am not sure what is the best way to generate if for example I already have 1 address in the wallet. In this case, do we generate 3 new addresses, or we want to generate 4 new addresses?

Thanks for the feedback! We plan to allow "hidding" address with 0 fund. In that case it would be easy to :
1) Create a new wallet
2) Generate the 4 "mining" addresses
3) Hide the initial unused address.

> The more complicated case is that I might have 2 addresses in the wallet and they can be in the same or different groups.

What makes this case more complicated? If we follow what I said above, we systematically generate a new set of 4 addresses. The user can then transfer funds if needed and hide unused addresses.

And don't forget that it's still possible to generate addresses 1 by 1! So if a user already has some addresses - say 2 of them - she could generate the 2 others manually, selecting the groups etc.

nop33:
I totally agree with @mvaivre. This is actually something I proposed during a call with @tdroxler and @mika (having the 4 addresses generated automagically at the wallet creation step, hide 3 of them, and only display them when the user needs them.

> so we'll have to check 80 addresses?

More realistically, we'll be checking more than 80 addresses if we start the discovery process from address index `0` until we find `20` addresses in each group. A quick test shows that `107` addresses were discovered until at least 20 belong in each group:

<details><summary>Click to see code</summary>

```ts
it.only('test discovery', () => {
      const addresses = [[] as Array<string>, [] as Array<string>, [] as Array<string>, [] as Array<string>]
      const skipAddressIndexes = []

      while (
        addresses[0].length < 20 ||
        addresses[1].length < 20 ||
        addresses[2].length < 20 ||
        addresses[3].length < 20
      ) {
        const newAddressData = walletUtils.deriveNewAddressData(seed, undefined, undefined, skipAddressIndexes)
        const groupOfNewAddress = addressToGroup(newAddressData.address, TOTAL_NUMBER_OF_GROUPS)
        addresses[groupOfNewAddress].push(newAddressData.address)
        skipAddressIndexes.push(newAddressData.addressIndex)
      }

      console.log(addresses)
      console.log(skipAddressIndexes.length)
    })

```

result:
```
[
      [
        '1L92NdU2paBXNKHGkx5hhCYntsBb6KK1dtorgdXVe5Ui',
        '127rWcqJ3HZuEpAqyGQksQKbNEaVLBk5uiaQ7hQNTAGjV',
        '1FnjgFvzatf5jSXdXhGCKb2LmWdhHNAHbAM9FPh1tJNbq',
        '16BXwDUKP7BXNsDbwjCcyxPA49MCizj9cBXPuWSGcyBC6',
        '1GGXCGA85ZpuimwxfuiHz3KHQU1mSGARbFbHHH5qGePMS',
        '1CLzhk8sBfidSNko6CLcRTg7UPDzy5xGd6dA9ntzB2Uy2',
        '19XJ9xiG7FCSY53KsyN5VfGfbzcTTbqUuN8GvVeyTNDvz',
        '1H1p9ydjjJcYhmw7Td8pKbGFqYwZvdDmfD1CSeMK1cNDG',
        '1hHMeSBUvwNgfdfKcxWuL8LgnRT4uCNNmbdz7xYEww5J',
        '16jFdLZWmgUmVovvhakX6WxTaotFmaLV8M53KXJyFRnaE',
        '1APbfTfvpvvsjKLKqYPyo3RrvgsJfjR4ZpFUUnNzJmEiP',
        '1DjALtzKSViLGaEgQxRQXSjjdnQJYwT9A6Lf7BYMf2H5E',
        '1CRsa9qMNcaagEpyfxFFxY1Hd4ZSAkS2v6BWV5NARWj2K',
        '1Q8ZnsW9KA3MJDVdn7tbQFdJczHC7W3FqnTpkPnWD9Mp',
        '1HFZAciSob9pRZu9xsaX2V1syyTMHiaoUkAL3Ka8QLW4F',
        '1579iJauqZuF24hVGnVjA7BmoQJWibWPzS6kSpb3VKbKv',
        '16NZKwRi3eovSE4ypNEHC2G4Rx8BfiQhiikkMXpJ63yK8',
        '13jMv31cPxPBfPNjZZTTnjLPGj2HAJi1znNqccCnuimrm',
        '12y9rNzJRUFv855pSHaARsuf1GhH3TS9ziog2SwcgjFDH',
        '175s2xgZmBFLsx6jXpDFiUo9QTRp9x6wpTvFSi1etL7jt',
        '1HSUZPJgwa7xeWY5CnSiTdovDfX9yZ5e6ioCMsWa6G5Xa',
        '17apERZdZsGnkFq1KVxGWnVyxEdu6Nd3r2N2D8QiwbHjU',
        '13wKEKJwvX9FRDPBc3zZejZMujGcQNdu3Tu6uxMaAETmX'
      ],
      [
        '1GMqZF89xD9MEorzVpjbUsEQAjypgsEze8gawbvPapfPf',
        '1diLimTvGjoWhWSiBYcfjQgVuTwuFuTmDE5ZTY7hfMCW',
        '1FXmT48rPXneNH4tqUaAKfWjBqAdxAhbRfZguZx5LSA4E',
        '1CjPSqJ5CcvruF8XiDpEUMhChKc5wsNW6x2CCSb3ZDibY',
        '14aTk3asKrCaiLtyXwnxrPosy391W85hxCz491uq5E9q9',
        '18GD2QRjL1vcStn5sYqXSvTHkqJq6LAssomumciQ5N7y4',
        '1E5Rr88sj5oE1NwgCdPubt7m5vnnY2LrSLkLDJyMwWQyV',
        '16R1mZXqkkmad3w8w7dpYu3FfAUkNsXewaC7v2uijRnb9',
        '13XbvVJ4LmTHTiA2Sm1Zve4epXLxREKP8CmBMDwi3v3nd',
        '12GKvSprCuvwCB93fAJY8PrLpoUAWiwkCuVejWDChHQax',
        '19oaX5Vyb2FLAfyLarJDKx76ju9jHkfXm4Dja21mmjckx',
        '1EYcQPgR7D6URdDAkHeSbwJXXHhhZveiJ4qkG7ZW3kKPK',
        '1EARTefFu47K8xuRA3ntT1tZiLMoaJUA3djxBbbh8W1db',
        '14mXNeyeXZbMryMCcWUvbS7NsWMiVwM37RqjJHjEEZEzr',
        '1Hvp6KaydavK95ErQTrH7wbQgMUJnnKzgdBEWAMT9WNGT',
        '1AN5nuK5YoQA9YSM1ykPDWj8Kb1mHzbd1pXt1N8dorDzL',
        '157k4femc52hSkHDQNZ3FiCe8FBNP1qEsacNB1QFGvWr5',
        '1Cwc68CQhas2CDfEpKGhVWTBotbg9Zj6mpaW5v6VoVZQz',
        '18NAG4t6BPripFgaTZsR2nfjQSN5TyWBBzJd8pCy8QLuN',
        '18xfJSJoC2d14jVxK2hJmSiBGH3CZmyenWsQRMMXkAomJ',
        '12W81MyfG54s15KzRe2MBJbRqp834UiYVMawQGnbMCBan',
        '1Ed4vcBBruc5LcRXYspwz5wD4tnERWuGaav94zwDwF44W',
        '1EYEE9TAk5R8FYtGSXBexzWEZS7JSzJZKnpnp6xJJ52dt',
        '1DhLN5nWEBt92U2ThhJK33XRGDRmFuLwgzMd8c6x4pwso',
        '19mjaW8JyjMQiAhjNvKf6sC3f9RMmRiBGnXVXDFonBQLb',
        '13gtjG7VnbEaQVsHD7ps5Uuwi2TTruv6vzcHrHsDK5uWw',
        '1CAXm5yz8BANF6mcKbBScG83jLPvaCgtrKMcmYb2HuuUk',
        '18yu3Lw1guz8zeDXcmAGViFRgB8pn5v5kDPpuh2pQwBY2',
        '16odRbzap7hDzN8ppg8CaXRxsZiQ6wFJxnPWzftoJ1JyB',
        '13Trc8Qj3KdtnmDzwUMWt6DbD6MrvTefZLRXwfeKRoP6w',
        '12SmMSrW5ASrP5ZtYWm88U5yyVU7rhX5qCDmyCs6eo4aq',
        '1BYxSJn6hhDb2sRz3jsrtovJjVjDKCTzCWhTEa5TRMJj1',
        '176khXfaaLDPcaSf7z6TmNFTrEvCz9ZLENzHu17huby68'
      ],
      [
        '12kMjpXxCHDebp8ChJYSEyWBgJzmUMSnvgD4EXK6E4kCa',
        '1HbGH8W9b1t9DJLHv2VkFVYhTTLJE5ZTmU9J8CPAnovfF',
        '19z38nux3uJVhpHN6jTWVVdis28LfwhumogDBGiDH8gb4',
        '1BXNquCmCDnxbXWZGVgDXwoVeMmGwLFgEjedDott6y5HM',
        '15jvBQBCmgaagEuqtUBjtvJrv6rBPg4EpSUqLWhZuZuk8',
        '16mjMgUTeUPVd3sE8k1GPkCSyR5B13HC6g18uQXdzJA3V',
        '1FVNS8XQiDKhcBPmRtXhGaabjLaeGtWYpHUrQxeWAmASk',
        '1CQuZKP75js9eJ9EayRy62bzvqFyKAmi8TUwx9nZwpXYC',
        '14n67mDMkMuj7aNPjj2vS2RtjkKFT9UfAZLwjwYS31Kyq',
        '1DCamxMSu3djFc22KPCzZGHfr7yGPsDPu9LcKksHRaNQD',
        '1EjKPC1qj5zLDs2hJAsjH8pRXw4hJECkqywvjAKpsxGAu',
        '1JBGxsQ8MvfjjHDXXJWnTKiysq2DKUai2np1UobdKA6bq',
        '1EcRgeRd8oueGuQHU7KiVVt3PfN7fuCokHv89dxbVsi3H',
        '1FeJsLnuMuuoiCkhuD1NQw1RUH9P3n5d3kM3BNm4rky5k',
        '1Nft1MesUVA7Qjg8asNs2utygodLqQuMuyyh4eBwmzxA',
        '12Ha3bXQ6Nmj2dSJSXZNa27ZqNoJJm9aMGPQ2ocxXP2ff',
        '1gs7qbroz32UYs9xvYxUw4fLvNu5SbdpQUZviQ15ZRHD',
        '1B2KkoNLiM42uAYcNVLjucdTQt7Qa7rJgaWzvoNhvX8aq',
        '1DDvPNgoYex8o8YffhRhv2LDdBxMucuxv5oBiG3GUAnXa',
        '12tTWD7yk5J2q9gzQccULwFkNbQ18A4oMtWBKuj9MiEYT',
        '17pu1U3rDvMywwWTVs5u5ddqyRjZqWpT5iGazypHuVzmU',
        '186YcZzNAmqCwZ1ujfgpHEcAV2usBRtVb6XZ6uB1zVvLk',
        '12UNSZrXfEaUPjFRD8Uxh6FyBqYzSak48PTUKVGgmJvge',
        '196isPugkCvUrmJFDpsiPP3gxrYSbzvkTC1hEV5J1f9kW',
        '1GC8RJ3984MER58oyz3aAmBrZ169wYrug3NXTuNURs62t',
        '149xu862nQjwjzLph97gnb3JwCBu8NKe8CpWVjEjYkpSZ',
        '1Do3rRfvpbKvZR1QprfDZNbguReGi7AD2WQBVJkFAiW8v',
        '17ofPEarerDGkXTgceAt6psBTdT6Bbu2o86Yhoz5TueeK',
        '1D5qR8ZK4QwbXjxHjM3byLwGzdD6XEcEMtdSmJA5kFbP7',
        '1E2N2HqMDtuuCMgGPuzkaioDyPzsVdezqwLNa6RwExGve',
        '1C9EUXUBbCYqJtBMm85mvJERxLWVZ9e1x9tJYXr56Y3Sp'
      ],
      [
        '19rYMDuBUcZrZag71mhCU6U8RzAeTCyNkhEhLc7is9UgR',
        '17BPwpEuxEnDGTeJmiFfByYL8tWc6rAVAqb2f29T6Ryfk',
        '19JWRAxPKuyKm6k1xidq2G9VJAcJ4yuQ89MdekmZub75d',
        '14nXs88nvZjMK6vwFumS3wsU1JFqLSoxrHAtJbJPkjmNE',
        '1FaYSRUmB68aHEVoC1TTd9sedSQKM3viDfKXnM1CX1YCc',
        '13XybYf4V52cyo9pdGw8koD6XUKxzzLxVTZiTe4H8geS9',
        '1BgSQrrmQo5WjcMZrTQUvUdRFRqArMqgGYK5z1wsLrRwb',
        '12qB8hJKSTk34GcQEf8WDus6NDG9vj1LRwSRqwKPPoDDY',
        '12FPeXwJiJgpiXvyVx6nCHDoDeG8cxe5B2Bb5M539NTjo',
        '1AxaFwVn9DRHAn2aZUM6PBQD98yy15b9fWqkJSDiaCggp',
        '19NhDd9cUfxo88KxMooWfSVe3LLzvHyv4D2qQxXTzUSgQ',
        '17aojiDMy9fiFjBcqh5w8n6gAQQfPe4DxakuW9qSEnuMK',
        '19sJf2ZttkVuMFbnuUdYGjXArgfsLEc4Rpkx1vZLcjhex',
        '194gZAyuwboHGDkRXJ7aaUAH6fzvyDFSjQBbDrYdtV4GW',
        '16wN7GQpvDQi57PYgrezAMn6iN6tii9GTHwBUeoYCvqWk',
        '141rCdLSVvWX3EZKYfd4JTji2vtBDJSk6N2uFqKHDrNJN',
        '18cWAZhyzkoZoCnYeWDz1DCrX6PXX1Knat9aK2PRvhKFa',
        '1CJfYcCZyvcTvQUzgFeUiUT4sQr9aLfMpWxPDFRJm7vK9',
        '17U9ZgMUxw8vAXCey8YbQMkUmGYwGQpLah5mZrDpqejP3',
        '1FkBkho42CRFGp6WjvgHiqGV1rVAL1LM8kcP1tb6fptUg'
      ]
    ]

  107
```

</details>

nop33:
>>>
> LGTM! Great job 🚀
> 
> It's a first PR, so I don't dive into layout stuff for now. We'll keep that for the next one :)

@mvaivre, Does this mean I can merge? 🎉

mvaivre:
@nop33 yes sir.

PR #120, opened 2022-01-31
Deriving new addresses (part 2)

nop33:
- Better state management of addresses data and settings (initialize, save, update, refresh, clear)
- First draft of Address Details page
- First draft of Address Options modal

https://user-images.githubusercontent.com/1579899/151820929-572c0099-b192-4ed7-b1a0-0f4d097b6807.mp4


polarker:
The balance shown on the top left is the total amount of all addresses right? This might cause trouble for transactions, as users might think that the balance is enough on his main address.

Maybe we should do like Metamask and only show on the overview page with the balance of the main address. Another advantage of Metamask’s approach is that users always know that they are using one of the addresses when doing all of the operations.

nop33:
Discussion moved here: https://github.com/alephium/alephium-wallet/issues/71#issuecomment-1025933873

src/components/Buttons/ClipboardButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+  cursor: pointer;
+  color: ${({ theme }) => theme.font.secondary};
+`
+
+const StyledCheckIcon = styled(Check)`
+  margin-left: 10px;
+  color: ${({ theme }) => theme.font.primary};
+`
+
+export default ClipboardButton
```
mvaivre, line 73: Suggestion for this one, related to my other comment below. Curious to hear what you think about this way of wrapping the style + exporting + reusing shared style. 

```ts
import { Check, Clipboard } from 'lucide-react'
import { useEffect, useState } from 'react'
import ReactTooltip from 'react-tooltip'
import styled, { css } from 'styled-components'

import { useGlobalContext } from '../../contexts/global'

const ClipboardButton = ({ textToCopy, className }: { textToCopy: string; className?: string }) => {
  const [hasBeenCopied, setHasBeenCopied] = useState(false)
  const { setSnackbarMessage } = useGlobalContext()

  const handleClick = () => {
    navigator.clipboard
      .writeText(textToCopy)
      .catch((e) => {
        throw e
      })
      .then(() => {
        setHasBeenCopied(true)
      })
  }

  useEffect(() => {
    // Reset icon after copy
    if (hasBeenCopied) {
      ReactTooltip.rebuild()
      setSnackbarMessage({ text: 'Copied to clipboard!', type: 'info' })

      setInterval(() => {
        setHasBeenCopied(false)
      }, 3000)
    }
  }, [hasBeenCopied, setSnackbarMessage])

  return !hasBeenCopied ? (
    <StyledClipboardIcon className={className} size={15} data-tip={'Copy to clipboard'} onClick={handleClick} />
  ) : (
    <StyledCheckIcon className={className} size={15} />
  )
}

const BaseIconStyle = css`
  margin-left: 10px;
  color: ${({ theme }) => theme.font.secondary};
`

const StyledClipboardIcon = styled(Clipboard)`
  ${BaseIconStyle}
  cursor: pointer;
`

const StyledCheckIcon = styled(Check)`
  ${BaseIconStyle}
`

export default styled(ClipboardButton)``

```

src/components/Buttons/QRCodeButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+  margin-left: 10px;
+  cursor: pointer;
+  color: ${({ theme }) => theme.font.secondary};
+`
+
+const Tooltip = styled(ReactTooltip)`
+  opacity: 1 !important;
+`
+
+export default QRCodeButton
```
mvaivre, line 47: Same remark about style wrapping.

mvaivre (commented):
Review part 1.

src/pages/Wallet/AddressDetailsPage.tsx
```diff
@@ -0,0 +1,157 @@
...
+  { title: 'Direction' },
+  { title: 'Timestamp' },
+  { title: 'Address(es)' },
+  { title: 'Amount', align: 'end' }
+]
+
+const AddressDetailsPage = () => {
+  const [isAddressOptionsModalOpen, setIsAddressOptionsModalOpen] = useState(false)
+  const { addressesState } = useAddressesContext()
+  const { address } = useParams<{ address: string }>()
```
mvaivre, line 48: Tiny detail related to my comment in the address context file, we could use the `AddressHash` type for full explicitness.

src/pages/Wallet/AddressDetailsPage.tsx
```diff
@@ -0,0 +1,157 @@
...
+        <DataListRow>
+          <DataListCell>ALPH balance</DataListCell>
+          <DataListCell>
+            {addressData.details?.balance ? (
+              <AmountStyled value={BigInt(addressData.details.balance)} fadeDecimals />
+            ) : (
+              '-'
+            )}
+          </DataListCell>
+        </DataListRow>
```
mvaivre, line 108: Let's directly add the locked balance - as in the explorer - except that I would maybe display first the locked balance, and then the total balance:

- Locked ALPH balance (if any)
- Total ALPH balance

src/pages/Wallet/WalletHomePage.tsx
```diff
@@ -145,6 +146,11 @@ const WalletHomePage = () => {
     [client, loadedTxList, setLoadedTxList, setSnackbarMessage, wallet]
   )
 
+  const refreshData = async () => {
```
mvaivre, line 149: There's no awaited calls in this function, so no need for the `async` keyword.

mvaivre: Could be useful to properly await results though, in order to display the refresh animation while loading.

src/utils/addresses.ts
```diff
@@ -31,33 +34,27 @@ export const checkAddressValidity = (address: string) => {
   return match[0] === address && address
 }
 
-export const loadStoredAddressesInfoOfAccount = (username: string): AddressInfo[] => {
-  const storedAddressIndexes = localStorage.getItem(`${username}-address-info`)
+export const loadStoredAddressesIndexAndSettingsOfAccount = (username: string): AddressIndexAndSettings[] => {
```
mvaivre, line 37: That's a big fat name, dammn 🤩
We may add other keys in the future, so I'm thinking that the types `StoredAddressInfo` or `StoredAddressMetadata` would be more convenient? And thus `loadStoredAddressesInfo` or `loadStoredAddressesMetadata`?

src/utils/addresses.ts
```diff
@@ -31,33 +34,27 @@ export const checkAddressValidity = (address: string) => {
...
+  const data = localStorage.getItem(`${username}-address-indexes-and-settings`)
```
mvaivre, line 38: Same remark here. This will be tricky to migrate if we decide to store more metadata than index + settings one day.

mvaivre (commented):
Nice one!

src/components/Buttons/ClipboardButton.tsx
nop33: Nice catch, thanks for raising this topic!

First of all, the `StyledCheckIcon` has `color: ${({ theme }) => theme.font.primary};`, not `color: ${({ theme }) => theme.font.secondary};`, so the only common property is `margin-left: 10px;`.

To be honest, I don't rly like the approach of "base-styles" like this. The reason is that to understand the styles of the component I need to maintain things in my memory (base style) to construct the complete styles, because of the "flat" hierarchy of styles. Maybe it's because my brain is more used in the "traditional" nested SCSS patterns. I would much more prefer one of the 2 following options:

## Option 2
Wrapping React icon components in `styled()` so that we can then keep all styles in one block, similar to how we do with traditional SCSS with classes (`&.classname`) ([docs](https://styled-components.com/docs/advanced#caveat)):

```tsx
import { Check, Clipboard } from 'lucide-react'
...
const ClipboardButton = ({ textToCopy, className }: { textToCopy: string; className?: string }) => {
  ...
  if (!hasBeenCopied) {
    return <ClipboardIcon className={className} />
  } else {
    return <CheckIcon className={className} />
  }
}

const ClipboardIcon = styled(Clipboard)``
const CheckIcon = styled(Check)``

export default styled(ClipboardButton)`
  margin-left: 10px;

  &${ClipboardIcon} {
    cursor: pointer;
    color: ${({ theme }) => theme.font.secondary};
  }

  &${CheckIcon} {
    color: ${({ theme }) => theme.font.primary};
  }
`
```

## Option 3
Styled-components doesn't forbid us from using CSS classes:

```tsx
import { Check, Clipboard } from 'lucide-react'
...
const ClipboardButton = ({ textToCopy, className }: { textToCopy: string; className?: string }) => {
  ...
  if (!hasBeenCopied) {
    return (
      <Clipboard className={`${className} clipboard`} />
    )
  } else {
    return <Check className={`${className} check`} />
  }
}

export default styled(ClipboardButton)`
  margin-left: 10px;

  &.clipboard {
    cursor: pointer;
    color: ${({ theme }) => theme.font.secondary};
  }

  &.check {
    color: ${({ theme }) => theme.font.primary};
  }
`
```

My order of preference is:
- Option 3: For me this is the cleanest solution. It does not require unnecessary lines of code to wrap React components with `styled()`, while using plain old vanilla CSS/SCSS features. It also looks visually nice IMO :) Moreover, it only needs 11 lines of code (the other 2 options require 15).
- Option 2
- Option 1: The one you proposed.

src/components/Buttons/QRCodeButton.tsx
nop33: Hm, I don't think your previous remark applies here, because we have 2 individual components, wrapped with a fragment. How would you change this code?

src/contexts/addresses.tsx
```diff
@@ -0,0 +1,172 @@
...
+import {
+  AddressSettings,
+  loadStoredAddressesIndexAndSettingsOfAccount,
+  storeAddressIndexAndSettingsOfAccount
+} from '../utils/addresses'
+import { useGlobalContext } from './global'
+
+export interface AddressesContextProps {
+  saveNewAddress: (address: AddressState) => void
+  addressesState: Map<string, AddressState>
```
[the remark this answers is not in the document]
nop33: OK cool! Something else I want to point out related to this: It'd be nice if we start referring to the address hash as `AddressHash` instead of `address`. It'd make sense to use the variable name `address` when referring to objects containing many info about the address.

src/contexts/addresses.tsx
```diff
@@ -0,0 +1,172 @@
...
+  overrideContextValue
+}) => {
+  const [addressesState, setAddressesState] = useState<Map<string, AddressState>>(new Map())
+  const { currentUsername, wallet, client } = useGlobalContext()
+
+  const updateAddressState = useCallback((newAddressState: AddressState) => {
+    setAddressesState((prevState) => {
+      const newAddressesState = new Map(prevState)
+      newAddressesState.set(newAddressState.hash, newAddressState)
+      console.log('updating state of address: ', newAddressState.hash)
```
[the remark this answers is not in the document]
nop33: haha, sure! I am planning to remove this log once I am satisfied that the state system works well (I know, I should use use proper dev tools 💩 )

src/contexts/addresses.tsx
```diff
@@ -0,0 +1,172 @@
...
+  // Clean state when locking the wallet
+  useEffect(() => {
+    if (wallet === undefined) {
+      setAddressesState(new Map())
+    }
+  }, [wallet])
+
+  return (
+    <AddressesContext.Provider
+      value={merge(
```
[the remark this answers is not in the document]
nop33: Hm, it's not an array, but a Map. Is Map also not considered a deep object (its key is a string and its value is deep objects)? I'll check it.

src/utils/addresses.ts
nop33: Ah, yes, good old naming problem! Thanks fo the `StoredAddressMetadata`, I really like it! I didn't like my initial `AddressInfo` idea, because there is a `AddressInfo` type in `alephium-js`. Metadata sounds great.

src/pages/Wallet/WalletHomePage.tsx
nop33: Yeah, I need to work on loading animations in general.

src/components/Buttons/ClipboardButton.tsx
mvaivre: Interesting discussion :) 
I'm generally against defining static classnames as we're using styled-components to avoid just that. 
BUT in this case - and as we're anyway setting the `className` prop in both our child components - I believe that I prefer option 3 as well, less convoluted.

src/contexts/addresses.tsx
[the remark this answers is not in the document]
mvaivre: Fully agreed 💯

src/components/Buttons/QRCodeButton.tsx
mvaivre: If our goal is to systematically allow for subsequent styling (by wrapping the styled component), then I'd do:

```ts
const QRCodeButton = ({ textToEncode, className }: { textToEncode: string; className?: string }) => {
  return (
    <>
      <StyledClipboardIcon className={className} data-tip data-for="qr-code-tooltip" data-event="click" size={15} />
      <Tooltip id="qr-code-tooltip" backgroundColor="black" globalEventOff="click">
        <QRCode size={150} value={textToEncode} bgColor="black" fgColor="white" />
      </Tooltip>
    </>
  )
}

const StyledClipboardIcon = styled(QrCode)`
  margin-left: 10px;
  cursor: pointer;
  color: ${({ theme }) => theme.font.secondary};
`

const Tooltip = styled(ReactTooltip)`
  opacity: 1 !important;
`

export default styled(QRCodeButton)``
```

It encourages us to always pass the `className` down (linked to our previous discussion, last week I believe).

I'm still not 100% sure about this, it's a tradeoff between coherence (always being able to extend style) and added complexity.

src/contexts/addresses.tsx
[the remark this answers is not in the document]
mvaivre: You're right, sorry. But my point is still partially valid: if only the map needs a deep merge, we could restrict the operation to this field only.

But honestly, this might be over engineering at this point, so maybe just something to keep in mind in the future :)

src/components/Buttons/QRCodeButton.tsx
nop33: I see, you're right.

This is not needed: `export default styled(QRCodeButton)` and can be replaced by simply `export default QRCodeButton` (or even better, bringing it up to `export const QRCodeButton = ({...`)

Updating...

src/pages/Wallet/WalletHomePage.tsx
nop33: I will remove the `async` keyword for now, and in a next PR I'll introduce some loading animations, after deciding with you how we want to approach it (loading animation per component that displays data, or a global loader?)

nop33:
@mvaivre I believe I addressed all your valuable feedback. Re-requested review :D Can't wait to open the next one, muahaha 😈 😆

mvaivre (approved):
🚀

PR #121, opened 2022-02-02
Derive new addresses (part 3)

nop33:
- Installing Electron React dev tools (the extension needs to already be installed in Chrome/Brave).
   image
- Fetching and displaying confirmed transactions (pending transactions PR coming up)

   image
   image

### DIFF
--- a/src/components/Badge.tsx
+++ b/src/components/Badge.tsx
@@ -28,14 +28,14 @@ interface BadgeProps {
   prefix?: string
 }
 
-let Badge = ({ content, className, amount, prefix }: BadgeProps) => (
+const Badge = ({ content, className, amount, prefix }: BadgeProps) => (
   <div className={className}>
     {prefix && <span>{prefix}</span>}
     {amount && content ? <Amount value={BigInt(content)} fadeDecimals /> : content}
   </div>
 )
 
-Badge = styled(Badge)`
+export default styled(Badge)`
   color: ${({ type, theme }) => theme.badge.font[type]};
   text-align: center;
   border-radius: 3px;
@@ -43,5 +43,3 @@ Badge = styled(Badge)`
   float: left;
   white-space: nowrap;
 `
-
-export default Badge
--- a/src/components/Badge.tsx
+++ b/src/components/Badge.tsx
@@ -20,10 +20,8 @@ import styled, { DefaultTheme } from 'styled-components'
 
 import Amount from './Amount'
 
-type BadgeType = 'plus' | 'minus' | 'neutral' | 'neutralHighlight'
-
 interface BadgeProps {
-  type: BadgeType
+  type: keyof DefaultTheme['badge']['font']
   content: string | number | undefined
   className?: string
   amount?: boolean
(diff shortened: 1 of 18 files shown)


src/components/Badge.tsx
```diff
@@ -33,40 +31,13 @@ interface BadgeProps {
...
-      color = theme.name === 'dark' ? 'rgba(255, 255, 255, 0.8)' : 'rgba(255, 255, 255, 1)'
-      break
-    case 'neutralHighlight':
-      backgroundColor = 'rgba(101, 16, 247, 1)'
-      color = 'rgba(255, 255, 255, 1)'
-  }
-
-  return { backgroundColor, color }
-}
-
```
nop33, line 64: I moved all the color definitions in the theme, let me know if you disagree

src/pages/Wallet/AddressDetailsPage.tsx
```diff
@@ -126,6 +174,31 @@ const AddressDetailsPage = () => {
   )
 }
 
+const IOList = ({ currentAddress, isOut, outputs, inputs, timestamp }: IOListProps) => {
```
nop33, line 177: I am reusing the component of the wallet home page, which will eventually be adjusted, and this will move to a new component to keep the code dry.

src/style/styled.d.ts
```diff
@@ -49,5 +49,19 @@ declare module 'styled-components' {
       valid: string
       highlightGradient: string
     }
+    badge: {
+      font: {
+        plus: string
```
nop33, line 54: I did not use the bright new colors from Figma, because personally, I find the "pastel'-er version of the green and red more pleasing. However, I trust your opinion more! We can easily change the colors in the theme settings.

mvaivre: Fine by me, I was still iterating. I'll do a final tweaking pass at the very end 🎨

mvaivre:
I forgot a little thing in the previous PR: it would be great to show the text cursor (blinker) in the `ColoredLabelInput`. 
Tricky, but feasible I believe :) Do you prefer to take a look at this in this PR, or do you prefer to integrate this in the next one?

nop33:
Nice catch. I'll try to figure it out on this one.

nop33:
@mvaivre done!

nop33:
https://user-images.githubusercontent.com/1579899/152171369-e5aa9b73-7673-4b9b-a490-89351e202981.mp4

mvaivre:
> Kapture.2022-02-02.at.15.18.23.mp4

Credible!!! <3

src/components/Badge.tsx
```diff
@@ -33,40 +31,13 @@ interface BadgeProps {
...
 Badge = styled(Badge)`
-  background-color: ${({ type, theme }) => getBadgeColor(type, theme).backgroundColor};
-  color: ${({ type, theme }) => getBadgeColor(type, theme).color};
+  color: ${({ type, theme }) => theme.badge.font[type]};
```
mvaivre, line 39: A badge without a background color isn't really a badge anymore I would say, but just a stylised string.
Your `DarkLabel` (used for from/to labels) looks more like a badge.

That's a purely style related discussion we should have. In the explorer, I'm using badges for amounts in tables + many other things. In the wallet, I tried another approach to make the UI feel less cluttered (we will have more content to display than in the explorer). Which approach do you prefer?

IMO I tend to think that badges should be used carefully for things we wanna highlight or label, be it for address labels, important status, total amounts etc.. Badges can be with or without stroke, depending on the visual importance we wanna give.

We would keep repeating amounts (in table) or other inline data as simple strings (and make the change in the explorer as well).

As a side note, I'd like to reduce a bit the use of colors accros my mockups (for instance, not use green and red for every amount diff in a table). I'll experiment around that.

**TLDR: Let's keep the badge component, but let's not use it for TX "direction" label or table amounts. :) And I need to keep iterating.**

src/pages/Wallet/AddressDetailsPage.tsx
```diff
@@ -162,4 +235,15 @@ const MainAddress = styled.div`
   position: absolute;
 `
 
+const DarkLabel = styled.span`
```
mvaivre, line 238: This could actually use the `Badge` component.

src/components/Badge.tsx
mvaivre: I'm making changes to the mockups right now, so don't spend too much time on this. This was mostly to keep you up to date with my thoughts 🧠

src/contexts/addresses.tsx
```diff
@@ -34,13 +34,17 @@ import { useGlobalContext } from './global'
...
   publicKey: string
   privateKey: string
   group: number
   index: number
   settings: AddressSettings
   details?: AddressInfo
+  transactions?: {
+    confirmed: Transaction[]
+  }
+  lastUsed?: number
```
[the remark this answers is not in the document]
nop33: yes, makes sense!

src/components/Badge.tsx
nop33: Thanks for sharing your thoughts. I agree with all of them.

src/pages/Wallet/AddressDetailsPage.tsx
nop33: Updated. I am sure we'll update it again once we re-iterate over the styles of labels and badges.

nop33:
@mvaivre, All comments were addressed! 🙏

mvaivre:
LGTM then 🚀

PR #124, opened 2022-02-09
Implement display pending transactions and sweep address feature

nop33:
- Massively improved code for managing state of fetched data for each address using ES6 classes
- Implemented display of pending transactions of current network
- Implemented address sweeping feature

### 🗣️ sound on

https://user-images.githubusercontent.com/1579899/153241237-be7c1274-0dde-462a-a5c3-903b788a7735.mp4


mvaivre:
Verrrry nice! (looking at the video). Great job @nop33  👏  
Onto the code review now.

src/contexts/addresses.tsx
```diff
@@ -32,55 +32,124 @@ import {
...
+  fromAddress: string
+  toAddress: string
+  amount: string
+  timestamp: number
+  type: TransactionType
+  network: NetworkType
+}
+
 export type AddressHash = string
 export type TimeInMs = number
```
mvaivre, line 48: This should probably be defined out of the `addresses` file, as it can be used globally.

src/contexts/addresses.tsx
```diff
@@ -32,55 +32,124 @@ import {
...
 
-type AddressState = {
-  network?: NetworkType
-  hash: AddressHash
-  publicKey: string
-  privateKey: string
-  group: number
-  index: number
+export class Address {
+  readonly hash: AddressHash
```
mvaivre, line 51: Nice.

mvaivre (commented):
Looks really good to me so far! Love the `Address` class with its handy methods.

polarker:
If the current address is not labelled, what will you show at the top-right? In the video, the label `vault` is shown

nop33:
> If the current address is not labelled, what will you show at the top-right? In the video, the label vault is shown

image

nop33:
@mvaivre All comments were addressed 🚀

src/contexts/addresses.tsx
```diff
@@ -346,17 +338,26 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
...
 
   return (
     <AddressesContext.Provider
       value={merge(
         {
-          addressesState: getAddressesStateOfCurrentNetwork(),
+          addresses: addressesOfCurrentNetwork,
+          mainAddress: addressesOfCurrentNetwork.find((address) => address.settings.isMain),
+          getAddress,
+          setAddress,
```
mvaivre, line 356: It would be great to make it clear when we're storing something in the local storage or not.
Also, we need to define if we want to specify Address**Data** all the time or not.

I would suggest to rename:
 `updateAddressSettings` into `updateAndStoreAddressSettings`
`refreshAddressesData` into `fetchAndStoreExistingAddresses`
`saveNewAddress` into `fetchAndStoreNewAddress`

Those are only propositions, not sure about all of them. WDYT?

src/contexts/addresses.tsx
```diff
@@ -90,161 +159,105 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
...
-            })
-          }
+      for (const address of addressesStateToRefresh) {
+        await address.fetchDetails(client)
+        await address.fetchConfirmedTransactions(client)
+
+        const initialNumberOfPendingTransactions = address.transactions.pending.length
+
+        // Filter pending addresses and remove the ones that are now confirmed
+        address.updatePendingTransactions()
```
mvaivre, line 233: Mhh, I don't know what to think about this. Calling a function which mutate some properties and doesn't return anything (when you need a comment to explain what a function does, it's usually that something could be more explicit :D). I'm starting to wonder if we shouldn't adopt a more immutable and functional way of coding instead of always mutate the same address object...

I'll do some research and see if I can propose something different. In the meantime, let's continue like that.

src/pages/Wallet/NewAddressPage.tsx
```diff
@@ -63,39 +61,33 @@ const NewAddressPage = () => {
...
-        },
-        transactions: {}
-      })
-      if (isMainAddress && currentMainAddress && currentMainAddress.index !== newAddressData.addressIndex) {
-        saveNewAddress({
-          ...currentMainAddress,
-          settings: {
-            ...currentMainAddress.settings,
-            isMain: false
+      saveNewAddress(
```
mvaivre, line 64: Shouldn't we await this and stop the flow if the fetch fails?

mvaivre (commented):
Some more thoughts. My main "concern" is related to how much we mutate the address objects, which IMO makes the code quite opaque sometimes...
It's certainly a matter of personal preference, but I wonder if we could look into use immutable and more functional way where we return new objects (a la redux) instead. 
To be discussed and investigated.

src/pages/Wallet/NewAddressPage.tsx
nop33: I think the user should be able to generate and locally save new addresses, even if the internet connection fails (fetch fails). WDYT?

src/contexts/addresses.tsx
nop33: > I would suggest to rename:  
> `updateAddressSettings` into `updateAndStoreAddressSettings`  
> `refreshAddressesData` into `fetchAndStoreExistingAddresses`  
> `saveNewAddress` into `fetchAndStoreNewAddress`  


Hm, that's how I had initially named it. However, after a few code iterations, I thought it'd be nice to create a level of abstraction for the address context API. It made sense to me that the function that takes care of everything that needs to be done when generating a new adddress (store in local storage and fetch data from API) is called `save`.

Same goes for using `updateAddressSettings` that uses the word `update` instead of explicitly listing everything the function does in its name.

Same for `refreshAddressesData`.

I don't have a strong opinion on this, although I do prefer the names I chose. I fully understand your opinion that you prefer the function names to be as explicit as possible. My only argument is to have a (imo) cleaner context API.

```
          addresses,
          mainAddress,
          getAddress,
          setAddress,
          saveNewAddress,
          updateAddressSettings,
          refreshAddressesData,
          isLoadingData
```

VS

```
          addresses,
          mainAddress,
          getAddress,
          setAddress,
          fetchAndStoreNewAddress,
          updateAndStoreAddressSettings,
          fetchAndStoreExistingAddresses,
          isLoadingData
```

> It would be great to make it clear when we're storing something in the local storage or not.

Lastly, if we were to use the word `store` only for when we store things in local storage we should then agree to use the word `set` when storing things in the state. Based on this, we would then need to update the names to sth like this:

- `storeFetchAndSetNewAddress` (instead of `saveNewAddress` and `fetchAndStoreNewAddress`)
- `storeAndSetAddressSettings` (instead of `updateAddressSettings` and `updateAndStoreAddressSettings`)
- `fetchAndSetExistingAddresses` (instead of `refreshAddressesData` and `fetchAndStoreExistingAddresses`)

> Also, we need to define if we want to specify AddressData all the time or not.

"Fetching address data" sounds better to me than "fetch address", but again, no strong opinion.

---

I look forward to hearing your thoughts on all these.

src/contexts/addresses.tsx
nop33: Also here I [initially implemented it by creating new objects](https://github.com/alephium/alephium-wallet/blob/master/src/contexts/addresses.tsx#L188-L197), but I felt that the code becomes very confusing, so I replaced it with instances of the `Address` class (instead of an object) and always using the `setAddress` every time an `Address` instance is being modified. It feels so much cleaner to code this way for me. Happy to discuss to find a better approach.

src/pages/Wallet/NewAddressPage.tsx
mvaivre: Would the "address data" be fetched properly (== clean state) once the internet connection is back? If yes, then yes, all good!

nop33: It would be fetched property when the internet connection is back and one of the following happens:
- the user clicks the "refresh" button
- the wallet gets locked and unlocked

Basically, there is no implemented mechanism to set an interval to keep fetching address data if a request fails.

src/contexts/addresses.tsx
nop33: >  I'm starting to wonder if we shouldn't adopt a more immutable and functional way

Here you mean "should", right?

src/contexts/addresses.tsx
mvaivre: If you're fully comfortable with this level of abstraction, then I'm fine with it. My initial reaction was caused by the fact that the methods' names were not enough for me to clearly understand what these do: memory storage vs. local storage; remote fetch vs. offline... Just wanted to have your opinion on this, so thanks for that.

But it's true that the proposed naming is a mouthful! I wonder what Lee would say about this hehe ;) 

**TLDR: I trust your guts on this one.**

nop33:
Does that mean we're good to merge? :)

mvaivre (approved):

PR #128, opened 2022-02-15
Derive new addresses features (pagination, advanced ops, send modal)

nop33:
- Implemented transaction list pagination (making sure it works well even when there are pending txs)
- Implemented advanced operations:
   - UTXO consolidation
   - 1 address per group generation
- Fully functional send modal that works with multiple addresses

### 🗣️ sound on!

https://user-images.githubusercontent.com/1579899/154123435-d25597a6-9937-4330-9be2-07b90d894c24.mp4


src/contexts/addresses.tsx
```diff
@@ -125,6 +159,60 @@ export class Address {
...
+    const totalAmountOfPendingTxs = pendingSweep
+      ? this.availableBalance
+      : this.transactions.pending.reduce((acc, tx) => (tx.amount ? acc + BigInt(tx.amount) : acc), BigInt(0))
+    this.availableBalance = this.availableBalance - totalAmountOfPendingTxs
+  }
+
+  async buildSweepTransactions(
+    client: Client,
+    toHash: AddressHash
+  ): Promise<{
```
mvaivre, line 174: Here we describe the returned type explicitly, but we don't do that for `signAndSendTransaction` below.. or any of the other class' methods actually.

Should we remove this? (the IDE is clever enough in this case, and there are no possible confusion between string / number / bigint like we've seen in the discussion about amounts)

src/contexts/addresses.tsx
```diff
@@ -93,26 +98,55 @@ export class Address {
   }
 
   async fetchDetails(client: Client) {
-    if (!client) return
+    if (!client) throw new Error('No client provided')
```
mvaivre, line 101: We pass `client` to each methods and return an error if none are provided. Do you see a way to make it dryer?

src/contexts/addresses.tsx
```diff
@@ -181,6 +271,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
```
mvaivre, line 271: Mhh I was thinking about the fact that we store the `network` name in the address class... It's even used in the key identifying addresses in the local storage. What happens with the `custom` network, which could represent different networks? The network `name` could stay the same (=== custom) but the client (and the address amounts, tx etc.) could actually be different..
Could this be an issue?

src/contexts/addresses.tsx
mvaivre: Also, the client depends solely on the `network`, which is stored in the client class. 
Please see the comment below about the `custom` network that could cause issues.

src/pages/Wallet/SendModalCheckTransaction.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { SendTransactionData } from './SendModal'
+
+interface SendModalCheckTransactionProps {
+  data: SendTransactionData
+  fees: bigint
+  onSend: () => void
+  onCancel: () => void
+}
+
+const SendModalCheckTransaction = ({ data, fees, onSend, onCancel }: SendModalCheckTransactionProps) => {
```
mvaivre, line 33: Could we maybe store those "step" pages in `pages/Wallet/SendModal/...` ? WDYT?

mvaivre:
General remark, I don't see any default address for most of my current accounts. I get this:

image

src/contexts/addresses.tsx
nop33: Yeah, this would be an issue if you are planning to use 2 different custom networks at the same time (same time = same login session). Need to think about it a bit and come back with a solution (even though this is an extreme corner case)

nop33:
> General remark, I don't see any default address for most of my current accounts. I get this:

That's weird. It means that the `balance` or the `lockedBalance` that you receive from the API is `undefined`, which should never be the case.

src/pages/Wallet/SendModalCheckTransaction.tsx
nop33: Currently we are mixing components that are modals with components that are pages. For a while now I think that we should split `pages` and `modals`. Modals can be displayed on top of every page. Pages take hold of the whole visual area of the app (except top app bar and side bar). What would you think of the following structure?

```
- components
- pages
- modals
```

or

```
- components
  - modals
- pages
```

or even

```
- components
  - modals
  - pages
```

src/contexts/addresses.tsx
nop33: I think I found an elegant solution to this. Instead of passing `Client` as an argument to `Address` functions, I think it makes much more semantic sense to pass `Address` to `Client` functions. See the following diff to understand what I mean. I feel this is a much cleaner approach, keeping concerns of storing address state and querying the API separately. Let me know your thoughts.

<details><summary>Click here plz</summary>

```diff
diff --git a/src/contexts/addresses.tsx b/src/contexts/addresses.tsx
index 74ffd53..7ac5237 100644
--- a/src/contexts/addresses.tsx
+++ b/src/contexts/addresses.tsx
@@ -32,7 +32,7 @@ import { getHumanReadableError } from '../utils/api'
 import { NetworkType } from '../utils/settings'
 import { Client, useGlobalContext } from './global'
 
-type TransactionType = 'consolidation' | 'transfer' | 'sweep'
+export type TransactionType = 'consolidation' | 'transfer' | 'sweep'
 
 type SimpleTx = {
   txId: string
@@ -96,58 +96,6 @@ export class Address {
     return `${this.settings.isMain ? '★ ' : ''}${this.displayName()}`
   }
 
-  async fetchDetails(client: Client) {
-    if (!client) throw new Error('No client provided')
-    console.log('⬇️ Fetching address details: ', this.hash)
-
-    const { data } = await client.explorer.getAddressDetails(this.hash)
-    this.details = data
-
-    if (data.balance) this.availableBalance = BigInt(data.balance)
-    if (data.lockedBalance) this.availableBalance -= BigInt(data.lockedBalance)
-
-    return data
-  }
-
-  async fetchConfirmedTransactions(client: Client, page = 1) {
-    if (!client) throw new Error('No client provided')
-    console.log(`⬇️ Fetching page ${page} of address confirmed transactions: `, this.hash)
-
-    const { data } = await client.explorer.getAddressTransactions(this.hash, page)
-
-    const isInitialData = page === 1 && data.length > 0 && this.transactions.confirmed.length === 0
-    const latestTxHashIsNew =
-      page === 1 &&
-      data.length > 0 &&
-      this.transactions.confirmed.length > 0 &&
-      data[0].hash !== this.transactions.confirmed[0].hash
-
-    if (isInitialData) {
-      this.transactions.confirmed = data
-      this.transactions.loadedPage = 1
-    } else if (latestTxHashIsNew || page > 1) {
-      const newTransactions = data.filter(
-        (tx) => !this.transactions.confirmed.find((confirmedTx) => confirmedTx.hash === tx.hash)
-      )
-      this.transactions.confirmed =
-        page > 1
-          ? this.transactions.confirmed.concat(newTransactions)
-          : newTransactions.concat(this.transactions.confirmed)
-      if (page > 1) {
-        this.transactions.loadedPage = page
-      }
-    }
-
-    this.lastUsed = this.transactions.confirmed.length > 0 ? this.transactions.confirmed[0].timestamp : this.lastUsed
-
-    return data
-  }
-
-  async fetchConfirmedTransactionsNextPage(client: Client) {
-    if (!client) throw new Error('No client provided')
-    await this.fetchConfirmedTransactions(client, this.transactions.loadedPage + 1)
-  }
-
   addPendingTransaction(transaction: SimpleTx) {
     console.log('🔵 Adding pending transaction sent from address: ', transaction.fromAddress)
 
@@ -168,47 +116,6 @@ export class Address {
       : this.transactions.pending.reduce((acc, tx) => (tx.amount ? acc + BigInt(tx.amount) : acc), BigInt(0))
     this.availableBalance = this.availableBalance - totalAmountOfPendingTxs
   }
-
-  async buildSweepTransactions(client: Client, toHash: AddressHash) {
-    if (!client) throw new Error('No client provided')
-
-    const { data } = await client.clique.transactionConsolidateUTXOs(this.publicKey, this.hash, toHash)
-    const fees = data.unsignedTxs.reduce((acc, tx) => acc + BigInt(tx.gasPrice) * BigInt(tx.gasAmount), BigInt(0))
-
-    return {
-      unsignedTxs: data.unsignedTxs,
-      fees
-    }
-  }
-
-  async signAndSendTransaction(
-    client: Client,
-    txId: string,
-    unsignedTx: string,
-    toHash: AddressHash,
-    type: TransactionType,
-    network: NetworkType,
-    amount?: bigint
-  ) {
-    if (!client) throw new Error('No client provided')
-
-    const signature = client.clique.transactionSign(txId, this.privateKey)
-    const response = await client.clique.transactionSend(toHash, unsignedTx, signature)
-
-    if (response.data) {
-      this.addPendingTransaction({
-        txId: response.data.txId,
-        fromAddress: this.hash,
-        toAddress: toHash,
-        timestamp: new Date().getTime(),
-        amount,
-        type,
-        network
-      })
-    }
-
-    return response.data
-  }
 }
 
 export type AddressesStateMap = Map<AddressHash, Address>
@@ -311,8 +218,8 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
 
       for (const address of addressesStateToRefresh) {
         try {
-          await address.fetchDetails(client)
-          await address.fetchConfirmedTransactions(client)
+          await client.fetchAddressDetails(address)
+          await client.fetchAddressConfirmedTransactions(address)
 
           const initialNumberOfPendingTransactions = address.transactions.pending.length
 
@@ -348,7 +255,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
   const fetchAddressTransactionsNextPage = async (address: Address) => {
     if (!client) return
     setIsLoadingData(true)
-    await address.fetchConfirmedTransactionsNextPage(client)
+    await client.fetchAddressConfirmedTransactionsNextPage(address)
     setIsLoadingData(false)
   }
 
diff --git a/src/pages/Wallet/AddressSweepModal.tsx b/src/pages/Wallet/AddressSweepModal.tsx
index 2acf3e5..d72f602 100644
--- a/src/pages/Wallet/AddressSweepModal.tsx
+++ b/src/pages/Wallet/AddressSweepModal.tsx
@@ -60,7 +60,7 @@ const AddressSweepModal = ({ sweepAddress, onClose, onSuccessfulSweep }: Address
       if (!client || !sweepAddresses.from || !sweepAddresses.to) return
       setIsLoading(true)
       try {
-        const { unsignedTxs, fees } = await sweepAddresses.from.buildSweepTransactions(client, sweepAddresses.to.hash)
+        const { unsignedTxs, fees } = await client.buildSweepTransactions(sweepAddresses.from, sweepAddresses.to.hash)
         setBuiltUnsignedTxs(unsignedTxs)
         setFee(fees)
       } catch (e) {
@@ -81,8 +81,8 @@ const AddressSweepModal = ({ sweepAddress, onClose, onSuccessfulSweep }: Address
     setIsLoading(true)
     try {
       for (const { txId, unsignedTx } of builtUnsignedTxs) {
-        const txSendResp = await sweepAddresses.from.signAndSendTransaction(
-          client,
+        const txSendResp = await client.signAndSendTransaction(
+          sweepAddresses.from,
           txId,
           unsignedTx,
           sweepAddresses.from.hash,
diff --git a/src/pages/Wallet/SendModal.tsx b/src/pages/Wallet/SendModal.tsx
index 480f835..b15b732 100644
--- a/src/pages/Wallet/SendModal.tsx
+++ b/src/pages/Wallet/SendModal.tsx
@@ -103,7 +103,7 @@ const SendModal = ({ title, onClose }: SendModalProps) => {
 
       try {
         if (sweep) {
-          const { unsignedTxs, fees } = await fromAddress.buildSweepTransactions(client, toAddress)
+          const { unsignedTxs, fees } = await client.buildSweepTransactions(fromAddress, toAddress)
           setSweepUnsignedTxs(unsignedTxs)
           setFees(fees)
         } else {
@@ -153,7 +153,7 @@ const SendModal = ({ title, onClose }: SendModalProps) => {
 
       setIsLoading(true)
       const { fromAddress } = transactionData
-      const { unsignedTxs, fees } = await fromAddress.buildSweepTransactions(client, fromAddress.hash)
+      const { unsignedTxs, fees } = await client.buildSweepTransactions(fromAddress, fromAddress.hash)
       setSweepUnsignedTxs(unsignedTxs)
       setFees(fees)
       setIsLoading(false)
@@ -176,8 +176,8 @@ const SendModal = ({ title, onClose }: SendModalProps) => {
           const transactionType = consolidationRequired ? 'consolidation' : 'sweep'
 
           for (const { txId, unsignedTx } of sweepUnsignedTxs) {
-            const data = await fromAddress.signAndSendTransaction(
-              client,
+            const data = await client.signAndSendTransaction(
+              fromAddress,
               txId,
               unsignedTx,
               sendToAddress,
@@ -190,8 +190,8 @@ const SendModal = ({ title, onClose }: SendModalProps) => {
             }
           }
         } else {
-          const data = await fromAddress.signAndSendTransaction(
-            client,
+          const data = await client.signAndSendTransaction(
+            fromAddress,
             unsignedTxId,
             unsignedTransaction,
             toAddress,
diff --git a/src/utils/api-clients.ts b/src/utils/api-clients.ts
index 394666f..82793ec 100644
--- a/src/utils/api-clients.ts
+++ b/src/utils/api-clients.ts
@@ -18,7 +18,8 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { CliqueClient, ExplorerClient } from 'alephium-js'
 
-import { loadStoredSettings, Settings } from './settings'
+import { Address, AddressHash, TransactionType } from '../contexts/addresses'
+import { loadStoredSettings, NetworkType, Settings } from './settings'
 
 export async function createClient(settings?: Settings['network']) {
   const loadedSettings = settings || loadStoredSettings().network
@@ -42,7 +43,102 @@ export async function createClient(settings?: Settings['network']) {
     // Init clients
     await cliqueClient.init(isMultiNodesClique)
 
-    return { clique: cliqueClient, explorer: explorerClient }
+    const fetchAddressDetails = async (address: Address) => {
+      console.log('⬇️ Fetching address details: ', address.hash)
+
+      const { data } = await explorerClient.getAddressDetails(address.hash)
+      address.details = data
+
+      if (data.balance) address.availableBalance = BigInt(data.balance)
+      if (data.lockedBalance) address.availableBalance -= BigInt(data.lockedBalance)
+
+      return data
+    }
+
+    const fetchAddressConfirmedTransactions = async (address: Address, page = 1) => {
+      console.log(`⬇️ Fetching page ${page} of address confirmed transactions: `, address.hash)
+
+      const { data } = await explorerClient.getAddressTransactions(address.hash, page)
+
+      const isInitialData = page === 1 && data.length > 0 && address.transactions.confirmed.length === 0
+      const latestTxHashIsNew =
+        page === 1 &&
+        data.length > 0 &&
+        address.transactions.confirmed.length > 0 &&
+        data[0].hash !== address.transactions.confirmed[0].hash
+
+      if (isInitialData) {
+        address.transactions.confirmed = data
+        address.transactions.loadedPage = 1
+      } else if (latestTxHashIsNew || page > 1) {
+        const newTransactions = data.filter(
+          (tx) => !address.transactions.confirmed.find((confirmedTx) => confirmedTx.hash === tx.hash)
+        )
+        address.transactions.confirmed =
+          page > 1
+            ? address.transactions.confirmed.concat(newTransactions)
+            : newTransactions.concat(address.transactions.confirmed)
+        if (page > 1) {
+          address.transactions.loadedPage = page
+        }
+      }
+
+      address.lastUsed =
+        address.transactions.confirmed.length > 0 ? address.transactions.confirmed[0].timestamp : address.lastUsed
+
+      return data
+    }
+
+    const fetchAddressConfirmedTransactionsNextPage = async (address: Address) => {
+      await fetchAddressConfirmedTransactions(address, address.transactions.loadedPage + 1)
+    }
+
+    const buildSweepTransactions = async (address: Address, toHash: AddressHash) => {
+      const { data } = await cliqueClient.transactionConsolidateUTXOs(address.publicKey, address.hash, toHash)
+      const fees = data.unsignedTxs.reduce((acc, tx) => acc + BigInt(tx.gasPrice) * BigInt(tx.gasAmount), BigInt(0))
+
+      return {
+        unsignedTxs: data.unsignedTxs,
+        fees
+      }
+    }
+
+    const signAndSendTransaction = async (
+      address: Address,
+      txId: string,
+      unsignedTx: string,
+      toHash: AddressHash,
+      type: TransactionType,
+      network: NetworkType,
+      amount?: bigint
+    ) => {
+      const signature = cliqueClient.transactionSign(txId, address.privateKey)
+      const response = await cliqueClient.transactionSend(toHash, unsignedTx, signature)
+
+      if (response.data) {
+        address.addPendingTransaction({
+          txId: response.data.txId,
+          fromAddress: address.hash,
+          toAddress: toHash,
+          timestamp: new Date().getTime(),
+          amount,
+          type,
+          network
+        })
+      }
+
+      return response.data
+    }
+
+    return {
+      clique: cliqueClient,
+      explorer: explorerClient,
+      fetchAddressDetails,
+      fetchAddressConfirmedTransactions,
+      fetchAddressConfirmedTransactionsNextPage,
+      buildSweepTransactions,
+      signAndSendTransaction
+    }
   } catch (error) {
     console.error(error)
     return false
```
</details>

src/contexts/addresses.tsx
nop33: My proposal:

The information of the addresses depends only on the blockchain (each address will have a different balance on `mainnet` than on `testnet` and on each of the infinite numbers of `custom` networks). So when we store information about an address in the state of the app, we need to associate this info with the blockchain that this info corresponds to. The solution I found so far was to simply store the _name_ of the blockchain/network(`['mainnet','testnet','local','custom']`) along with the address's information.

A better approach to be able to support the infinite number of custom blockchains would be to store the `nodeHost` URL instead of the network name. However, I think this is a bit of overengineering.

Our app's state only exists for as long as the user is logged in. It is flushed on user logout. Do we really see a use case where someone would need to use the wallet with 2 different custom settings within the same login session? In the extreme case that someone wants to do that (switch from one custom network to another custom network while being logged in) I propose that we only keep the state of the latest custom network, flushing and re-initializing every time the custom network changes. The only drawback of this approach would be that the pending transactions of the 1st custom network would not anymore be visible when switching back to it from the 2nd custom network. Everything else would work as expected.

Lemme know WYTAI (😆).

mvaivre: I definitely see some cases where users would need to switch from one custom network to the other (developers during testing, miners etc.). I wouldn't say that it's an extreme case - even for everyday users - especially when you're trying to connect to other nodes if the one you usually use is down etc. I had to do exactly that one week ago with another wallet.

As for the drawback you described - it's good enough for now.
Let's keep it like that and see if some people are complaining down the line.

Thanks for taking the time to investigate :)

src/pages/Wallet/SendModalCheckTransaction.tsx
mvaivre: So, i would say:

- A component is an functional item that can be reused in multiple places.
- Pages and modals are layouts made of multiple components.

I'm in favor of either the first solution, or:

```
- components
- layout
  - pages
  - modals 
```

src/contexts/addresses.tsx
mvaivre: Yes! It makes way more sense to me. 💯

src/pages/Wallet/SendModalCheckTransaction.tsx
nop33: Ah, ok, so in your mental model, "components" are not "React components" but rather atomic, functional elements that are purpose is to be reused. I proposed option 3 with the "React components" meaning of "components" in mind. But since we have other categorization of React components (ex: `routes`, `contexts`), then I also agree that I am not in favor of option 2 and 3.

I would like to keep things simple and flat and avoid overengineering for the moment. Our project is still relatively small, our team is also small, and our goal is to save time during development by making things intuitive. I would like to go with option 1 for now and reconsider the categorization of our files if the need comes (like it did now 😄).

```
- components
- modals
- pages
```

In this case, and to answer your initial question, I'll move the send modal steps in: `/modals/SendModal/...`.

mvaivre: Sounds good 👍

nop33: Would you mind if I refactor this (as well as https://github.com/alephium/alephium-wallet/pull/128#discussion_r811194189)  in the next PR? It will cause quite some conflicts that will be annoying to solve.

mvaivre (approved):

PR #131, opened 2022-02-23
Derive new addresses (probably last major PR)

nop33:
## New features
- Created transaction details side modal
- Implemented overview page with account & addresses summary cards and global transaction list
- Added account switch in sidebar

## Small things
- Added paper airplane SVG in send modal
- Improved state management
- Several small UI fixes

## Cleanup
- Moved all routes to their correct respective places
- Introduced `modals` component directory and moved all modals there
- Fixed bug with storing network settings

## 🗣️ 🔉 sound on!
https://user-images.githubusercontent.com/1579899/155396643-36632282-93ae-4873-aa44-802dd24e583a.mp4


nop33:
Sorry for this enormous PR @mvaivre, most of it is due to the relocation of the modal components. It should be the last big PR regarding the address derivation feature that implements your beautiful mockups. I'll work tomorrow on reviewing your explorer PR while waiting for feedback on this one.

mvaivre:
That's great, thank you! 
I'll focus again on reviewing the code, then after this one I'll open an issue to centralize the little UI issues that I may find here and there.

src/components/ModalCentered.tsx
```diff
@@ -0,0 +1,176 @@
...
+const ModalLoadingSpinner = styled.div`
+  position: absolute;
+  display: flex;
+  width: 100%;
+  height: 100%;
+  justify-content: center;
+  align-items: center;
+`
+
+export default ModalCentered
```
mvaivre, line 176: Do we put the default export at the top of the styled component for consistency sake (as discussed above), or do we make a diff wether the container is styled or not?

src/components/ModalCentered.tsx
```diff
@@ -0,0 +1,176 @@
...
+  title,
+  subtitle,
+  onClose,
+  focusMode,
+  isLoading,
+  header,
+  children
+}) => (
+  <Modal onClose={onClose} focusMode={focusMode}>
+    <StyledModal
```
mvaivre, line 47: Are we wrapping successively 2 animated components? (Modal, and StyledModal?).
We could maybe simply allow overriding the `motion` related props in the `Modal` component? (initial, animate, exit and transition).

Something like:

```ts
export interface ModalProps extends ComponentPropsWithoutRef<typeof motion.div> {
  onClose: () => void
  focusMode?: boolean
  hasPadding?: boolean
}

const Modal: FC<ModalProps> = ({
  children,
  onClose,
  focusMode,
  initial = { opacity: 0 },
  animate = { opacity: 1 },
  exit = { opacity: 0 },
  transition = { duration: 0.2, ease: 'easeOut' },
  className,
  hasPadding = true
}) => {

  [...]

  return (
    <motion.div initial={initial} animate={animate} exit={exit} transition={transition} className={className}>
      {children}
      <ModalBackdrop onClick={onClose} focusMode={focusMode} />
    </motion.div>
  )
}

export default styled(Modal)`
  position: fixed;
  top: 0;
  bottom: 0;
  right: 0;
  left: 0;
  display: flex;
  padding: ${({ hasPadding }) => hasPadding && css`var(--spacing-4)`};
  z-index: 1000;
`
```

src/components/TransactionBadge.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import styled from 'styled-components'
+
+import Badge from './Badge'
+
+const TransactionBadge = styled(Badge)`
```
mvaivre, line 23: Is this standalone component really necessary, or should we find a way to reduce differences in Badges design?
This is my bad: I have a hard time deciding between using badges with straight edges (problem: this may be confused with buttons), or 100% round corners...

The former is easier on the eye, as it aligns better in tables etc.

Where I was currently going (tried in the explorer) : rise buttons a bit with a `box-shadow` to imply the possibility to interact; keep badges flat with straight edges, no shadows.

Maybe you have an opinion about this?

I'd propose to take the easy road for now:
- Keep one type of badge only (remove `TransactionBadge`)
- Use straight corners

I'll take care of cleaning this up one a coherent design language emerges.

mvaivre (commented):
First part 🥇

mvaivre:
About the "hero background image": I thought that we could select a few, which would be randomly selected when the user opens an account, adding a bit of variety. 
Should we [fetch them](https://unsplash.com/developers) from Unsplash directly (we define a list of URLs) in order to keep the app's weight low? We could keep only a couple of images locally for offline use.

src/components/AccountSummaryCard.tsx
```diff
@@ -0,0 +1,111 @@
...
+            Locked: <Amount value={totalLockedBalance} />
+          </span>
+        </BalanceLocked>
+      </Balances>
+      <AccountName>{currentUsername}</AccountName>
+    </div>
+  )
+}
+
+export default styled(AccountSummaryCard)`
```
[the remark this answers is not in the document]
nop33: Yes, I find it useful to have the styles of the main component as close to the main component as possible.

src/components/AddressSummaryCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <div className={className}>
+      <ClickableArea
+        onClick={() => clickable && history.push(`/wallet/addresses/${address.hash}`)}
+        clickable={clickable}
+      >
+        <AddressNameSection>
+          <Label color={address.settings.color} truncate>
+            {address.labelDisplay()}
```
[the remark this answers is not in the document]
nop33: I agree that the method I've been using so far is not the best. And the proposal to have `get...` functions makes total sense. However, I'd like to propose using computer properties instead of functions, like I do for `availableBalance` for example (instead of creating a function called `getAvailableBalance`. I'll explain in a small commit.

src/components/ModalCentered.tsx
nop33: Let's keep them always at the top for consistency

src/components/ModalCentered.tsx
nop33: Hm, but do we rly want to override the animation? We want the overall backdrop to fade in (not move vertically) and the actual modal to fade in AND move vertically.

But I agree that maybe it's not the best to have 2 successive animated components. That's why I propose to move the first animation only on the backdrop, and leave the second animation where it belongs. See diff:

<details><summary>Click to see diff</summary>

```diff
diff --git a/src/components/Modal.tsx b/src/components/Modal.tsx
index 9660639..55c37aa 100644
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -18,15 +18,16 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { motion } from 'framer-motion'
 import { FC, useCallback, useEffect } from 'react'
-import styled, { css } from 'styled-components'
+import styled from 'styled-components'
 
 export interface ModalProps {
   onClose: () => void
   focusMode?: boolean
   hasPadding?: boolean
+  className?: string
 }
 
-const Modal: FC<ModalProps> = ({ children, onClose, focusMode, hasPadding = true }) => {
+const Modal: FC<ModalProps> = ({ children, onClose, focusMode, hasPadding = true, className }) => {
   // Prevent body scroll on mount
   useEffect(() => {
     document.body.style.overflow = 'hidden'
@@ -54,31 +55,32 @@ const Modal: FC<ModalProps> = ({ children, onClose, focusMode, hasPadding = true
   }, [handleEscapeKeyPress, onClose])
 
   return (
-    <ModalContainer
-      initial={{ opacity: 0 }}
-      animate={{ opacity: 1 }}
-      exit={{ opacity: 0 }}
-      transition={{ duration: 0.2, ease: 'easeOut' }}
-      hasPadding={hasPadding}
-    >
+    <div className={className}>
       {children}
-      <ModalBackdrop onClick={onClose} focusMode={focusMode} />
-    </ModalContainer>
+      <ModalBackdrop
+        initial={{ opacity: 0 }}
+        animate={{ opacity: 1 }}
+        exit={{ opacity: 0 }}
+        transition={{ duration: 0.2, ease: 'easeOut' }}
+        onClick={onClose}
+        focusMode={focusMode}
+      />
+    </div>
   )
 }
 
-const ModalContainer = styled(motion.div)<{ hasPadding?: boolean }>`
+export default styled(Modal)<{ hasPadding?: boolean }>`
   position: fixed;
   top: 0;
   bottom: 0;
   right: 0;
   left: 0;
   display: flex;
-  padding: ${({ hasPadding }) => hasPadding && css`var(--spacing-4)`};
+  padding: ${({ hasPadding }) => hasPadding && 'var(--spacing-4)'};
   z-index: 1000;
 `
 
-export const ModalBackdrop = styled.div<{ focusMode?: boolean; light?: boolean }>`
+export const ModalBackdrop = styled(motion.div)<{ focusMode?: boolean; light?: boolean }>`
   position: absolute;
   top: 0;
   right: 0;
@@ -95,5 +97,3 @@ export const ModalBackdrop = styled.div<{ focusMode?: boolean; light?: boolean }
       ? 'rgba(0, 0, 0, 0.15)'
       : 'rgba(0, 0, 0, 0.6)'};
 `
-
-export default Modal
diff --git a/src/components/ModalCentered.tsx b/src/components/ModalCentered.tsx
index d40a0ed..04bfa71 100644
--- a/src/components/ModalCentered.tsx
+++ b/src/components/ModalCentered.tsx
@@ -44,7 +44,7 @@ const ModalCentered: FC<ModalCenteredProps> = ({
   children
 }) => (
   <Modal onClose={onClose} focusMode={focusMode}>
-    <StyledModal
+    <CenteredBox
       initial={{ opacity: 0, y: 20 }}
       animate={{ opacity: 1, y: 0 }}
       exit={{ opacity: 0, y: 20 }}
@@ -71,7 +71,7 @@ const ModalCentered: FC<ModalCenteredProps> = ({
           </ModalLoadingSpinner>
         </>
       )}
-    </StyledModal>
+    </CenteredBox>
   </Modal>
 )
 
@@ -88,7 +88,7 @@ export const HeaderLogo = styled.div`
   width: 100%;
 `
 
-const StyledModal = styled(motion.div)`
+const CenteredBox = styled(motion.div)`
   display: flex;
   flex-direction: column;
   margin: auto;
```

</details>

src/components/TransactionBadge.tsx
nop33: Yes, I like your approach:
- raised (box shadow) elements for buttons
- flat elements for labels/badges

At the moment we are a bit messy. We have 3 complete different components (no inheritance) doing similar jobs:

image

I think that a badge with no background is not a badge at all. That's why I'd propose that:
- Labels are Badges with rounded corners and special coloring. This includes the "NetworkBadge".
- "From" and "Genesis TX" are normal Badges
- "Received" and the amounts are neither Labels nor Badges. The "special" thing about them is that they receive a coloring based on the transaction type. I'd call these "TransactionalInfo".

nop33:
> About the "hero background image"

Although I love the idea of variety, I think the idea of fetching images from the Unsplash API does not scale well. They limit production apps to 5'000 requests per hour. Unless we implement some kind of caching, there'll be 1 request every time a wallet user goes to the overview page. 5'000/h sounds quite a relaxed limit however and we should be fine for now.

Also, how will we confidentially manage the API key? We cannot ship it together with the app. We'd need a proxy backend server, wouldn't we ([guidelines](https://help.unsplash.com/en/articles/2511245-unsplash-api-guidelines))?

What do you think of the idea of simply choosing 2 photos for now (one for the dark theme and one for the light theme), and change the photos every time we release a new version of the wallet? :)

src/components/ModalCentered.tsx
mvaivre: Yes, LGTM.

src/components/TransactionBadge.tsx
mvaivre: We're on the same page 👌

src/modals/NewAddressModal.tsx
```diff
@@ -20,18 +20,18 @@ import { AddressAndKeys, addressToGroup, deriveNewAddressData, TOTAL_NUMBER_OF_G
...
-import { Address, useAddressesContext } from '../../contexts/addresses'
-import { useGlobalContext } from '../../contexts/global'
-import { getRandomLabelColor } from '../../utils/colors'
+import ExpandableSection from '../components/ExpandableSection'
+import InfoBox from '../components/InfoBox'
+import ColoredLabelInput from '../components/Inputs/ColoredLabelInput'
+import KeyValueInput from '../components/Inputs/InlineLabelValueInput'
+import Select from '../components/Inputs/Select'
+import Toggle from '../components/Inputs/Toggle'
+import ModalCentered, { ModalFooterButton, ModalFooterButtons } from '../components/ModalCentered'
```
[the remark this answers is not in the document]
mvaivre: And maybe rename `Modal` into `ModalContainer` to be 100% explicit?

src/pages/Wallet/OverviewPage.tsx
```diff
@@ -0,0 +1,237 @@
...
+import ActionLink from '../../components/ActionLink'
+import AddressSummaryCard from '../../components/AddressSummaryCard'
+import Badge from '../../components/Badge'
+import Button from '../../components/Button'
+import Label from '../../components/Label'
+import { MainContent } from '../../components/PageComponents/PageContainers'
+import { PageH2 } from '../../components/PageComponents/PageHeadings'
+import Table, { TableCell, TableCellPlaceholder, TableProps, TableRow } from '../../components/Table'
+import { Address, useAddressesContext } from '../../contexts/addresses'
+import DayskyImage from '../../images/daysky.jpeg'
```
mvaivre, line 37: `DayskyImageSrc` ? :D

src/style/globalStyles.ts
```diff
@@ -22,9 +22,9 @@ import { isElectron } from '../utils/misc'
 import resets from './resets'
 import tags from './tags'
 
-export const appHeaderHeight = '50px'
+export const appHeaderHeight = '60px'
```
mvaivre, line 25: We could name it `appHeaderHeightPx` and use a number.
Same with `walletSidebarWidthPx`

src/modals/NewAddressModal.tsx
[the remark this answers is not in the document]
nop33: We could, but since we don't rly have the need for another `SideModal`, I'd rather leave it as it is. I can do it if you insist.

src/pages/Wallet/OverviewPage.tsx
nop33: haha YES SIR

src/utils/settings.ts
```diff
@@ -98,21 +98,17 @@ export const loadStoredSettings = (): Settings => {
...
-          theme: deprecatedThemeSetting as ThemeType
-        }
-      }
-    : {}
+  const migratedSettings = {
+    network: storedSettings.network ?? (storedSettings as unknown as DeprecatedNetworkSettings),
+    general: deprecatedThemeSetting
+      ? { ...storedSettings.general, theme: deprecatedThemeSetting as ThemeType }
+      : storedSettings.general
+  }
```
mvaivre, line 106: I forgot about this piece of code. Your solution is way more elegant, but it reminds me that we may want a test for that...?

nop33: There was a [bug ](https://github.com/alephium/alephium-wallet/pull/131/commits/927eb2d6286efa41f24151844b9b343cb158ebbc)in this code, that's why I updated it. Yeah, we probably need to test it somehow.

src/modals/NewAddressModal.tsx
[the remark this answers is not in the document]
mvaivre: The `components` folder start to be quite large - I know - and I'd be in favour of starting to regrouping in theme, like I did with `inputs`.
But I made this remark because I was looking for the "side" modal variant (as I saw the "center" version) and couldn't find it... 
I can definitely see other cases where the side modal would be used to display complementary informations :)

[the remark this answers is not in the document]
nop33: ok! Convienced! And I'll move it into the `modals` folder, together with all its other modal friends 🧑‍🤝‍🧑

mvaivre:
> > About the "hero background image"
> 
> Although I love the idea of variety, I think the idea of fetching images from the Unsplash API does not scale well. They limit production apps to 5'000 requests per hour. Unless we implement some kind of caching, there'll be 1 request every time a wallet user goes to the overview page. 5'000/h sounds quite a relaxed limit however and we should be fine for now.
> 
> Also, how will we confidentially manage the API key? We cannot ship it together with the app. We'd need a proxy backend server, wouldn't we ([guidelines](https://help.unsplash.com/en/articles/2511245-unsplash-api-guidelines))?
> 
> What do you think of the idea of simply choosing 2 photos for now (one for the dark theme and one for the light theme), and change the photos every time we release a new version of the wallet? :)

I didn't look into the API limitation, but yeah, I thought about a simple caching system.
Let's keep it simple yeah! We'll see about changing it for every release, but that's a nice idea :)
I also thought that it would be fun to let users choose a unique image for each account, to help distinguish them.

src/style/globalStyles.ts
nop33: done

src/pages/Wallet/OverviewPage.tsx
nop33: done

src/modals/NewAddressModal.tsx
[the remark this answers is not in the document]
nop33: done

src/components/ModalCentered.tsx
nop33: done

src/components/AddressSummaryCard.tsx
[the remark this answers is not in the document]
nop33: done

src/components/AccountSummaryCard.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+import { Check, Lock } from 'lucide-react'
+import styled from 'styled-components'
+
+import { useAddressesContext } from '../contexts/addresses'
+import { useGlobalContext } from '../contexts/global'
+import Amount from './Amount'
+
+const AccountSummaryCard = ({ className }: { className?: string }) => {
+  const { currentUsername } = useGlobalContext()
```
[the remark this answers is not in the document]
nop33: done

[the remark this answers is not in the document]
mvaivre: I didn't check the code yet, but did you make sure to migrate the associated key in the local storage as well?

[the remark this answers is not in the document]
nop33: I don't think there is a migration required. We save and load the encrypted wallet using the name of the account as key:
```js
Storage.save(accountName, walletEncrypted)
Storage.load(accountName)
```

The string `"username"` was not stored anywhere in local storage.

nop33:
> There was a [bug ](https://github.com/alephium/alephium-wallet/pull/131/commits/927eb2d6286efa41f24151844b9b343cb158ebbc)in this code, that's why I updated it. Yeah, we probably need to test it somehow.

@mvaivre tests added

src/tests/settings.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  localStorage.setItem('theme', 'pink')
+  expect(localStorage.getItem('theme')).toEqual('pink')
+  expect(loadStoredSettings().general.theme).toEqual('pink')
+  expect(localStorage.getItem('theme')).toBeNull()
+
+  localStorage.setItem('settings', JSON.stringify(mockSettings))
+  expect(loadStoredSettings().general.walletLockTimeInMinutes).toEqual(999)
+  expect(loadStoredSettings().network.nodeHost).toEqual('https://node')
+  expect(loadStoredSettings().network.explorerApiHost).toEqual('https://explorer-backend')
+  expect(loadStoredSettings().network.explorerUrl).toEqual('https://explorer')
```
mvaivre, line 69: Is there an advantage not to replace those 4 lines above with: 
```ts
expect(loadStoredSettings()).toEqual(mockSettings)
```
?

nop33: no :)

nop33: fixed

src/tests/settings.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      })
+    ).toEqual('custom')
+})
+
+it('Should load settings from local storage', () => {
+  expect(loadStoredSettings()).toEqual(defaultSettings)
+
+  localStorage.setItem('theme', 'pink')
+  expect(localStorage.getItem('theme')).toEqual('pink')
+  expect(loadStoredSettings().general.theme).toEqual('pink')
```
mvaivre, line 62: Looking at this, I'm wondering if we should maybe split the `loadStoredSettings` into two explicit functions:

- `migrateDeprecatedSettings` => which reuse most of the code found in our current `loadStoredSettings`
- `loadStoredSettings` => a rename of `returnSettingsObject`
- `loadAndMigrateStoredSettings` => `loadStoredSettings()` + `migrateDeprecatedSettings()`

The tests would be clearer like that, as we could test the migration and the loading process individually. WDYT?

src/tests/settings.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  localStorage.setItem('settings', JSON.stringify(mockSettings))
+  expect(getStoredSettings('general')).toEqual(mockSettings.general)
+})
+
+it('Should update stored settings', () => {
+  const newNetworkSettings = {
+    nodeHost: 'https://node1',
+    explorerApiHost: 'https://explorer-backend1',
+    explorerUrl: 'https://explorer1'
+  }
```
mvaivre, line 84: Should we declare all static objects at the top of the file (where you put the mockSettings)?

nop33: Hm, I think I'd go case by case. This particular object is not being reused like the `mockSettings` one. So, I'd personally keep it here.

src/tests/settings.test.ts
nop33: At the moment we call the `loadStoredSettings` in 4 distinct places in our app:
- `src/components/Address.tsx` (this should be replaced by getting the settings from the global context
- `src/contexts/global.tsx`, once to set the `settings` property of the `initialGlobalContext` object and once to set the default value of the state.
- `src/utils/api-clients.ts` (IMO, the `settings` argument of the `createClient` function should not be optional. That way we don't need to call the `loadStoredSettings` there and we only call it in the global context file.

My opinion is that we should have only 2 functions:
1. `migrateDeprecatedSettings`, which migrates any deprecated settings found in local storage and returns the final settings object. This function could be used only in one place: to set the `settings` property of the `initialGlobalContext` object. This way, it would be called exactly 1 time when the app runs.
2. `loadStoredSettings`, which just loads the stored settings (same what `returnSettingsObject` currently does). This function would be used only 1 one place: to set the default value of the state.

Moreover, I would delete `getStoredSettings` since it's not used anywhere.

mvaivre: I full agree with the general idea of simplifying settings loading and avoid unnecessary calls. I'm not so happy with [how](https://github.com/alephium/alephium-wallet/commit/41a131322db5e1a49e07464df4959e4ba28b1a5c#r67647699) the migration method is called though. I'd make it more explicit, and migrate the settings before passing them in the context.

A proposition: we could have a `migrationRequired` flag in the setting object, set by a "test method" called asap on app launch. I would make sure to call this method as early as possible, either in the `contexts/global.tsx`file or even in the good old `index.tsx` , so that we are 100% sure that any following access to the settings is safe (ie. settings are migrated, and in a correct state).
Once the migration is done, the `migrationRequired` is set to `false`, and no more calls are done, until the "migration test method" is changed by a subsequent app update.

WDYT?

src/tests/settings.test.ts
mvaivre: As you wish!

src/tests/settings.test.ts
nop33: yeah, I like this approach, I'll work on it tomorrow! Thanks for taking the time!

nop33: @mvaivre I have just pushed a commit that takes your feedback into consideration for a better solution.

https://github.com/alephium/alephium-wallet/pull/131/commits/8d75b2140ec5bc656b371b85f9f04524f0ca0b4c

src/style/themes.ts
```diff
@@ -49,7 +49,7 @@ export const lightTheme: DefaultTheme = {
     valid: '#4ebf08',
     highlightGradient: 'linear-gradient(45deg, rgba(18,0,218,1) 0%, rgba(255,93,81,1) 100%)'
   },
-  badge: {
+  txInfo: {
```
mvaivre, line 52: This is too specific for a theme IMO. I would keep "badge".

I think we could follow [this](https://github.com/alephium/explorer/pull/55/commits/a4257809037832a555cfc8d6bca4892877017d97) approach, as we discussed earlier: "badges" are flat and can have a border. On the explorer, so far I'm using badges for certain amounts I want to highlight + tx direction (from / to). The same could be done here.

mvaivre: (As you can see in my commit, I'm no longer using specific colors for badges, so this "txInfo" part of the theme could even be removed entirely... for now :) In the future, if we want to have more contrasted options, we will make sure to use secondary colors defined in the "global" property - like alert, warning, valid etc.)

nop33: I'm a little confused 😅  

So, [currently in the wallet](https://github.com/alephium/alephium-wallet/blob/derive-new-addresses/src/components/Badge.tsx#L37-L47), the `Badge`'s text color is set from the props of the component, its background color is a more transparent version of the text color, and the border (if set) is the same as the text color. That was done because [we agreed](https://github.com/alephium/alephium-wallet/pull/131#discussion_r813888470) that "labels" are essentially "badges".

If I remove the `txInfo` from the theme, then we need a way to define the color palette for the `TransactionalInfo` component (used to display the words "↓ Received", "↑ Sent", "Pending", as well as the amounts in the transaction list tables). Those elements are not badges, since they have no background at all. Right?

Do you suggest that I move those color definitions from the theme into the `TransactionalInfo` component?

Thanks for clarifying and helping me understand!

nop33: I need the `txInfo` colors in 2 different places, so moving them in the `TransactionalInfo` component doesn't make sense.

| Transaction lists | Transaction details |
| ---------------- | ------------------- |
| image | image |

mvaivre: Ah, I'm sorry, didn't realise that this was specifically used for the `TransactionalInfo` component now. 
My comment still stands though, even more in that case - we don't want one entry in the theme that is here only to be used by one single component. 

The `TransactionalInfo` component should use colors options we already have.
In that specific case, `received` could use the `global.valid` color, `send` the `font.primary`, `pending` would be `font.secondary` etc.

Does that make sense? Sorry for the confusion.

Let's do that: I'll sync the palette with the explorer's one when doing the overall testing and cleaning, so that you don't have to spend time on this. Would that be ok for you? 

Thanks!

nop33: Yes, that makes sense! I'll reuse existing colors, delete the `txInfo` ones, and you can update the colors when you want to!

mvaivre: That's it 💯 . Thanks!

src/components/AccountSummaryCard.tsx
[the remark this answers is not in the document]
mvaivre: that's perfect.

mvaivre (approved):

Issue #115, opened 2022-01-17
ARM64 support on Linux


blackmennewstyle:
Hello talented devs :)

Can we hope for some arm64 support in the future? I'm planning to install the wallet on a Raspberry Pi 4 8GB running GNU/Linux Debian 11 - kernel 5.10.0-9-arm64, is it possible at the current stage of the project? If not, will it be possible in the future? :$

mvaivre:
Hey there, thanks for your message!
The ARM64 architecture is not yet supported on Linux, mostly due to the lack of demand so far. But here you are... 
So yeah, we can definitely hope for some arm64 in the future, probably in the next release! ;)

blackmennewstyle:
> Hey there, thanks for your message! The ARM64 architecture is not yet supported on Linux, mostly due to the lack of demand so far. But here you are... So yeah, we can definitely hope for some arm64 in the future, probably in the next release! ;)

Nice :)
Merci beaucoup Mikaël !
Enjoy your week :)

mvaivre:
@nop33 Hey Ilias, could you look into this while I'm testing the address derivation & UX?
It should basically be [1 flag to add](https://www.electron.build/cli.html) in the package.json script + adapt the corresponding github build action.
Do you still have VMWare running on your M1 mac? If so, should be quite easy to spin up an arm64 build of debian / ubuntu. If not, don't spend to much effort on this RN. KMP, thanks!

nop33:
Hey @mvaivre, I am working on implementing tests for the settings utility functions of the wallet project, as we discussed in the PR. I don't have a VMWare installed, but I can install one. I can do that after I implement the tests.

nop33:
Hello @blackmennewstyle! I've added a new npm script that you can use to build the app for ARM64 from source (I updated the README with instructions). Unfortunately, I couldn't get GitHub's CI to work with ARM64 just yet. The process to build for ARM64 from source is very simple:
```
npm install
npm run electron-pack-linux:arm64
```

Keep in mind that the `master` branch is cutting-edge code and should not be used in production. Use code from the releases instead.

blackmennewstyle:
> Hello @blackmennewstyle! I've added a new npm script that you can use to build the app for ARM64 from source (I updated the README with instructions). Unfortunately, I couldn't get GitHub's CI to work with ARM64 just yet. The process to build for ARM64 from source is very simple:
> 
> ```
> npm install
> npm run electron-pack-linux:arm64
> ```
> 
[...]

PR #134, opened 2022-02-28
Build linux arm64 package

nop33:
Related to #115

I managed to build a linux arm64 package with:
```
npm run build && npx electron-builder -l --arm64
```

Then, I installed [Debian 10.4 (LDXE)](https://mac.getutm.app/gallery/debian-10-4-ldxe) using the [UTM](https://mac.getutm.app/) software on my M1 mac and installed the `alephium-wallet_1.1.0_arm64.deb` package:

Screenshot 2022-02-28 at 16 57 05


nop33:
I don't understand why the CI is not running the last 2 tests... After waiting for 15h [I canceled them](https://github.com/alephium/alephium-wallet/runs/5362745004?check_suite_focus=true) and [re-triggering them](https://github.com/alephium/alephium-wallet/runs/5371704049?check_suite_focus=true), but they keep being stuck at "Queued"...

nop33:
> I don't understand why the CI is not running the last 2 tests... After waiting for 15h [I canceled them](https://github.com/alephium/alephium-wallet/runs/5362745004?check_suite_focus=true) and [re-triggering them](https://github.com/alephium/alephium-wallet/runs/5371704049?check_suite_focus=true)[](https://github.com/nop33), but they keep being stuck at "Queued"...

The reason was that the OS list was also being used to define the GitHub's runner OS:
```
runs-on: ${{ matrix.os }}
```

My experiment above worked because I built the Debian ARM64 package using my M1 MacBook which also is ARM64. But since [GitHub Runners do not support](https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job#choosing-github-hosted-runners) at the moment an ARM64 Linux OS, more research needs to be done to manage to build an ARM64 artifact on an ARM64 Runner.

PR #140, opened 2022-03-03
Allow address derivation from welcome screen + small fixes

nop33:
## 🗣️ 🔈 sound on!
https://user-images.githubusercontent.com/1579899/156526643-11400e71-5d7d-443b-a58d-adf67f491f04.mp4


src/pages/WalletManagement/WalletWelcomePage.tsx
```diff
@@ -16,27 +16,71 @@ You should have received a copy of the GNU Lesser General Public License
...
   useTimeout(() => {
     setConfettiRunning(false)
   }, 3000)
 
+  const onButtonClick = () => {
+    if (shouldGenerateOneAddressPerGroup && wallet?.seed && mainAddress) {
+      const newAddresses = []
+      const color = getRandomLabelColor()
+
+      for (let group = 0; group < TOTAL_NUMBER_OF_GROUPS; group++) {
```
mvaivre, line 58: We already have a `generateOneAddressPerGroup()` method in `NewAddressModal`. We could probably make it an export in `AddressesContextProvider` to go a bit DRYier?

src/pages/WalletManagement/WalletWelcomePage.tsx
```diff
@@ -47,14 +91,43 @@ const WalletWelcomePage = () => {
         <SubParagraph>Welcome to Alephium.</SubParagraph>
       </Section>
       <FooterActionsContainer apparitionDelay={0.3}>
-        <Button onClick={() => history.push('/wallet')} submit>
+        <Button onClick={onButtonClick} submit>
           {"Let's go!"}
         </Button>
+        <div>
+          <AdvancedUserMessage>
+            Advanced user: want to start with{' '}
```
mvaivre, line 99: Tiny remark: if you're using the `<strong>`tag to go bold, I'd rather wrap the "normal text" in a `<span>` instead of using `{' '}` which feels a bit dirty IMO.

src/pages/WalletManagement/WalletWelcomePage.tsx
nop33: Good point, fixed now!

src/pages/WalletManagement/WalletWelcomePage.tsx
nop33: hm, but the point of `{' '}` is to ensure that the space between the word "with" and the word "one" is the same as all other spaces between words in a normal text. If I wrap the first part of the text with `<span>` then I need to hardcode a margin. I think that's dirtier. Besides, the [`<strong>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong) is meant to be used to wrap a piece of text to increase its importance. It's not meant to be a sibling of other tags. WDYT?

image

FYI, this is the compiled code:

```
<div class="sc-jnrVZQ kKfewa">
  Advanced user: want to start with <strong>one address per group for mining or DeFi? <svg ...></svg></strong>
</div>
```

mvaivre: I'm not sure I understand the issue - I don't see why we would need margins. Can't we [do that](https://codesandbox.io/s/react-playground-forked-zuy7pk?file=/index.js)? Spaces in `span` or `strong` are taken into account properly.
Also, why do you say that `strong` isn't meant to be used as a sibling of other tags? Isn't it its [typical use](https://html.com/tags/strong/), to highlight parts of a text, inline with some `b`, `span` etc.?

nop33: Sorry, you're right, I could have put a space within the span, I was putting it outside of it. Brainfart.

> Also, why do you say that strong isn't meant to be used as a sibling of other tags? Isn't it its [typical use](https://html.com/tags/strong/), to highlight parts of a text, inline with some b, span etc.?

What I meant was that its siblings don't need to be tags, it can be plain text as well.

I personally don't like the solution of adding unnecessary HTML elements to fix a coding style issue (making the HTML heavier), but I also think I used too much of our time on this so I just committed a fix as you preferred :)

mvaivre: Haha, well, ok...! Just FYI though, if you don't like the extra html tags, you could remove one unnecessary span easily, leading to an arguably more logical semantic (especially since you don't have to make `b` a flex container) 🙃 

<details>

```diff
diff --git a/src/pages/WalletManagement/WalletWelcomePage.tsx b/src/pages/WalletManagement/WalletWelcomePage.tsx
index 2f6ba1d..588882d 100644
--- a/src/pages/WalletManagement/WalletWelcomePage.tsx
+++ b/src/pages/WalletManagement/WalletWelcomePage.tsx
@@ -80,14 +80,14 @@ const WalletWelcomePage = () => {
         </Button>
         <div>
           <AdvancedUserMessage>
-            <span>Advanced user: want to start with </span>
-            <b>
-              <span>one address per group for mining or DeFi? </span>
-              <InfoIcon
-                size="12px"
-                onClick={() => openInWebBrowser('https://wiki.alephium.org/Solo-Mining-Guide.html')}
-              />
-            </b>
+            <span>
+              Advanced user: want to start with
+              <b> one address per group for mining or DeFi? </b>
+            </span>
+            <InfoIcon
+              size="12px"
+              onClick={() => openInWebBrowser('https://wiki.alephium.org/Solo-Mining-Guide.html')}
+            />
           </AdvancedUserMessage>
           <ExpandableSectionStyled
             sectionTitleClosed="Show advanced options"
@@ -151,13 +151,13 @@ const AdvancedUserMessage = styled.div`
   color: ${({ theme }) => theme.font.secondary};
   text-align: center;
   flex: 1;
+  display: inline-flex;
+  align-items: center;
+  gap: var(--spacing-1);
 
   b {
     font-weight: var(--fontWeight-medium);
     color: ${({ theme }) => theme.font.primary};
-    display: inline-flex;
-    align-items: center;
-    gap: var(--spacing-1);
   }
 `
```
</details>

nop33: Today is not my HTML day. Fixed. Also, nice job with the diff 👏

mvaivre: Haha, thanks for the diff thing yes! Happy to finally use it 🤝

src/pages/WalletManagement/WalletWelcomePage.tsx
```diff
@@ -47,14 +75,45 @@ const WalletWelcomePage = () => {
...
+        <Button onClick={onButtonClick} submit>
           {"Let's go!"}
         </Button>
+        <div>
+          <AdvancedUserMessage>
+            <span>
+              Advanced user: want to start with <b>one address per group for mining or DeFi?</b>
+            </span>
+            <InfoIcon
+              size="16px"
```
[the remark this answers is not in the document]
mvaivre: Oh, so nice!

mvaivre (approved):

PR #142, opened 2022-03-03
Add context menu

### LINKED ISSUE #119: Create a right click menu for copy/paste operations
Let's keep it simple for now, with simple copy / cut / paste choices.

nop33:
Closes #119

It is installed as a dependency and not a dev dependency. That's the only way I could make it work. The size of the app increased from `375.8 MB` to `376.1 MB`, but hey, I (almost) didn't have to write a single line of code for the context menu :D

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
       "name": "alephium-wallet",
       "version": "1.1.0",
       "dependencies": {
+        "electron-context-menu": "^3.1.2",
         "electron-is-dev": "^2.0.0"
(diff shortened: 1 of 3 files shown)


mvaivre:
image

The "inspect element" option will non be visible on non-dev version, right?

nop33:
> The "inspect element" option will non be visible on non-dev version, right?

right ;)

nop33:
image

mvaivre (approved):

PR #144, opened 2022-03-07
Clear password and mnemonic after account creation

lf94:
After account creation, the password given and the mnemonic created were being
left in memory even after the wallet is locked. This commit moves the password
and mnemonic management completely to the WalletContext and clears them at the
appropriate times.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@types/jest": "^27.0.2",
         "@types/lodash": "^4.14.173",
         "@types/node": "^16.9.6",
-        "@types/react": "^17.0.24",
+        "@types/react": "^17.0.39",
         "@types/react-color": "^3.0.6",
         "@types/react-dom": "^17.0.9",
         "@types/react-router-dom": "^5.3.0",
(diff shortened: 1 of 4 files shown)


lf94:
The past hour I've been in the process of trying to use Ghidra to verify the values are actually out of memory at the correct moments. Still trying.

nop33 (approved):
Nice job 👏

mvaivre (approved):
Thanks! 🥇

polarker:
> The past hour I've been in the process of trying to use Ghidra to verify the values are actually out of memory at the correct moments. Still trying.

Nice job! Would the virtual machine keep the password in memory for a short period of time before garbage collect it?

lf94:
> Nice job! Would the virtual machine keep the password in memory for a short period of time before garbage collect it?

That's why clearing a value is much more reliable than hoping for GC to be invoked at certain times. Since there is more than one popular JS VM, it's not safe enough to assume GC behavior.

I was not able to see values in Ghidra last night because it had a hard time searching all the memory regions. I believe I need to establish a process of checking for secrets like this so in the future we can state to users with a guarantee that secrets are not present during certain periods. It can be documented so people can even reverify themselves.

polarker:
> > Nice job! Would the virtual machine keep the password in memory for a short period of time before garbage collect it?
> 
> That's why clearing a value is much more reliable than hoping for GC to be invoked at certain times. Since there is more than one popular JS VM, it's not safe enough to assume GC behavior.
> 
> I was not able to see values in Ghidra last night because it had a hard time searching all the memory regions. I believe I need to establish a process of checking for secrets like this so in the future we can state to users with a guarantee that secrets are not present during certain periods. It can be documented so people can even reverify themselves.

What I meant was: if you set the value to null, the old object is not immediately removed (GC delays) from memory

lf94:
Heh yeah, exactly. I was implicitly agreeing with you. It's not reliable to know when GC fires.

lf94:
https://github.com/alephium/alephium-wallet/pull/145 requires this change and also requires to change the change, so I'm closing this one.

PR #145, opened 2022-03-08
WIP: Require password after some minutes

lf94:
The wallet's private keys are currently exposed for the entire duration of the
  account being unlocked. While this is a nice user experience it is not security
  oriented. As an alternative, the wallet's private keys will be available for
  additional signing for a certain amount of adjustable time after a transaction.
  Each subsequent transaction will reset the timer. This enables users to continue
  to view their funds, while keeping their private keys safe.

This PR still has to be sufficiently tested. I'm having difficulty mixing my local testnet node and the official testnet explorer backend.

### DIFF
--- a/src/components/AccountSummaryCard.tsx
+++ b/src/components/AccountSummaryCard.tsx
@@ -29,7 +29,11 @@ const AccountSummaryCard = ({ className }: { className?: string }) => {
 
   const totalBalance = addresses.reduce((acc, address) => acc + BigInt(address.details.balance), BigInt(0))
   const totalAvailableBalance = addresses.reduce((acc, address) => acc + address.availableBalance, BigInt(0))
-  const totalLockedBalance = addresses.reduce((acc, address) => acc + BigInt(address.details.lockedBalance), BigInt(0))
+  const totalLockedBalance = addresses.reduce((acc: bigint | undefined, { details: { lockedBalance } }) => {
+    if (acc === undefined) return acc
+    if (lockedBalance === undefined) return undefined
+    return acc + BigInt(lockedBalance)
+  }, BigInt(0))
 
   return (
     <div className={className}>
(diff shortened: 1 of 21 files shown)


lf94:
This PR still needs to:

* Adjust the original account lock time higher (3 minutes is now pretty small)
* Adjust the default password lock time
* Determine what information is needed from `plainWallet` from other components in order to secure it while not breaking anything else

nop33:
Is this still WIP? Or can we review? We usually assign for review once the PR stops being WIP.

lf94:
Yeah it's still WIP, my bad. I've unassigned and will re-assign when ready :slightly_smiling_face:

lf94:
It's been decided that leaving secrets in memory is ok, as if an attacker gets memory access, it's most likely unknown by the victim, and they will not know until it's too late. Hardware wallets will be necessary for much better secret isolation.

nop33:
But, I mean, since the work was done and this improves security, why drop it? Even if the threat model is different, this PR does something good. What was there left to be done?

src/components/AccountSummaryCard.tsx
```diff
@@ -29,7 +29,11 @@ const AccountSummaryCard = ({ className }: { className?: string }) => {
 
   const totalBalance = addresses.reduce((acc, address) => acc + BigInt(address.details.balance), BigInt(0))
   const totalAvailableBalance = addresses.reduce((acc, address) => acc + address.availableBalance, BigInt(0))
-  const totalLockedBalance = addresses.reduce((acc, address) => acc + BigInt(address.details.lockedBalance), BigInt(0))
+  const totalLockedBalance = addresses.reduce((acc: bigint | undefined, { details: { lockedBalance } }) => {
+    if (acc === undefined) return acc
+    if (lockedBalance === undefined) return undefined
+    return acc + BigInt(lockedBalance)
+  }, BigInt(0))
```
nop33, line 36: I disagree with the "regardless this issue should still be fixed in the wallet UI". We don't have to make the wallet backwards compatible every time the API changes.

PR #156, opened 2022-03-15
UI fixes

nop33:


lf94:
Code looks good to me :+1: The only thing I'd maybe change is turn the `'dark'` string into an enum to avoid mistyping and for code editors to auto-complete all Theme types. It's a separate issue though and can be handled in another PR.

lf94 (approved):

nop33:
> The only thing I'd maybe change is turn the 'dark' string into an enum to avoid mistyping and for code editors to auto-complete all Theme types.

I personally think it is not necessary to create enums. I think this is a tooling problem, not a coding problem. TypeScript and linting (and VSCode) are enough to detect wrong values and won't let me chose anything but the values defined in the [theme type](https://github.com/alephium/alephium-wallet/blob/master/src/style/themes.ts#L22). If I add a wrong value, linting fails.

image

image

image

lf94:
Oh it is typed then. Probably something like `type ThemeType = 'dark' | 'light'` - perfect then!

nop33:
> Probably something like type ThemeType = 'dark' | 'light' - perfect!

Yes! I linked to this LOC in my prev comment

PR #162, opened 2022-03-16
Fix test bugs

nop33:
- Keeping code DRYer in the forms to create/edit address metadata and adding a max length of 50 characters for address label
- Fixed offline issues and make sure to re-initialize the state when network settings change (even if still offline after changing them)
- Tide up snackbar so that it doesn't expand to the whole width of the app, but only takes the space that it needs (important for small screens)


lf94:
Looks good, I particularly like the refactor for `AddressMetadataForm` :+1:

lf94 (approved):

PR #163, opened 2022-03-16
Fix up tooltips

lf94:
This is a fix up commit for the 1.2.0 release. It fixed the tooltips on the
AppHeader's main address, network; and the clipboard button.


src/components/Badge.tsx
```diff
@@ -30,11 +31,15 @@ interface BadgeProps {
   rounded?: boolean
 }
 
-const Badge: FC<BadgeProps> = ({ className, children, truncate }) => {
+const Badge: FC<HasTooltip<BadgeProps>> = ({ className, children, truncate, ...props }) => {
```
nop33, line 34: hm, why `...props` here and not directly `dataTip`?

nop33: besides that, LGTM! 👏

lf94: Because the property name is "data-tip" unfortunately... A choice made by ReactTooltip.

lf94: I could do `{ <other properties here>, 'data-tip': dataTip }` I think, if you'd like?

lf94: I've gone ahead and changed it, I agree the explicitness is much better :slightly_smiling_face:

nop33 (approved):

Issue #172, opened 2022-03-21
Allow importing (NOT CREATING) wallets with less than 24 words seed in the desktop wallet. 


MaudSimon:
**Is your feature request related to a problem? Please describe.**
It is possible to create a wallet with less than 24 words seed via the node. However, the desktop wallet only allows to import wallet with 24 words seed. Which means someone that first created their wallet via the node with less than 24 words seed cannot import it to the desktop wallet. 
This problem was encountered by at least one user: https://www.reddit.com/r/Alephium/comments/tj7gzi/cannot_import_wallet_15_vs_24_words/

**Describe the solution you'd like**
Ideally, the desktop wallet would allow importing wallet with less than 24 words seed. In addition to that, we may want to consider restricting the node wallet and only allowing 24 words seed there too (feature on the desktop wallet would probably still be needed for the existing less than 24 words seed wallets, even though people could just migrate their funds).

nop33:
It should be easy to allow import of wallet with less than 24 words. @polarker @mvaivre, is this request verified and should be implemented? Or is it still at discussion phase?

mvaivre:
I've quickly discussed this with Maud yesterday. I don't see any issue allowing importing any mnemonic length, as long as we keep enforcing 24 words on creation. I'd like @polarker to confirm that it's fine for him too :)

tdroxler:
it's acualy not "any" size but: [12, 15, 18, 21, 24](https://github.com/alephium/alephium/blob/5b5e8cf1eecc9d16a5feda5616e9095834c6b201/crypto/src/main/scala/org/alephium/crypto/wallet/Mnemonic.scala#L54), tho I can't remember why those numbers.

polarker:
The only strict rule is that the mnemonic length should be at least 12 words to make it secure enough. `12, 15, 18, 21, 24` is just a tradition from Bitcoin as far as I remember

polarker:
Passphrase sometimes is considered as the 13-th or 25-th words. So if you see length of 13 or 25, it might be due to passphrase.

nop33:
So, I interpret all the above comments as "yes, we want this feature, go ahead and implement it".

polarker:
This is definitely good to have as some people use the same mnemonic for different coins and different wallets

PR #179, opened 2022-03-23
Add ability to restore wallets which use less than 24 word mnemonics

lf94:
The desktop-wallet made an assumption that all wallets will use 24 word
mnemonics but the Alephium node allows between 12 and 24 word mnemonics, so
some miners or node users are unable to import their wallets into the desktop-
wallet software.

This commit also cleans up the language around "24 words" and "secret phrase",
making it more consistent.

Resolves https://github.com/alephium/desktop-wallet/issues/172

### DIFF
--- a/src/pages/WalletManagement/ImportWordsPage.tsx
+++ b/src/pages/WalletManagement/ImportWordsPage.tsx
@@ -45,7 +45,7 @@ const ImportWordsPage = () => {
 
   const [phrase, setPhrase] = useState<{ value: string }[]>([])
   const allowedWords = useRef(bip39Words.split(' '))
-  const defaultPlaceholder = 'Type your 24 words'
+  const defaultPlaceholder = 'Type your secret phrase'
   const [customPlaceholder, setCustomPlaceholder] = useState(defaultPlaceholder)
   const tagifyRef = useRef<Tagify<TagData> | undefined>()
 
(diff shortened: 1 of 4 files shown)


src/pages/WalletManagement/ImportWordsPage.tsx
```diff
@@ -45,17 +45,15 @@ const ImportWordsPage = () => {
 
   const [phrase, setPhrase] = useState<{ value: string }[]>([])
   const allowedWords = useRef(bip39Words.split(' '))
-  const defaultPlaceholder = 'Type your 24 words'
+  const defaultPlaceholder = 'Type your 24 (or less) secret words'
```
nop33, line 48: How about just `Type your secret words` or `Type your secret phrase`?

src/pages/WalletManagement/ImportWordsPage.tsx
```diff
@@ -84,7 +82,8 @@ const ImportWordsPage = () => {
     }
   }
 
-  const isNextButtonActive = phrase.length === 24
+  // Alephium's node code uses 12 as the minimal mnemomic length.
```
nop33, line 85: Shall we put a link to the scala file? Or do you think it's unnecessary?

mvaivre (commented):
LGTM. I'd only change the wording according to Ilias's comment :)

src/pages/WalletManagement/ImportWordsPage.tsx
lf94: Nah, while I doubt it, it's possible the file could be renamed, moved, or the content in the file no longer refers to what we mentioned.

src/pages/WalletManagement/ImportWordsPage.tsx
lf94: Done :+1:

lf94:
@mvaivre all ready to go

PR #184, opened 2022-03-28
Replace Alef character with SVG

### LINKED ISSUE #176: Replace the "ALEF" utf8 symbol with a SVG
For consistency sake across OSes. As of today, the symbol is rendered in very different ways.

Note: this component could later be shared in a "common" UI library.

nop33:
Closes #176

### DIFF
--- a/src/components/AlefSymbol.tsx
+++ b/src/components/AlefSymbol.tsx
@@ -1,0 +1,53 @@
+/*
+Copyright 2018 - 2022 The Alephium Authors
+This file is part of the alephium project.
+
+The library is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import styled from 'styled-components'
+
+import alefSymbol from '../images/alef.svg'
+
+interface AlefSymbolProps {
+  color?: string
+  className?: string
+}
+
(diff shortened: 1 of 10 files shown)


src/components/AlefSymbol.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import alefSymbol from '../images/alef.svg'
+
+interface AlefSymbolProps {
+  color?: string
+  className?: string
+}
+
+const AlefSymbol = ({ className, color }: AlefSymbolProps) => {
+  return (
```
lf94, line 29: Maybe to save you and future us some pain we can use `useTheme` here to get the color? Then the symbol can color itself.

lf94 (commented):
Other than my one comment your code is great :+1: 

I still need to do a physical/visual test.

src/components/AlefSymbol.tsx
nop33: I don't see how using this hook can help me customize the color of the symbol based on the color of the text where it is used at. I need to set it to red when used in error messages, gray when used in input field labels, and green when used in confirmed transaction details.

src/components/Inputs/index.tsx
```diff
@@ -20,11 +20,12 @@ import { motion, Variants } from 'framer-motion'
 import styled, { css } from 'styled-components'
 import tinycolor from 'tinycolor2'
 
-export interface InputProps extends React.InputHTMLAttributes<HTMLInputElement> {
-  error?: string
+export interface InputProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'placeholder'> {
```
lf94, line 23: Mmm, `placeholder` not already optional on `HTMLInputElement`? Interesting they did that.

nop33: I need to override it because it expects `string | undefined` and it conflicts with the `null` or `ReactNode`.

lf94:
@nop33 I'm quite busy actually and I trust what you've done, I'll approve and merge.

lf94 (approved):

mvaivre:
Mhh, I was actually reviewing it, and testing "visually".
When we're not in a hurry, let's wait for all reviews before merging. Or at least ask for confirmation before merging. 
Thanks!

PR #190, opened 2022-04-04
Last UI fixes

nop33:
image

image

### DIFF
--- a/src/components/Inputs/Select.tsx
+++ b/src/components/Inputs/Select.tsx
@@ -82,10 +82,6 @@ function Select<T>({
     }
   }, [options, setInputValue, value])
 
-  const onSelectContainerClick = () => {
-    setShowPopup(true)
-  }
-
   return (
     <>
       <SelectContainer
@@ -93,7 +89,7 @@ function Select<T>({
         animate={canBeAnimated ? (!disabled ? 'shown' : 'disabled') : false}
         onAnimationComplete={() => setCanBeAnimated(true)}
         custom={disabled}
-        onClick={onSelectContainerClick}
+        onClick={() => setShowPopup(true)}
       >
         <InputLabel variants={inputPlaceHolderVariants} animate={!value ? 'down' : 'up'} htmlFor={id}>
           {placeholder}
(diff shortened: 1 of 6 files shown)


src/components/Inputs/Select.tsx
```diff
@@ -82,14 +83,19 @@ function Select<T>({
...
+
   return (
     <>
       <SelectContainer
         variants={sectionChildrenVariants}
         animate={canBeAnimated ? (!disabled ? 'shown' : 'disabled') : false}
         onAnimationComplete={() => setCanBeAnimated(true)}
         custom={disabled}
-        onClick={() => setShowPopup(true)}
+        onClick={onSelectContainerClick}
```
mvaivre, line 98: Nitpicking (really): I like to prefix my "action" callbacks with "handle". `handleSelectContainerClick`.

src/components/Inputs/Select.tsx
```diff
@@ -98,12 +104,13 @@ function Select<T>({
           <MoreVertical />
         </MoreIcon>
         <ClickableInput
-          type="button"
+          type="text"
           className={className}
-          disabled={disabled}
+          disabled
```
mvaivre, line 109: Isn't this equivalent to `disabled={true}`?

nop33: it is, I want it to always be disabled.

src/components/Inputs/Select.tsx
nop33: Hm, and I like to prefix actions with `on` which is more of the HTML standard 😅 Since it's a nitpick, can we just leave it as it is? 😇

lf94: (FWIW I prefer `on` as well, it's sticking to event terminology.)

mvaivre: Alright! Just a couple of resources on the subject:

- https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364
- https://reactjs.org/docs/handling-events.html
- https://jaketrent.com/post/naming-event-handlers-react

nop33: event handler removed, cause it's not needed :P

mvaivre (approved):

PR #204, opened 2022-04-12
Initial support for WalletConnect

lf94:
You will have to use the walletconnect-monorepo walletconnect fork in our organization in order to test this (use npm link).


lf94:
(Yes there are currently conflicts; I'll resolve them after walletconnect-monorepo stuff is done)

src/contexts/walletconnect.tsx
```diff
@@ -0,0 +1,153 @@
...
+  const [dappTransactionData, setDappTransactionData] = useState<SendTransactionData | undefined>(undefined)
+  const [requestEvent, setRequestEvent] = useState<SessionTypes.RequestEvent | undefined>(undefined)
+
+  useEffect(() => {
+    if (walletConnect === undefined) {
+      WalletConnectClient.init({
+        controller: true,
+
+        // TODO: configurable during building?
+        projectId: '6e2562e43678dd68a9070a62b6d52207',
```
lf94, line 61: Need to make this configurable

src/modals/WalletConnectModal.tsx
```diff
@@ -0,0 +1,194 @@
...
+  }, [walletConnect, uri, onConnect])
+
+  const onApprove = useCallback(async () => {
+    if (proposal === undefined) {
+      setState(State.InitiateSession)
+      return
+    }
+
+    const accounts: string[] = addresses.map((a) => `alephium:${a.network}:${a.hash}`)
+    await walletConnect?.approve({ proposal, response: { state: { accounts } } })
```
lf94, line 72: I should return only an address marked as Main, or the first address if none are otherwise.

walletconnect-project-id.txt
```diff
@@ -0,0 +1 @@
+6e2562e43678dd68a9070a62b6d52207
```
lf94, line 1: I should probably delete this file

lf94:
I need to debug why WalletConnect breaks on Electron builds.

lf94:
Tracking these two external PRs:

* https://github.com/WalletConnect/walletconnect-monorepo/pull/940/files
* https://github.com/WalletConnect/walletconnect-docs/pull/167

nop33:
Is this WIP (judging by the comments) or should it be reviewed?

src/modals/SendModal/TransactionForm.tsx
```diff
@@ -43,13 +43,17 @@ const SendModalTransactionForm = ({ data, onSubmit, onCancel }: TransactionFormP
...
   const [toAddress, setToAddress] = useState(data?.toAddress ?? '')
+  const [addressError, setToAddressError] = useState('')
   const [amount, setAmount] = useState(data?.amount ?? '')
   const [gasAmount, setGasAmount] = useState(data?.gasAmount ?? '')
-  const [gasPrice, setGasPrice] = useState(data?.gasPrice ?? '')
-  const [addressError, setToAddressError] = useState('')
   const [gasAmountError, setGasAmountError] = useState<ReactNode>()
-  const minimalGasPriceInALPH = formatAmountForDisplay(MINIMAL_GAS_PRICE)
+  const minimalGasPriceInALPH = formatAmountForDisplay(MINIMAL_GAS_PRICE, true)
+  const [gasPrice, setGasPrice] = useState(data?.gasPrice ?? minimalGasPriceInALPH)
```
nop33, line 51: Why not leaving it blank as it was before (`''`) to let the node specify the fee?

src/modals/SendModal/TransactionForm.tsx
```diff
@@ -131,10 +139,10 @@ const SendModalTransactionForm = ({ data, onSubmit, onCancel }: TransactionFormP
           </InfoBoxStyled>
         )}
       </ModalContent>
-      <ExpandableSectionStyled sectionTitleClosed="Advanced settings">
+      <ExpandableSectionStyled sectionTitleClosed="Gas">
         <Input
           id="gas-amount"
-          placeholder="Gas amount"
+          placeholder="Amount"
```
nop33, line 145: I'd leave it as it was, same below. Better be explicit, it doesn't cost us much real estate imo

src/modals/SendModal/TransactionForm.tsx
```diff
@@ -156,6 +164,25 @@ const SendModalTransactionForm = ({ data, onSubmit, onCancel }: TransactionFormP
...
+          value={contractState}
+          onChange={(e) => setContractState(e.target.value)}
+        />
+        <Input
+          id="issue-token-amount"
+          placeholder="Tokens to issue"
+          value={issueTokenAmount}
+          type="number"
+          onChange={(e) => setIssueTokenAmount(e.target.value)}
+        />
```
nop33, line 184: Just to understand this properly, does every call to smart contract going to issue new tokens? :/

src/modals/SendModal/index.tsx
```diff
@@ -59,40 +74,59 @@ const SendModal = ({ onClose }: SendModalProps) => {
...
 
   useEffect(() => {
-    if (step === 1) {
-      setTitle('Send')
-    } else if (step === 2) {
-      setTitle('Info Check')
-    } else if (step === 3) {
-      setTitle('Password Check')
-    }
+    setTitle(stepToTitle.get(step) || 'Unknown')
```
nop33, line 93: Are you sure this works fine when going back and forth the steps (pressing the back button)? `'Unknown'` should never happen, we don't want to display that.

src/modals/SendModal/index.tsx
```diff
@@ -59,40 +74,59 @@ const SendModal = ({ onClose }: SendModalProps) => {
...
+    if (isContractTx) {
+      if (transactionData.amount == '') {
+        amount = '0.000001'
+      }
+    }
+    if (isScriptTx) {
+      if (transactionData.amount == '') {
+        amount = '0'
+      }
+    }
```
nop33, line 118:
```suggestion
    if (isContractTx && transactionData.amount == '') {
      amount = '0.000001'
    }
    if (isScriptTx && transactionData.amount == '') {
      amount = '0'
    }
```

src/modals/SendModal/index.tsx
```diff
@@ -59,40 +74,59 @@ const SendModal = ({ onClose }: SendModalProps) => {
...
+    const { fromAddress, toAddress, gasAmount, gasPrice, contractCode, contractState, issueTokenAmount, script } =
+      transactionData
+    let { amount } = transactionData
+
+    const isContractTx = contractCode !== ''
+    const isScriptTx = script !== ''
+
+    if (isContractTx) {
+      if (transactionData.amount == '') {
+        amount = '0.000001'
```
nop33, line 111: Maybe we should extract this value as a constant somewhere?

src/modals/WalletConnectModal.tsx
```diff
@@ -0,0 +1,194 @@
...
+  Proposal,
+  RequireUnlock
+}
+
+const WalletConnectModal = ({ onClose, onConnect }: Props) => {
+  const { addresses } = useAddressesContext()
+  const { walletConnect } = useWalletConnectContext()
+  const [uri, setUri] = useState('')
+  const [state, setState] = useState(addresses.length > 0 ? State.InitiateSession : State.RequireUnlock)
+  const [proposal, setProposal] = useState<SessionTypes.Proposal | undefined>(undefined)
```
nop33, line 47:
```suggestion
  const [proposal, setProposal] = useState<SessionTypes.Proposal>()
```

src/modals/WalletConnectModal.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  useEffect(() => {
+    walletConnect?.on(CLIENT_EVENTS.session.proposal, async (proposal: SessionTypes.Proposal) => {
+      setProposal(proposal)
+      setState(State.Proposal)
+    })
+
+    walletConnect?.on(CLIENT_EVENTS.session.created, onClose)
+  }, [onClose, walletConnect])
```
nop33, line 56: Shouldn't there be an unsubscribe return method here?

src/modals/WalletConnectModal.tsx
```diff
@@ -0,0 +1,194 @@
...
+}
+
+export default WalletConnectModal
+
+const WalletConnectTitle = styled.img`
+  width: 12rem;
+`
+
+const Notice = styled.div`
+  border: solid 1px #ccc;
```
nop33, line 173: don't we have a theme var for this? (and the following color code)

nop33 (commented):
Very good job @lf94 👏 

Man, that was a humongous PR, took me hours to just go through the code...

I still need to checkout the code and test it on my machine

nop33:
As a future recommendation, I think (@mvaivre would agree), let's try to keep PRs less than 1000 LOC and make small and iterative PRs. Even if a feature is not fully completed in a single PR.

lf94:
> As a future recommendation, I think (@mvaivre would agree), let's try to keep PRs less than 1000 LOC and make small and iterative PRs. Even if a feature is not fully completed in a single PR. angel

I agree. It's why I decided to draw the line at getting it minimally working, with no fancy display of contract or script details. I couldn't really be simpler... Sorry it was larger than 1000 LOC but I don't think it really could have been smaller.

nop33:
It could have been done in smaller steps imo. You could have created the context without using it. Next PR could update the context if needed. Then another could update the send modal. Another could create the Wallet connect modal. 

I know it's more work for you, but I think now it's more work for everyone 😅

src/modals/WalletConnectModal.tsx
lf94: Yeah, the styling for the whole modal is undetermined; it's mostly all placeholder until Mika can get time on it

lf94:
Oh it's ok, I don't care if it would be more work, but the context was heavily dependent on how WalletConnect was being used by the rest of the application. It's all pretty tightly coupled. I don't have enough godly foresight to know what the context would've needed before using it...

The modal is useless by itself because it just establishes a connection but does nothing else - really the "juicy" part of the PR is the SendModal and walletconnect context talking to each other.

The PR is about 400 LOC of code - 1800 additions is from the `package-lock.json` being changed. It's really hard for me to imagine this PR being any smaller without just being a bunch of half-thought PRs.

src/modals/SendModal/index.tsx
lf94: Yeah I'd like that... to `utils/constants`!

src/contexts/walletconnect.tsx
```diff
@@ -0,0 +1,153 @@
...
+  requestEvent: undefined
+}
+
+export const Context = createContext<ContextType>(initialContext)
+
+export const WalletConnectContextProvider: FC = ({ children }) => {
+  const { setIsSendModalOpen, settings } = useGlobalContext()
+  const { addresses } = useAddressesContext()
+  const [isWalletConnectModalOpen, setIsWalletConnectModalOpen] = useState(false)
+  const [walletConnect, setWalletConnect] = useState<WalletConnectClient | undefined>(undefined)
```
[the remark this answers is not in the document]
lf94: Ah if it internally does `T | undefined` then that is way nicer... Yeah, the types are quite long when you have to type out `| undefined` for the states.

src/contexts/walletconnect.tsx
lf94: It comes from WalletConnect - you have to get one from them, I guess for their relay to work properly and to avoid spam on their network...

src/contexts/walletconnect.tsx
```diff
@@ -0,0 +1,153 @@
...
+        case 'alephium_getServices': {
+          await walletConnect.respond({
+            topic,
+            response: {
+              id,
+              jsonrpc: '2.0',
+              result: settings.network
+            }
+          })
+          return
```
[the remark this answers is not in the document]
lf94: Nope, `break` are not obligatory in `switch-case` blocks!

src/contexts/walletconnect.tsx
```diff
@@ -0,0 +1,153 @@
...
+          }
+          setDappTransactionData(txData)
+          setIsSendModalOpen(true)
+          break
+        }
+        default:
+          console.warn('Unknown request given')
+          break
+      }
+    }
```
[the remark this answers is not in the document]
lf94: I like it better as if-else too.

src/modals/SendModal/CheckTransaction.tsx
```diff
@@ -56,9 +57,7 @@ const SendModalCheckTransaction = ({ data, fees, onSend, onCancel }: SendModalCh
         <ModalFooterButton secondary onClick={onCancel}>
           Cancel
         </ModalFooterButton>
-        <ModalFooterButton onClick={onSend} disabled={!isSendButtonActive}>
```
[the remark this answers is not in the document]
lf94: So this was a particular thing I had to think about a bit.

Consider the empty truth table below:

|                      |fromAddress|toAddress|amount|gasAmount|contractCode|contractState|script|issueTokenAmount|
|----------------------|-----------|---------|------|---------|------------|-------------|------|----------------|
|isSendButtonActive    |           |         |      |         |            |             |      |                |

Given all these parameters - when should `isSendButtonActive` be true?

All that's actually required is:
* `fromAddress`, `gasAmount`, and `script` for script case.
* `fromAddress`, `gasAmount`, `amount` for contract case.
* `fromAddress`, `toAddress`, `amount` for regular case.

We do validation on the TransactionForm view more or less already.

It's also why I renamed it from "Info Check" to "Review" - the user must review what they've entered, but what they enter may very well be valid based on a combination of parameters.

So I considered the best thing is to pass this to the accepting public node in the end if there is anything actually wrong, but for anything else, it's pretty much impossible for us to account for all user intent now with the introduction of contracts and scripts.

src/modals/SendModal/TransactionForm.tsx
lf94: Because if I remember correctly, when submitting a contract or script, it just wouldn't work properly with the values specified by the node.

src/modals/SendModal/TransactionForm.tsx
lf94: I found it was redundant, having "Gas" twice in close proximity. @mvaivre what do you think of cases like this? Leave the redundancy?

src/modals/SendModal/TransactionForm.tsx
lf94: Nope - you can specify 0 if you want, but yes it happens specifically for contracts.

src/modals/SendModal/index.tsx
lf94: It should never happen but the TypeScript type checker was not happy (in all fairness) because `.get()` can be returned undefined if the key is not found. I don't think there's a way to avoid this? Even with a POJO, it's possible to get `undefined`... But maybe TypeScript will allow me to be strict in that the object will never have it, i.e. `const stepToTitle: StepToTitle = { [Step.Send]: 'Send', ... }; type StepToTitle = { [k: Step]: string };` might not work though because `k` must be a `string` or `Symbol`

src/modals/SendModal/CheckTransaction.tsx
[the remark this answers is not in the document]
nop33: > We do validation on the TransactionForm view more or less already.

That's a good point. I just want to avoid enabling buttons when there is missing info.

> All that's actually required is:
>
> - fromAddress, gasAmount, and script for script case.
> - fromAddress, gasAmount, amount for contract case.
> - fromAddress, toAddress, amount for regular case.

Well, that's easy to code then, isn't it?

```js
const isSendButtonActive = (fromAddress && gasAmount && script) || (fromAddress && gasAmount && amount) || (fromAddress && toAddress && amount)
```

Anyway, I'll verify this when I do my local tests.

src/modals/SendModal/TransactionForm.tsx
nop33: Have you reported this to the backend team? I'd be interested to see the conversation

src/modals/SendModal/TransactionForm.tsx
nop33: Let's see how it actually looks like.

nop33:
> The PR is about 400 LOC of code - 1800 additions is from the package-lock.json being changed. It's really hard for me to imagine this PR being any smaller without just being a bunch of half-thought PRs.

You are absolutely right, sorry about that!

nop33:
@lf94 I would like to checkout your code and do some e2e tests on my machine. Can you guide me through the process? I need the voting demo I suppose. Anything else? Also, when I start the dev server `npm run electron-dev` the electron window doesn't open (i suspect wrong port). Thanks! 🌟

src/modals/SendModal/CheckTransaction.tsx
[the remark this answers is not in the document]
mvaivre: Don't spend too much energy on the Send modal tuning. I have the feeling that we'll build a separated dedicated modal for contract related transactions.

src/contexts/addresses.tsx
```diff
@@ -32,12 +32,12 @@ import { getHumanReadableError } from '../utils/api'
...
 import { useGlobalContext } from './global'
 
-export type TransactionType = 'consolidation' | 'transfer' | 'sweep'
+export type TransactionType = 'consolidation' | 'transfer' | 'sweep' | 'contract'
 
 type SimpleTx = {
   txId: string
   fromAddress: string
-  toAddress: string
+  toAddress?: string
```
mvaivre, line 40: This type starts to be a bit to generic IMHO :) 

We could maybe define a `BaseTX` type with all required parameters, which could then be extended as `TransferTX`, `ContractTX` with specific params etc. Something like:

```ts
type BaseTx = {
  txId: string
  fromAddress: string
  timestamp: number
  type: TransactionType
  network: NetworkType
}

type TransferTx = BaseTx & { toAddress: string, amount: bigint }
```

WDYT?

src/modals/SendModal/index.tsx
```diff
@@ -34,14 +35,28 @@ import ConsolidateUTXOsModal from '../ConsolidateUTXOsModal'
 import SendModalCheckTransaction from './CheckTransaction'
 import SendModalTransactionForm from './TransactionForm'
 
-type StepIndex = 1 | 2 | 3
+enum StepIndex {
+  Send = 1,
+  InfoCheck = 2,
+  PasswordCheck = 3
```
mvaivre, line 41: Nit: if we change this (more explicit that using numbers 👌 !), why not string literals, which are lighter to use (and less code when compiled)?
```ts
type StepIndex = 'send' | 'infoCheck' | 'passwordCheck'
```

[Related read](https://szaranger.medium.com/whats-an-enum-65db967d8bb6#:~:text=The%20benefits%20of%20using%20a,is%20used%20in%20other%20files.)

src/contexts/addresses.tsx
mvaivre: Also, let's extract Transaction types to their own file (in the `types` folder)?

This would allow us to add some handy type predicates there, like `isContractTx` or `isScriptTx`.

src/modals/SendModal/index.tsx
```diff
@@ -59,40 +74,59 @@ const SendModal = ({ onClose }: SendModalProps) => {
...
   }
 
   const buildTransaction = async (transactionData: SendTransactionData) => {
-    setTransactionData(transactionData)
+    const { fromAddress, toAddress, gasAmount, gasPrice, contractCode, contractState, issueTokenAmount, script } =
+      transactionData
+    let { amount } = transactionData
+
+    const isContractTx = contractCode !== ''
+    const isScriptTx = script !== ''
```
mvaivre, line 107: If we decide to be a bit more strict with Tx Types (as suggested above, defining `ContractTx`, `TransferTx` etc.), we could take the opportunity to define some [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), like those `isContractTx` and `isScriptTx`.

mvaivre (commented):
Awesome job Lee! Some little comments 🎸

mvaivre:
We've said it already, but I'm really looking forward to trying this PR on testnet :)) for some easy E2E testing.

src/modals/SendModal/CheckTransaction.tsx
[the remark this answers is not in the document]
lf94: > I just want to avoid enabling buttons when there is missing info.

100% - but I think this makes more sense on the TransactionForm, where it can give actual feedback too!

> Don't spend too much energy on the Send modal tuning

Basically my thoughts when implementing - the design is not final in any way.

src/modals/SendModal/TransactionForm.tsx
lf94: Yeah I did - the conversation is on Slack but it was pretty brief; to paraphase: "contract gas estimation is non-trivial" . 

But you bring up a good point of focus: this shouldn't always be set by default - only if dappTransactionData has contract data.

src/contexts/addresses.tsx
lf94: Yerp :+1:

src/modals/SendModal/index.tsx
lf94: The only issue with this is the steps can't be iterated vs an enum which constructs both a value and type.

For this context though yea, I think using string literal type is better DevX :slightly_smiling_face:

src/modals/WalletConnectModal.tsx
nop33: Maybe add a TODO in the code then, that'd be nice 😇

src/contexts/addresses.tsx
lf94: I did rebase and I didnt see it renamed?.. :s

src/modals/SendModal/index.tsx
lf94: This is a bit less trivial than it appears: pending tx is for any type of pending tx, where some properties are optional. There's going to be a bunch of refactoring that'll be needed. Let's push it to another PR.

src/contexts/walletconnect.tsx
lf94: Added a better todo, but it should be done later and not important now.

lf94:
Someone will probably hit the bad transaction error: it just means you need to fiddle with the gas amount and price.

public/electron.js
```diff
@@ -127,7 +127,9 @@ function createWindow() {
     minHeight: 700,
     titleBarStyle: process.platform === 'win32' ? 'default' : 'hidden',
     webPreferences: {
-      spellcheck: true
+      spellcheck: true,
+      nodeIntegration: true,
+      contextIsolation: false
```
nop33, line 132: I read the docs of these 2 options, but I didn't rly get why they are needed. Could you give me a short explanation as to why you added them? Thank you!

https://www.electronjs.org/docs/latest/api/browser-window

nop33 (commented):
I am getting an error in the console:

```
index.js:1 {context: 'client'} {context: 'client/pairing'} 'Unauthorized pairing update request'
```

nop33:
Regarding the failing macOS electron test: I see this in the logs:
```
    prebuild-install http 404 https://github.com/JoshuaWise/better-sqlite3/releases/download/v7.5.0/better-sqlite3-v7.5.0-electron-v98-darwin-arm64.tar.gz
```

When looking into the [releases of better-sqlite3](https://github.com/JoshuaWise/better-sqlite3/releases) I indeed cannot find it...
```
better-sqlite3-v7.5.0-electron-v98-darwin-arm64.tar.gz --> (what we expect)
better-sqlite3-v7.5.0-electron-v98-darwin-x64.tar.gz   --> (what is available)
```

public/electron.js
lf94: Ah, this can be removed - it was for deeplinking which I gutted because it was just too complicated to include with this PR. I'll remove it

lf94:
Closing as the PR will become stale when Cheng merges his changes.

PR #207, opened 2022-04-13
Tweak AddressBadge colors, change default badge palette

mvaivre:
and bg for overview header

### DIFF
--- a/src/components/AddressBadge.tsx
+++ b/src/components/AddressBadge.tsx
@@ -24,7 +24,7 @@ import Badge from './Badge'
 
 type AddressBadgeProps = ComponentPropsWithoutRef<typeof Badge> & {
   addressName: string
-  seeThroughBg?: boolean
+  opaqueBg?: boolean
 }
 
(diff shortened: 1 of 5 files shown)


src/components/AddressBadge.tsx
```diff
@@ -34,20 +34,25 @@ const AddressBadge = ({ addressName, className, ...props }: AddressBadgeProps) =
 )
 
 export default styled(AddressBadge)`
-  ${({ color, seeThroughBg, theme }) => {
+  ${({ color, opaqueBg, theme }) => {
+    console.log(tinycolor(color).getBrightness())
```
nop33, line 38: Remove logging

nop33 (commented):
Hm, I see that you changed the solid bg color of the badges in the light theme. I think they the labels now are not very visible... There's not enough contrast between the yellow text and yellow bg. WDYT?

If you think it's fine, then I'm fine with it. I personally prefered the solid bgs.

image

image

PR #211, opened 2022-04-13
Update release docs

nop33:
During the release of v1.2.0 today we faced an issue. The issue was that the release workflow was triggered after I performed step 2, but the commit that updated the version of the project from 1.2.0-rc.0 to 1.2.0 was not pushed (that was step 3 of our process). After performing step 3, the new release workflow worked correctly.

This commit updates the docs to avoid this issue in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,11 @@ npm run electron-pack-linux:arm64
 
 To release a new version:
 
-1. Create a commit that updates the package version in package.json and package-lock.json and a tag with:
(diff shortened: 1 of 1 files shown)


lf94 (approved):
Looks good to me. I was thinking of how you'd avoid this with a command which gets the latest tag for you but it is impossible to know if it'll ever be the right one, so manual entry of the tag is needed.

PR #224, opened 2022-04-28
Fix param to cliqueClient.transactionSend

nop33:
Looking at the SDK, I noticed that we are passing the wrong parameter to the [`transactionSend` function](https://github.com/alephium/js-sdk/blob/master/lib/clique.ts#L115).

This didn't matter because so far we did not use the a multi node clique (`isMultiNodesClique` was set to `false` in the desktop wallet), so the result of `getClientIndex` was always `0`, no matter the address we passed to it.

### DIFF
--- a/src/utils/api-clients.ts
+++ b/src/utils/api-clients.ts
@@ -102,22 +102,22 @@ export async function createClient(settings: Settings['network']) {
     }
 
     const signAndSendTransaction = async (
-      address: Address,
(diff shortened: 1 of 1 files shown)


lf94:
I don't think this is a fix, but more of an improvement: I would expect that `toHash == fromAddress.hash` always. Nice catch.

lf94 (approved):

Issue #210, opened 2022-04-13
Switching from offline mode to online mode does not automatically refresh the balances


nop33:
**Describe the bug**
Switching from offline mode to online mode does not automatically refresh the balances. Clicking the refresh button does correctly fetch the balances.

**To Reproduce**
1. Open wallet
2. Start with offline mode (change network settings to blank)
3. Login into the wallet
4. Change network settings to an online network
5. Balances are not refreshed

**Expected behavior**
Balances are refreshed

PR #230, opened 2022-05-05
Refresh on reconnect to a network

lf94:
Resolves https://github.com/alephium/desktop-wallet/issues/210

Took 4 hours to work out the execution flow and create a solution which caused renders at the appropriate times.

There was a logic error which ran deeper than just "refresh balances on reconnect". Our wallet assumed on network change it's already connected which is impossible: a network is not connected to when it's URL is switched, but when it tries to make an attempt at a request. You can see this fixed in the PR and essentially is the core problem. I'm noting it here because it was a very subtle piece of logic and once realized, was easy to correct.


src/modals/SettingsModal/NetworkSettingsSection.tsx
```diff
@@ -72,6 +72,7 @@ const NetworkSettingsSection = () => {
   const handleNetworkPresetChange = useCallback(
     (option: typeof networkSelectOptions[number] | undefined) => {
       if (option && option.value !== selectedNetwork) {
+        setIsOffline(true)
```
nop33, line 75: > There was a logic error which ran deeper than just "refresh balances on reconnect". Our wallet assumed on network change it's already connected which is impossible: a network is not connected to when its URL is switched, but when it tries to make an attempt at a request. You can see this fixed in the PR and essentially is the core problem. I'm noting it here because it was a very subtle piece of logic and once realized, was easy to correct.

The logic I implemented was that:

> Assume we are online, unless a request fails, then mark as offline until a request succeeds then mark as online again

Your fix is a good solution against the problem that the issue refers to but introduces another problem, that of the flickering icon on the top right as well as the error message at the bottom right. Looking at your code I see that you mark the network as being offline the moment the user switches the network settings and it turns back online when the request to initialize the client succeeds.

To address this issue, I'd like to suggest holding the state of the network in a new property of the global context:

```ts
networkStatus: 'online' | 'offline' | 'connecting'
```

Then, instead of setting the network to `offline` when the user switches the network, we can set it to `connected` until the first request to initialize the client takes place. This way, we avoid the flickering of the icon and the momentary error message.

```suggestion
        setNetworkStatus('connecting')
```

WDYT?

nop33: Moreover, updating the network status is essentially a sideeffect of updating the network settings:
```ts
updateSettings('network', ...)
```

So, the network status should be updated also before line 93.

_(Ilias' "aha" moment: I see now how `useReduce` (or Redux) would help here... 👀 )_

nop33: Instead of exposing `setNetworkStatus` to the global context, maybe it'd be better to just expose this new function:

```ts
  const updateNetworkSettings(newNetworkSettings) {
    setNetworkStatus('connecting')
    updateSettings('network', newNetworkSettings)
  }
```

lf94: My concern is there are many implications which your solution doesn't consider that I'm going to have to look at and experiment with; I would like to focus on the fact the issue resolves a complicated problem first, then improve the UX in another PR.

lf94: Yep, I think so too. In all instances of a network settings change that needs to happen.

nop33 (approved):
Hey @lf94, since I had the implementation already prepared in my head after working on it on the mobile wallet, I just went ahead and implemented it. Check out my solution and feel free to merge if you agree with it (I pushed in this PR). Thanks!

Issue #69, opened 2021-12-07
Upgrade `react-router-dom` from v5 to v6


nop33:
There have been many breaking changes between v5 and v6. Here's a guide explaining the upgrade procedure: https://reactrouter.com/docs/en/v6/upgrading/v5

PR #232, opened 2022-05-06
Migrate react-router from v5 to v6

lf94:
Resolves https://github.com/alephium/desktop-wallet/issues/69


lf94:
This one is going to need some comprehensive clicking around to make sure all paths are still working; I tried myself but it's possible I missed something.

nop33 (changes requested):
Nice job!

I noticed that the back buttons in create/import wallet screens do not work.

nop33 (commented):
Spot the problem :P

image

When this is fixed, all good!

LeeAlephium:
@nop33 should be good now

mvaivre (approved):
LGTM! Thanks!

Issue #175, opened 2022-03-23
Create a test for broken links


mvaivre:
Many links aren't working right now (https://wiki.alephium.org/Desktop-Wallet-Guide.html#utxo-consolidation https://wiki.alephium.org/Desktop-Wallet-Guide.html#creating-a-mining-wallet-with-4-addresses). 

Let's add a test to catch any broken link?

nop33:
These are not working because we haven't published the wiki articles!

nop33:
https://github.com/alephium/wiki/pull/46

lf94:
It seems the links themselves are broken, the appropriate links are:

* https://wiki.alephium.org/wallet/Desktop-Wallet-Guide#utxo-consolidation
* https://wiki.alephium.org/wallet/Desktop-Wallet-Guide#creating-a-mining-wallet-with-4-addresses

lf94:
Creating a test which relies on an external service isn't great; I suggest we just make sure the links are deployed already / near the same time.

mvaivre:
IMO we will never be able to ensure that all the links are updated (and that a release is done) ahead of every article change (especially since other non-dev can be responsible for updating the wiki).
I prefer having a test fail rather than a dead link. 
If a test fails and we realise that it's just a temporary issue, we can simply bypass the test. But it would force us to investigate.

nop33:
I think adding a test will become every easier when we start translating our content. We will then have to move all our content to json or yaml or md files. We can then tell a GitHub Actions script to always parse those files, look for links, and validate them. I guess the same can be done with ts files.

lf94:
Do we truly want to test _all_ links to the Internet?

I can create a test that searches for links in our source code and test them all.

PR #233, opened 2022-05-06
Test for invalid links in the UI

lf94:
Resolves https://github.com/alephium/desktop-wallet/issues/175

This problem made me realize there are even some links which are not part of components which need to be checked, like the ones in the window's toolbar (like the Help option which leads to Discord). The test accounts for them.

It's implemented as a jest test, so there's no need to update the CI to include the test.

### DIFF
--- a/src/tests/links.test.ts
+++ b/src/tests/links.test.ts
@@ -53,11 +53,8 @@ const httpClientOptions = {
 }
 
 function request(link: string, callback) {
-  if (link.match(/^https:\/\//)) {
-    return https.get(link, httpClientOptions, callback)
-  } else {
-    return http.get(link, httpClientOptions, callback)
-  }
+  const client = link.match(/^https:\/\//) ? https : http
+  return client.get(link, httpClientOptions, callback)
 }
 
 function dedup(xs) {
--- a/src/tests/links.test.ts
+++ b/src/tests/links.test.ts
@@ -1,0 +1,98 @@
+/*
+Copyright 2018 - 2022 The Alephium Authors
+This file is part of the alephium project.
+
+The library is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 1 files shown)


src/tests/links.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+function dedup(xs) {
+  const uniques = []
+  for (const x of xs) {
+    if (uniques.indexOf(x) < 0) {
+      uniques.push(x)
+    }
+  }
+  return uniques
+}
```
nop33, line 71: Isn't there a lodash method to do this (`uniq` or sth like that)? Since we use the library, maybe it makes sense to utilize it...

src/tests/links.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  headers: { 'User-Agent': userAgent }
+}
+
+function request(link: string, callback) {
+  if (link.match(/^https:\/\//)) {
+    return https.get(link, httpClientOptions, callback)
+  } else {
+    return http.get(link, httpClientOptions, callback)
+  }
+}
```
nop33, line 61: nit: how about this? 😇 

```suggestion
function request(link: string, callback) {
  const client = link.match(/^https:\/\//) ? https : http
  return client.get(link, httpClientOptions, callback)
}
```

nop33 (commented):
Good idea to implement it as a Jest test! I rly like the implementation!

src/tests/links.test.ts
lf94: I was going to do this but I wasn't comfortable with the side effects it could have internally to `get`. If it worked for you then ok, I'll commit the suggestion :slightly_smiling_face:

src/tests/links.test.ts
nop33: but we already depend on lodash

nop33 (approved):

PR #237, opened 2022-05-09
Fix address metadata storage

### LINKED ISSUE #236: Additional wallet displayed is existing wallets
**Describe the bug**
Given a wallet with the name `walletXYZ`, when opening the wallet, an additional wallet is displayed with name `YZ-addresses-metadata`.

The encrypted data of a wallet are stored in a localStorage entry with the key `wallet-<WALLET-NAME>`. So, the wallet called `walletXYZ` has an entry with the key `wallet-walletXYZ`. The problem is that we [use the `accountName` (aka wallet name) as a prefix](https://github.com/alephium/desktop-wallet/blob/master/src/utils/addresses.ts#L62) to the key of the localStorage entry for storing metadata about the addresses of the wallet. So, the wallet `walletXYZ` has an entry called `walletXYZ-addresses-metadata`.

The [code that checks to find existing wallets](https://github.com/alephium/js-sdk/blob/master/lib/storage-browser.ts#L44-L54) when opening the app is looking for entries in the localStorage that [start with `wallet`](https://github.com/alephium/js-sdk/blob/master/lib/storage-browser.ts#L23). So, it thinks that the `walletXYZ-address-metadata` is a wallet, but it’s not, it’s just another entry to store metadata about this wallet. the character `X` is also omitted from the `YZ-addresses-metadata` because the code expects a `-` to be there, so it removes that character.

The wallet name should have not been used as a prefix but as a suffix, same way the code in the SDK stores data about the wallet.

nop33:
Closes #236

### DIFF
--- a/src/utils/addresses.ts
+++ b/src/utils/addresses.ts
@@ -16,6 +16,8 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
+import { getStorage } from '@alephium/sdk'
+
 import { Address } from '../contexts/addresses'
 
 const addressesMetadataLocalStorageKeyPrefix = 'addresses-metadata'
@@ -75,3 +77,17 @@ export const sortAddressList = (addresses: Address[]): Address[] =>
     if (b.settings.isMain) return 1
     return (b.lastUsed ?? 0) - (a.lastUsed ?? 0)
   })
+
+export const migrateAddressMetadata = () => {
+  const Storage = getStorage()
+  const walletNames = Storage.list()
+
+  for (const name of walletNames) {
+    const data = localStorage.getItem(`${name}-${addressesMetadataLocalStorageKeyPrefix}`)
+
+    if (data) {
+      localStorage.setItem(constructMetadataKey(name), data)
+      localStorage.removeItem(`${name}-${addressesMetadataLocalStorageKeyPrefix}`)
+    }
+  }
+}
--- a/src/utils/addresses.ts
+++ b/src/utils/addresses.ts
@@ -84,11 +84,12 @@ export const migrateAddressMetadata = () => {
   const walletNames = Storage.list()
 
   for (const name of walletNames) {
-    const data = localStorage.getItem(`${name}-addresses-metadata`)
+    const deprecatedKey = `${name}-addresses-metadata`
+    const data = localStorage.getItem(deprecatedKey)
 
     if (data) {
       localStorage.setItem(constructMetadataKey(name), data)
-      localStorage.removeItem(`${name}-${addressesMetadataLocalStorageKeyPrefix}`)
+      localStorage.removeItem(deprecatedKey)
     }
   }
 }
--- a/src/utils/addresses.ts
+++ b/src/utils/addresses.ts
@@ -78,12 +78,13 @@ export const sortAddressList = (addresses: Address[]): Address[] =>
     return (b.lastUsed ?? 0) - (a.lastUsed ?? 0)
   })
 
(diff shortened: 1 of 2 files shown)


lf94:
@nop33 test failed

src/utils/addresses.ts
```diff
@@ -73,3 +77,17 @@ export const sortAddressList = (addresses: Address[]): Address[] =>
     if (b.settings.isMain) return 1
     return (b.lastUsed ?? 0) - (a.lastUsed ?? 0)
   })
+
+export const migrateAddressMetadata = () => {
+  const Storage = getStorage()
+  const walletNames = Storage.list()
+
+  for (const name of walletNames) {
+    const data = localStorage.getItem(`${name}-${addressesMetadataLocalStorageKeyPrefix}`)
```
lf94, line 86: I think this could use a comment - seeing something with the description of prefix being used as a suffix is very confusing.

src/utils/addresses.ts
```diff
@@ -73,3 +77,17 @@ export const sortAddressList = (addresses: Address[]): Address[] =>
     if (b.settings.isMain) return 1
     return (b.lastUsed ?? 0) - (a.lastUsed ?? 0)
   })
+
+export const migrateAddressMetadata = () => {
```
lf94, line 81: It's possible there are more migrations in the future; because of this we now need to make descriptive and composable migrations. It's possible a user can be anywhere in the migration process so each migration task must stay for the application's life.

Here's an example (there are most likely things it's forgetting):

```
function migration20220509_MoveToNewStorageKeyName() { ... }
function migration20220811_AddSomePropertyToAddresses() { ... }

function migrate() {
  migration20220509_MoveToNewStorageKeyName()
  migration20220811_AddSomePropertyToAddresses()
}
```

This is standard for any software which has migration processes.

src/utils/addresses.ts
lf94: The old prefix / name should be local to the migration's function context.

nop33: OK, how about now? :)

src/utils/addresses.ts
nop33: > It's possible a user can be anywhere in the migration process so each migration task must stay for the application's life.

Yes, that's totally true!

I don't think it's necessary to introduce these weird function names on the very first migration function. Maybe it's worth it when there is more than one? I'd leave it as it is now.

lf94: Sure; this'll have to be remembered in the future though by not just you but anyone who may take the issue. :slightly_smiling_face:

lf94 (approved):

src/utils/addresses.ts
```diff
@@ -73,3 +77,18 @@ export const sortAddressList = (addresses: Address[]): Address[] =>
...
+  const Storage = getStorage()
+  const walletNames = Storage.list()
+
+  for (const name of walletNames) {
+    const data = localStorage.getItem(`${name}-addresses-metadata`)
+
+    if (data) {
+      localStorage.setItem(constructMetadataKey(name), data)
+      localStorage.removeItem(`${name}-${addressesMetadataLocalStorageKeyPrefix}`)
```
mvaivre, line 91: Why use once `${name}-addresses-metadata` and then `${name}-${addressesMetadataLocalStorageKeyPrefix}` to identify the old item?

nop33: cause I'm an idiot

lf94: I guess I brain farted on this one too because it didn't register in my head at all :upside_down_face:

nop33: fixed

mvaivre (approved):

PR #240, opened 2022-05-10
Simplify toggle

mvaivre:
Quick fix that keeps the animation, remove a variable and fixes the alignment issues.

image

### DIFF
--- a/src/components/Inputs/Toggle.tsx
+++ b/src/components/Inputs/Toggle.tsx
@@ -29,8 +29,6 @@ interface ToggleProps {
   handleColors?: [string, string]
 }
 
-const toggleMarginInPx = 2
-
 const Toggle = ({ toggled, onToggle, className, disabled, ToggleIcons, handleColors }: ToggleProps) => {
   const theme = useTheme()
   const [toggleWidth, setToggleWidth] = useState(0)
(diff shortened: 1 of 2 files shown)


nop33:
By the way, I noticed that in the explorer the theme toggle looks aligned (and has animation) 😉

mvaivre:
> By the way, I noticed that in the explorer the theme toggle looks aligned (and has animation) 😉

Yeah, it seems that it was mostly due to a pixel rounding issue. If you look closely at previous commits, the toggle in the header looked aligned whereas the exact same component in the settings looked miss-aligned (or the other way around). This should make it more predictable.

nop33 (approved):
Seems to work well 👏

One thing I have noticed (probably not related to this PR) is that the cursor doesn't change to a pointer always... It's quite unpredictable...

https://user-images.githubusercontent.com/1579899/167589846-0c5d87f4-9bd2-4c72-bf98-84bca32ca505.mp4

mvaivre:
Yes, it's covered by the "invisible header" that allows for dragging the window. I witnessed the same thing in other electron apps in the past.. Didn't find a quick way to fix it so far.

nop33:
> Yes, it's covered by the "invisible header" that allows for dragging the window. 

AHAAAA

Issue #17, opened 2021-09-23
Optional BIP39 passphrase at wallet creation and restore


tdroxler:
In order to secure a bit more the mnemonic, we can already do it in the backend https://github.com/alephium/alephium/blob/975b343312a0513d3023399d4761f7ba0fb11fb5/wallet/src/main/scala/org/alephium/wallet/api/model/WalletCreation.scala#L26 basically this passphrase change the seed for a given mnemonic. So if someone steal the 24 words, he would still need to know that passphrase.

polarker:
I am a bit against this proposal. I think that creation should go without a mnemonic and restorisaztion should go with a provided mnemonic always. Am I missing anything?

nop33:
@tdroxler you are referring to the so-called 25th word (aka passphrase) which given the 24 words it creates a new wallet, correct? (similar to Ledger: https://www.ledger.com/academy/passphrase-an-advanced-security-feature)

Since this would be an advanced feature, an idea would be that it doesn't need to be part of the wallet creation workflow. It could be found in the settings of the wallet (same as in Ledger).

polarker:
Sorry, forgot my brain fart 😓 It was about mnemonic passphrase!

nop33:
I have many questions on how to implement this!

Supposing we want a similar implementation as [Ledger's hardware wallet](https://www.ledger.com/academy/passphrase-an-advanced-security-feature), where using a passphrase essentially "unlocks" a new wallet.

## Creating a passphrase

I suggest that the creation of a passphrase takes place *AFTER* the initial wallet has been created.

| Step 1 | Step 2 |
| ------ | ------ |
| We could add a button in the "Accounts Settings" modal | Clicking the button will open another modal. Here, the user enters their passphrase and the password that they can use to log in to their wallet. |
| ![image](https://user-images.githubusercontent.com/1579899/148803472-02b72f41-1940-4d33-98a4-31be6110f68b.png) | ![image](https://user-images.githubusercontent.com/1579899/148804335-318c0d50-5475-4b95-938f-28666f9ab49d.png) |

## Accessing the new wallet

Nothing changes here. The user can simply choose their normal account and enter the new password which will unlock the new wallet.

## Restoring wallet with passphrase

I haven't thought this one through yet, but I think having the above things implemented is enough. All the user has to do to restore a wallet with 24 words + a passphrase is to:
1. Restore the wallet normally, using their 24 words
2. Access the Accounts Settings page
[...]

Issue #136, opened 2022-03-02
Automatic address discovery


mvaivre:
Address discovery should be triggered in two cases:

1. When restoring a wallet
2. Manually, for cases where a user is using a same account on multiple devices

This should follow the strategy defined in BIP44 and should be implemented on the backend by @tdroxler (following discussions with @polarker). 

Details must be discussed, but this feature will be considered in my future mockups. 
This will ship *after* 1.2.0 (address derivation). Before that, user will need to "rederive" addresses manually.

mvaivre:
@tdroxler This one would improve our UX a lot! Do you think we could increase priority and start looking at it / discussing it? Thanks :)

tdroxler:
Sure, let's have a call soon.

tdroxler:
I just checked which endpoint should be called and I think I'll rather add a new one `/addresses/<address>/active` which return `true/false` if a tx exists for that address. All other endpoint are way to heavy for that task, while we can find a super light sql query to do that check.

LeeAlephium:
May want to even make it take a range instead to reduce the network requests? `/addresses/active?offset=0&range=24` so the typical case of 4 addresses plus 20 unused

mvaivre:
The endpoint mentioned by Thomas is for one address only. It basically simply check if TX were done with that address.

tdroxler:
> May want to even make it take a range instead to reduce the network requests? `/addresses/active?offset=0&range=24` so the typical case of 4 addresses plus 20 unused, and can return a `bool[]`

we can't do this from the backend as we don't have access to the key used to derive next addresses

LeeAlephium:
@tdroxler user can pass a set of addresses then instead as a json array: `/addresses/active` payload: `['0xoeauaoueao', '0xsaotehusoatehuoatheua', ...]`

nop33:
> Otherwise we're going to have an initial 80 requests - 20 requests per group...

That's only true if the first 80 address indexes have 20 addresses per group. It will definitely be more than that, since we are scanning sequentially.

nop33:
I run a script that iterates through address indexes sequentially until it finds 20 addresses per group. I run it 100 times, and the average total addresses that had to be discovered is `109`.

<details><summary>Click here for source code</summary>

```js
const sdk = require("@alephium/sdk");
const wallet = sdk.walletGenerate();

const totals = [];

for (let i = 0; i < 100; i++) {
  const group0addresses = [];
  const group1addresses = [];
[...]

PR #244, opened 2022-05-19
BIP39 support

lf94:
I have gone with an approach that does not follow the original proposal. The original proposal does not support BIP39 but instead proposed its own decoy method (which is fine but is not as resilient as BIP39, which is much more preferable by everyone). This PR addresses everyone's concerns :slightly_smiling_face: 

Here is how it works. It fully retains plausible deniability while being user friendly. Note that the passphrase is considered an advanced feature but future iterations will make it less demanding.

* Wallet creation and importing remain the same. desktop-wallet encrypts and saves the wallet (mnemonic, addresses, etc) to disk. **This is the default decoy.**
* Wallet unlocking is different
  * If no passphrase is entered, the decoy is presented.
  * If passphrase is entered, a wallet is derived (based on the mnemonic from the wallet selected plus the passphrase), which may or may not have funds. **Other decoys may be derived here.**

And that's it! It means the advanced users will have to re-discover addresses but at least they retain full plausible deniability. When we tackle address discovery this will no longer be a concern either.

Resolves https://github.com/alephium/desktop-wallet/issues/17

### DIFF
--- a/src/contexts/addresses.tsx
+++ b/src/contexts/addresses.tsx
@@ -34,7 +34,7 @@ import {
   loadStoredAddressesMetadataOfAccount,
   storeAddressMetadataOfAccount
 } from '../utils/addresses'
-import { stringToDoubleSHA215HexString } from '../utils/misc'
+import { stringToDoubleSHA256HexString } from '../utils/misc'
 import { NetworkName } from '../utils/settings'
 import { useGlobalContext } from './global'
 
@@ -186,7 +186,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
   )
 
   const getAccountKey = useCallback(
-    () => stringToDoubleSHA215HexString(`${currentAccountName}-${passphraseDoubleHashed}`),
+    () => stringToDoubleSHA256HexString(`${currentAccountName}-${passphraseDoubleHashed}`),
     [currentAccountName, passphraseDoubleHashed]
   )
 
--- a/src/contexts/addresses.tsx
+++ b/src/contexts/addresses.tsx
@@ -212,13 +212,15 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
   const updateAddressSettings = useCallback(
     (address: Address, settings: AddressSettings) => {
       if (!wallet) return
-      storeAddressMetadataOfWallet({
-        mnemonic: wallet.mnemonic,
-        activeWalletName: activeWalletName,
-        index: address.index,
-        settings,
-        passphraseHash
-      })
+      storeAddressMetadataOfWallet(
+        {
+          mnemonic: wallet.mnemonic,
+          walletName: activeWalletName,
+          passphraseHash
+        },
+        address.index,
+        settings
+      )
       address.settings = settings
       setAddress(address)
     },
@@ -294,13 +296,15 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
   const saveNewAddress = useCallback(
     (newAddress: Address) => {
       if (!wallet) return
-      storeAddressMetadataOfWallet({
-        mnemonic: wallet.mnemonic,
-        walletName: activeWalletName,
-        index: newAddress.index,
-        settings: newAddress.settings,
-        passphraseHash
-      })
+      storeAddressMetadataOfWallet(
+        {
+          mnemonic: wallet.mnemonic,
+          walletName: activeWalletName,
+          passphraseHash
+        },
+        newAddress.index,
+        newAddress.settings
+      )
       setAddress(newAddress)
       fetchAndStoreAddressesData([newAddress])
     },
@@ -377,17 +381,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
       initializeCurrentNetworkAddresses()
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [
-    currentNetwork,
-    networkStatus,
-    client,
-    activeWalletName,
-    getAccountKey,
-    wallet,
-    explorerApiHost,
-    nodeHost,
-    usedPassphrase
-  ])
+  }, [currentNetwork, networkStatus, client, activeWalletName, wallet, explorerApiHost, nodeHost])
 
   // Whenever the addresses state updates, check if there are pending transactions on the current network and if so,
   // keep querying the API until all pending transactions are confirmed.
--- a/src/contexts/addresses.tsx
+++ b/src/contexts/addresses.tsx
@@ -34,7 +34,6 @@ import {
   loadStoredAddressesMetadataOfAccount,
   storeAddressMetadataOfAccount
 } from '../utils/addresses'
-import { stringToDoubleSHA256HexString } from '../utils/misc'
 import { NetworkName } from '../utils/settings'
 import { useGlobalContext } from './global'
 
@@ -217,11 +216,17 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
   const updateAddressSettings = useCallback(
     (address: Address, settings: AddressSettings) => {
       if (!wallet) return
-      storeAddressMetadataOfAccount(wallet.mnemonic, activeWalletName, passphraseHash, address.index, settings)
+      storeAddressMetadataOfWallet({
+        mnemonic: wallet.mnemonic,
+        activeWalletName: activeWalletName,
+        index: address.index,
+        settings,
+        passphraseHash
+      })
       address.settings = settings
       setAddress(address)
     },
-    [setAddress, wallet]
+    [setAddress, wallet, currentAccountName, passphraseHash]
   )
 
   const fetchAndStoreAddressesData = useCallback(
@@ -293,11 +298,17 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
   const saveNewAddress = useCallback(
     (newAddress: Address) => {
       if (!wallet) return
-      storeAddressMetadataOfAccount(wallet.mnemonic, activeWalletName, passphraseHash, newAddress.index, newAddress.settings)
+      storeAddressMetadataOfWallet({
+        mnemonic: wallet.mnemonic,
+        walletName: activeWalletName,
+        index: newAddress.index,
+        settings: newAddress.settings,
+        passphraseHash
+      })
       setAddress(newAddress)
       fetchAndStoreAddressesData([newAddress])
     },
-    [fetchAndStoreAddressesData, setAddress, wallet]
+    [fetchAndStoreAddressesData, setAddress, wallet, currentAccountName, passphraseHash]
   )
 
   const generateOneAddressPerGroup = (labelPrefix: string, labelColor: string, skipGroups: number[] = []) => {
@@ -324,7 +335,11 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
       console.log('🥇 Initializing current network addresses')
       if (!activeWalletName || !wallet) return
 
-      const addressesMetadata = loadStoredAddressesMetadataOfAccount(wallet.mnemonic, activeWalletName, passphraseHash)
+      const addressesMetadata = loadStoredAddressesMetadataOfAccount({
+        mnemonic: wallet.mnemonic,
+        walletName: activeWalletName,
+        passphraseHash
+      })
 
       if (addressesMetadata.length === 0) {
         saveNewAddress(
--- a/src/contexts/addresses.tsx
+++ b/src/contexts/addresses.tsx
@@ -162,7 +162,8 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
     settings: {
       network: { nodeHost, explorerApiHost }
     },
-    networkStatus
+    networkStatus,
+    usedPassphrase
   } = useGlobalContext()
   const previousWallet = useRef<Wallet | undefined>(wallet)
   const previousNodeApiHost = useRef<string>()
@@ -315,7 +316,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
 
       const addressesMetadata = loadStoredAddressesMetadataOfWallet(activeWalletName)
 
-      if (addressesMetadata.length === 0) {
+      if (usedPassphrase || addressesMetadata.length === 0) {
         saveNewAddress(
           new Address(wallet.address, wallet.publicKey, wallet.privateKey, 0, {
             isMain: true,
@@ -355,7 +356,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
       initializeCurrentNetworkAddresses()
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [currentNetwork, networkStatus, client, activeWalletName, wallet, explorerApiHost, nodeHost])
+  }, [currentNetwork, networkStatus, client, activeWalletName, wallet, explorerApiHost, nodeHost, usedPassphrase])
 
   // Whenever the addresses state updates, check if there are pending transactions on the current network and if so,
   // keep querying the API until all pending transactions are confirmed.
--- a/src/contexts/addresses.tsx
+++ b/src/contexts/addresses.tsx
@@ -91,8 +91,8 @@ export class Address {
     return this.settings.label || this.shortHash
   }
 
-  getLabelName() {
-    return `${this.settings.isMain ? '★ ' : ''}${this.getName()}`
+  getLabelName(showStar = true) {
+    return `${this.settings.isMain && showStar ? '★ ' : ''}${this.getName()}`
   }
 
   addPendingTransaction(transaction: SimpleTx) {
@@ -128,7 +128,7 @@ export interface AddressesContextProps {
   updateAddressSettings: (address: Address, settings: AddressSettings) => void
   refreshAddressesData: () => void
   fetchAddressTransactionsNextPage: (address: Address) => void
-  generateOneAddressPerGroup: (labelPrefix: string, color: string, skipGroups?: number[]) => void
+  generateOneAddressPerGroup: (labelPrefix?: string, color?: string, skipGroups?: number[]) => void
   isLoadingData: boolean
 }
 
@@ -311,10 +311,11 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
     [fetchAndStoreAddressesData, setAddress, wallet, activeWalletName, passphraseHash]
   )
 
-  const generateOneAddressPerGroup = (labelPrefix: string, labelColor: string, skipGroups: number[] = []) => {
+  const generateOneAddressPerGroup = (labelPrefix?: string, labelColor?: string, skipGroups: number[] = []) => {
     if (!wallet?.seed) return
 
     const skipAddressIndexes = addressesOfCurrentNetwork.map(({ index }) => index)
+    const hasLabel = !!labelPrefix && !!labelColor
     Array.from({ length: TOTAL_NUMBER_OF_GROUPS }, (_, group) => group)
       .filter((group) => !skipGroups.includes(group))
       .map((group) => ({ ...deriveNewAddressData(wallet.seed, group, undefined, skipAddressIndexes), group }))
@@ -322,8 +323,8 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
         saveNewAddress(
           new Address(address.address, address.publicKey, address.privateKey, address.addressIndex, {
             isMain: false,
-            label: `${labelPrefix} ${address.group}`,
-            color: labelColor
+            label: hasLabel ? `${labelPrefix} ${address.group}` : '',
+            color: hasLabel ? labelColor : ''
           })
         )
       })
--- a/src/contexts/addresses.tsx
+++ b/src/contexts/addresses.tsx
@@ -163,7 +163,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
       network: { nodeHost, explorerApiHost }
     },
     networkStatus,
-    passphraseHash
+    isPassphraseUsed
   } = useGlobalContext()
   const previousWallet = useRef<Wallet | undefined>(wallet)
   const previousNodeApiHost = useRef<string>()
@@ -212,19 +212,20 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
   const updateAddressSettings = useCallback(
     (address: Address, settings: AddressSettings) => {
       if (!wallet) return
-      storeAddressMetadataOfWallet(
-        {
-          mnemonic: wallet.mnemonic,
-          walletName: activeWalletName,
-          passphraseHash
-        },
-        address.index,
-        settings
-      )
+
+      if (!isPassphraseUsed)
+        storeAddressMetadataOfWallet(
+          {
+            mnemonic: wallet.mnemonic,
+            walletName: activeWalletName
+          },
+          address.index,
+          settings
+        )
       address.settings = settings
       setAddress(address)
     },
-    [setAddress, wallet, activeWalletName, passphraseHash]
+    [wallet, activeWalletName, isPassphraseUsed, setAddress]
   )
 
   const fetchAndStoreAddressesData = useCallback(
@@ -296,19 +297,20 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
   const saveNewAddress = useCallback(
     (newAddress: Address) => {
       if (!wallet) return
-      storeAddressMetadataOfWallet(
-        {
-          mnemonic: wallet.mnemonic,
-          walletName: activeWalletName,
-          passphraseHash
-        },
-        newAddress.index,
-        newAddress.settings
-      )
+
+      if (!isPassphraseUsed)
+        storeAddressMetadataOfWallet(
+          {
+            mnemonic: wallet.mnemonic,
+            walletName: activeWalletName
+          },
+          newAddress.index,
+          newAddress.settings
+        )
       setAddress(newAddress)
       fetchAndStoreAddressesData([newAddress])
     },
-    [fetchAndStoreAddressesData, setAddress, wallet, activeWalletName, passphraseHash]
+    [wallet, isPassphraseUsed, activeWalletName, setAddress, fetchAndStoreAddressesData]
   )
 
   const generateOneAddressPerGroup = (labelPrefix?: string, labelColor?: string, skipGroups: number[] = []) => {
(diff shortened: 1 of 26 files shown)


tdroxler:
I personally like this way of using the passphrase, but we need to be aware that we'll receive complains from the community. Because we had this in the node-wallet in the beginning and a lot of people we're crying that they unlocked their wallet and had 0 Alph, because they didn't realize or remember they put a passphrase.

But maybe here it's different because the passphrase isn't set on creation right?

Maybe it's the time to teach the community how to properly use BIP39? we could write a blog post and explain the benefits of those decoy wallets.

nop33:
I am not sure the user-flows are what @mvaivre and I had in mind. I see that you first create a wallet, then you lock it, and then you attach a passphrase to it. I think we wanted to go for adding the passphrase directly in the wallet creating/import flow.

mvaivre:
I'd rather keep the onboarding as simple as possible and let users add this level of security as a second step.

The reasoning is similar to what we are doing on the mobile, where the onboarding is even more direct: wallet creation doesn't require a backup. After the onboarding, users are notified about the need to backup their mnemonic. We're removing UX blockers by minimizing the mental load required to setup a wallet. 

I wouldn't advertise this feature too much at first, I'd like to see how it's received and used by our community.

polto:
I love how this works. Looks very much similar to the workflow of Ledger or a few others good hardware wallets. First 24 words, than optionally a different HD wallet with a passphrase.

tdroxler:
we could add later in the advanced option a: `read more about this` link and redirect either to one of our article or directly to BIP39 if we don't have an article.

nop33 (commented):
> For what it's worth I noticed another consequence for all decoy methods: they can't have address labeling.

Another problem I experienced is that unlocking a wallet by providing a passphrase will effectively override the labels of the decoy wallet. In the example below, in my decoy wallet I had set the label `Vault` to the address with index `0`, and `Main` to the address with index `4` (as well as making the latter the main address). After unlocking a passphrase wallet and logging back in to my decoy wallet, I noticed that the metadata of the address with index 0 were overriden. Now, I lost my label, and I have 2 main addresses.

We need to properly think how the labeling system can be adjusted.

image

src/contexts/addresses.tsx
```diff
@@ -319,7 +320,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
 
       const addressesMetadata = loadStoredAddressesMetadataOfAccount(currentAccountName)
 
-      if (addressesMetadata.length === 0) {
+      if (usedPassphrase || addressesMetadata.length === 0) {
```
nop33, line 323: By removing this newly added condition we could have shared labels between decoy and passphrase wallets. Not sure we would want that though. However, we cannot proceed without finding a solution to the labeling issue I mentioned in my other comment.

src/contexts/global.tsx
```diff
@@ -116,18 +119,28 @@ export const GlobalContextProvider: FC<{ overrideContextValue?: PartialDeep<Glob
...
+      if (passphrase) {
+        wallet = walletOpen(password, walletEncrypted)
+        let mnemonicFinal = wallet.mnemonic
+        mnemonicFinal += ' ' + passphrase
+        wallet = getWalletFromMnemonic(mnemonicFinal)
+        setUsedPassphrase(true)
+      } else {
+        wallet = walletOpen(password, walletEncrypted)
+      }
       if (!wallet) return
```
nop33, line 144:
```suggestion
      let wallet = walletOpen(password, walletEncrypted)
      if (!wallet) return
      if (passphrase) {
        wallet = getWalletFromMnemonic(`${wallet.mnemonic} ${passphrase}`)
        setUsedPassphrase(true)
      }
```

src/pages/WalletManagement/WalletWordsPage.tsx
```diff
@@ -16,9 +16,10 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { walletGenerate } from '@alephium/sdk'
```
nop33, line 19: Not really in favor of not using our sdk function and duplicating its code in here.

LeeAlephium: Ah yes; this is an artifact of an earlier revision

src/contexts/addresses.tsx
LeeAlephium: > Not sure we would want that though

Yeah; you know the implications.

I'll work on a solution for the labeling

LeeAlephium: This is a comment on what's been said so far and my thoughts:

> in my decoy wallet I had set the label Vault [...] 

So yes, this is the correct behavior, for labels not to be saved, since it would void plausible deniability...

This is related again to needing to store information about password derived wallets... As soon as we store anything, it opens the doors to revealing other wallets exist.

After some time has passed since the initial implementation, I'm...willing to accept to void plausible deniability for a friendlier UX, which would include the "this passphrase has no associated wallet" message. What I am happy about is at its core, we still follow BIP39, it's just augmented with some things which void some of its benefits... /cc @polto @mvaivre if you're ok with this.

> I have 2 main addresses.

This is a bug / unexpected though... There should only be the main address associated with the non passphrase derived wallet

mvaivre: Yeah, labels (and any other metadata) would void plausible deniability more clearly than the message we've debated a lot about ("this passphrase has no associated wallet") 🙃

Encrypting the metadata file would solve the issue though. Thinking out loud : we need to keep in mind that one day we may want to have a service to share metadata between multiple wallets, so we shouldn't encrypt the file using the device specific credentials (password, pin or biometrics). We could maybe use the X first words of the mnemonic?

User authenticates => wallet is decrypted => first X words of mnemonic are used to decrypt the metadata file => wallet is operational.

WDYT?

nop33: The argument for losing the plausible deniability feature was not the message "this passphrase has no associated wallet", but the fact that we are storing an encrypted entry in localStorage. @polto and @LeeAlephium pointed out that this would cause the advanced attacker to think that this is the passphrase, meaning they would then torture the victim to reveal it, even if they didn't set one. I raised the issue that _"Does this mean we can never have anything encrypted stored in localStorage"_ to which no one really properly replied.

mvaivre: Mhh ok, this is not what I understood, but I was not involved deeply in the discussion. I'm hence probably wrong. For me, here's the typical scenario where you want to deny that you have access to a big wallet : 

1) The attacker wants you to open your wallet(s) or you die.
2) You show the attacker the decoy wallet, with some coins in it. 
3) He/she says: but heyyy, I'm clever and I see that you have another encrypted file there! You used a passphrase! Open it!
4) You say, "oh yeah, forgot about this, I'm sorry... I experimented with it in the past, but I don't think I have anything on it!". You type any passphrase, it opens an empty wallet. "You see?". 
5) "Well okay then." he / she says. "Have a nice day!". ☀️  ... OR they kill you anyway, because they understand the tech deeply and they don't trust you. But this is not something we can fix.

This works only if we don't show the "this passphrase has no associated wallet" message.

Anyway : if we store non-encrypted metadata, the scenario above is compromised as soon as the attacker looks at the localstorage (if possible?).

LeeAlephium: There's no way to access devtools in the production build (I think they may even be entirely absent), so checking it that way is not possible. The only way is they bring their own debugging tools with them.

I think it's a good idea to continue to leave out the message.

> "Does this mean we can never have anything encrypted stored in localStorage" to which no one really properly replied.

No; we can have things stored encrypted, but a (sophisticated, highly focused) attacker will demand all data to be decrypted. A passphrase derived wallet leaves no trace behind, so there is still no way for the attacker to know it exists. When storing _anything related to a derived wallet_ though, that's what voids plausible deniability... Even 1 bit of information voids it... 

I'm being lenient on allowing labels since I'm in agreement with @nop33 that the feature should still be accessible to a large amount of users. 

I have come to accept that the desktop-wallet needs to remain friendly and not expose advanced features. There can be another wallet by someone else which offers it all. 

My comment is now leading into a tangent but we should re-evaluate the features the wallet offers based on this view of the wallet (things like sweeping are advanced IMO).

mvaivre: I believe that we can find a balance, like we have done until now. IMO we successfully added convenient features for everyday users (labels etc.), while hiding advanced features efficiently (but still offering them). We can do better, but it's a nice start.

Let's not forget that we're at the beginning of the network, and we **need** to onboard advanced users to help us grow the network (devs, miners, Defi users etc.). 

So it's not a black or white / easy vs. complex situation. I'm sure we can find UI tricks to hide the complexity away.

I'll do a big redesign pass later on, as I also need to take care of Cheng's additions. I'll find ways to make this simple again.

polto: if you start with a fresh wallet and add a 12-24 BIP39 words wallet, you will have wallet data including the seed encrypted with wallets's password. Labels are probably stored encrypted with that same password.

if you add a passphrase we could have an "amnesic" mode. Or store the seed without the passphrase (potentially with its doublehash) and account labels encrypted with password+passphrase. 

We can have several slots for passphrase enabled wallets and we can by default create 10 with random encrypted data. Users can overwrite slots. This way there is no possibility to tell if it's the 10 random ones created by default or if the user has a password + passphrase valid for that slot. 

This way we preserve plausible deniability and have nice features such as labels.

But first we can start with amnesic wallet (in terms of labels and local data storage) because it's more simple.

LeeAlephium: Might be easier to actually go ahead with the 10 random encrypted data... The wallet makes no special distinction between passphrase or non-passphrase derived wallets...

How is it known what's a considered as a wallet which can be replaced / is a "dummy" wallet? I think rather than slots, the wallets just exist in storage but arent listed in the wallet listing...

LeeAlephium: I've put the PR back into WIP to do this later :)

LeeAlephium: This is looking really damn good guys. I'm ultra happy we stuck with BIP39 and then augmented it.

I think there's no point in an arbitrary number of random wallets. I think it's better to stick to fundamentals and ask ourselves what it is that actually gives us plausible deniability here.

* 0 wallets: The user has no ALPH.
* 1 wallets: The user has ALPH.
* 1+n wallets: The user has ALPH but no way to know how much.

Based on this, I suggest we simply make sure there is always a single extra dummy wallet in storage. This means the user can even have 2 wallets and 1 dummy and it's impossible to know the user has more than 1.

Ready to be reviewed soon as of this comment...

LeeAlephium:
In order for the address metadata decoys to work, we had to encrypt the information. We use data inside the encrypted wallet (mnemonic) to encrypt this data. This means user data needs to be migrate in older wallets before this version. Luckily this is easy to detect and I begin to version our data for future data migrations, making it _very_ easy.

After this, I implement the "imp" which will randomly insert address metadata. It will roll a 13 sided die and if it lands on 1, it will create new address metadata, retaining plausible deniability.

This randomness insertion could be expanded to wallets but with BIP39 it would just be cluttering storage IMO. I don't see any benefit. We need to only do this for address metadata because we want to be able to store additional data about ephemeral  passphrase wallets. I think we've done something quite novel here gang :).

src/utils/addresses.ts
```diff
@@ -63,7 +71,13 @@ export const storeAddressMetadataOfAccount = (accountName: string, index: number
     Object.assign(existingAddressMetadata, settings)
   }
   console.log(`🟠 Storing address index ${index} metadata locally`)
-  localStorage.setItem(constructMetadataKey(accountName), JSON.stringify(addressesMetadata))
+  localStorage.setItem(
+    constructMetadataKey(accountName),
+    JSON.stringify({
+      version: latestUserDataVersion,
+      encryptedSettings: encrypt(mnemonic, JSON.stringify(addressesMetadata))
```
nop33, line 78: Why are we using the mnemonic to encrypt the address metadata and not the password?

src/utils/migration.ts
```diff
@@ -0,0 +1,55 @@
...
+      localStorage.setItem(
+        key,
+        JSON.stringify({
+          version: '2022-05-27T12:00:00Z',
+          encryptedSettings: encrypt(mnemonic, JSON.stringify(addressSettingsList))
+        })
+      )
+    }
+  }
+}
```
nop33, line 55: The migration function does not work properly. I had 10 accounts set up, and after logging in to one of them (called `Ilias Testnet`) I noticed that a new entry in localStorage was created with the key:
```
addresses-metadata-b2189fac31a86e96d637b492378fbda08f51bfebfa9e3b2f95f5d36246b75d85c2a9eea9e545c2bfd85dbf4d5b7deae4986dad6e18109c8842790d8bd695bc87
```
and value
```
{"version":"2022-05-27T12:00:00Z","encryptedSettings":"{\"salt\":\"87916ec83063b759891903b28b538403f3c42a633067302bd972c986010aef25346920675f38b4f23432da14c0c2c95eec50ea3c4342fc96a0d350781a419834\",\"iv\":\"9979900b2bc7327f3d9f623d783aa04a9ec30126b6c909f5372f169119645a191c3c9988ba6f7436eed1a9a0840dae10a080c981bc9832e099a765406fbc04c1\",\"encrypted\":\"956a0f2a62f03021f21d482b825f463d6652eb4b3ca30bece74e9b0645e3148ec7076aa62409303e3148f6\",\"version\":1}"}
```
but the entry with key
```
addresses-metadata-Ilias Testnet
```
still exists, but it's value has changed to:
```
{"version":"2022-05-27T12:00:00Z","encryptedSettings":"{\"salt\":\"b1d127f6ec4bc237cea3d3620c4495e01c4d84d47498cf0f026c6bd6e8c751b0fdd2444f5c54c187fe13403be6cfe51d047b158a66a5007606029586c9b55e0c\",\"iv\":\"0a0c88b067a00dbf60fec7c7bdd5bbd1060443011afecd56a4fe2bd56ea9496f5fbee834b4ed2924b0bc14174fb0c8a1f629af0db70106f4f6eca12e4f77a9c8\",\"encrypted\":\"318f9e0cf4326b5a0d173b121d6e8ddab828e4ed3f12ab2933d70ec9240713499a48d8b0393590dca88320\",\"version\":1}"}
```

Secondly, I lost all my metadata. No matter which account I login, there's no metadata there (no address labels, no additional addresses, nothing).

Third, I realize that all metadata of all wallets are encrypted with the mnemonic of the account I logged in with, not just the metadata of the account I logged in with. That doesn't make sense.

src/contexts/addresses.tsx
nop33: I finally got time to go through all discussions. Thank you @polto and @lf94 for the ideas.

I want to propose a solution that is somewhere in the middle between @polto's and @LeeAlephium's.

@polto's [idea](https://github.com/alephium/desktop-wallet/pull/244#discussion_r882482267) cannot really give us plausible deniability, because the system cannot know which of the "slots" contain random data and are available to be overridden and which actually contain useful metadata.

@LeeAlephium's [idea](https://github.com/alephium/desktop-wallet/pull/244#discussion_r884039650) of a dice roll creates an indefinite number of metadata in the localStorage and depends on how often the user uses the wallet. I don't like that, even if it provides 100% plausible deniability, but can create a lot of garbage along the way, which we will have to carry along.

My proposal for storing address metadata without sacrificing plausible deniability is:

## 1. address metadata storage key
Each wallet stores its metadata. The key is `'address-metadata-' + doublehash('my wallet name' + passphrase)` (same as @LeeAlephium's [proposal](https://github.com/alephium/desktop-wallet/pull/244#discussion_r881980540)). So, for a wallet called _"Ilias' badass wallet"_ with no passphrase, and for the same wallet, but with a passphrase, the keys of the 2 address metadata entries could look like this:
```
addresses-metadata-00f96c9b1db7a585d4ba3d2a4af32ca3f95d20ef5766e1773cf0c07e12672b6516e6f21c679f40c95cefdf41c1ce37724c6f3781caaabf0b481391a4144978ee
addresses-metadata-b2189fac31a86e96d637b492378fbda08f51bfebfa9e3b2f95f5d36246b75d85c2a9eea9e545c2bfd85dbf4d5b7deae4986dad6e18109c8842790d8bd695bc87
```

## 2. address metadata storage value
The value of the address metadata needs to be stored encrypted.

@polto proposed we use `password + passphrase` as the encryption key. This approach would work, but it would make the future transfer of metadata across devices problematic (@mvaivre's [request](https://github.com/alephium/desktop-wallet/pull/244#discussion_r881290829)).

@mvaivre proposed we use the `X` first words of the 24-words mnemonic. @LeeAlephium proposed to use the whole 24-words mnemonic. Both of these approaches suffer the same problem: The attacher knows the password to the 24-word mnemonic wallet, which means they know the 24-words, which means they can try and decrypt address-metadata, and if they find one where the decryption is successful, it would reveal that the user is using a passphrase.

To solve both problems mentioned above, I propose using the 24-words mnemonic + passphrase as the encryption key for the address metadata storage value.

## 3. Maintaining (most of) plausible deniability
Instead of 10 slots with random metadata which the wallet cannot know which it can override (@polto's idea), and instead of probabilistically generating metadata every time the user logs in (@LeeAlephium's idea), I propose that the wallet picks a random number `Y` between 5 and 20 (upper bound TBD), and it generates ONLY ONCE `Y` address metadata entries with garbage. So, a user with 3 wallets (_"My wallet"_, _"My mom's wallet"_, _"My sister's wallet"_) and 1 additional passphrase wallet will have 4 + `Y` entries. Plausible deniability is void only when that number exceeds the upper bound.

Each time a wallet is unlocked using a random passphrase, a new address metadata entry is created. To avoid cluttering the storage, we can avoid storing address metadata when there is only 1 address and when this address has no metadata associated with it (the default state of a new wallet).

src/utils/migration.ts
LeeAlephium: >  looks like it wasn't tested

It was tested but clearly not enough; I created wallets before the change, tested after the change, and I was able to see the old metadata...

> still exists

The old key shouldn't exist anymore; bug

> Each account metadata should be encrypted with its own mnemonic, no?

Yeah, nice catch, and is most likely why you lost the metadata originally... Rather than iterate over all address metadata properties I should only migrate based on the account which is selected

---

Thank you for testing; features like this need to really be scrutinized so I really appreciate it 🙂

src/contexts/addresses.tsx
LeeAlephium: > We hadn't decided which approach to go for, we're still discussing.

It was more about it's problematic than reaching a conclusion

> Cons: not 100% plausible deniability

It's either 0% or 100%, there's no inbetween - I think a better description is "maximum amount of wallets until plausible deniability is voided"

Another problem with proposition 3 is their **entire** wallet transitions from 100% plausibly deniable to 0% plausibly deniable. That's a big transition...

> On the contrary. People where Y is close to the low bound have a lot of space for creating wallets. People where Y is close to the upper bound and have many wallets will have their plausible deniability voided (example: Y is 18 and the user has 2 wallets and 1 passphrase wallet, the total number of address metadata entries is 21, which is above the upper bound of the random range for Y, which is 20 on the above proposal).

Maybe you misunderstood my comment; but what's being written isn't adding up:

* Generate a number 5 >= Y <= 20
* Generate Y wallets
* Let there be X user generated wallets
* Plausible deniability voided if X > Y

If Y is 5 it will be voided much sooner than if it were 20?...

LeeAlephium: @nop33 clarified offline: It should be if X + Y > UpperRange, then it's voided

So it's UpperRange - LowerRange which determines how many wallets the user can create until it's voided.

src/contexts/global.tsx
```diff
@@ -116,19 +122,33 @@ export const GlobalContextProvider: FC<{ overrideContextValue?: PartialDeep<Glob
...
     if (!walletEncrypted) {
       setSnackbarMessage({ text: 'Unknown wallet name', type: 'alert' })
       return
     }
     try {
-      const wallet = walletOpen(password, walletEncrypted)
+      let wallet = walletOpen(password, walletEncrypted)
       if (!wallet) return
+      if (passphrase) {
+        wallet = getWalletFromMnemonic(`${wallet.mnemonic} ${passphrase}`)
```
polarker, line 140: This is not how passphrase is used in BIP39. We'd better test against those test vectors.

bip39 supports passphrase already, but it's named as `password`.

src/utils/misc.ts
```diff
@@ -38,3 +40,15 @@ export const openInWebBrowser = (url: string) => {
     }
   }
 }
+
+export const stringToDoubleSHA215HexString = (data: string): string => {
```
polarker, line 44: I would suggest testing this with some test vectors. Crypto functions are hard to use sometimes...

src/utils/migration.ts
```diff
@@ -0,0 +1,55 @@
...
+export const latestUserDataVersion = '2022-05-27T12:00:00Z'
+
+//
+// Arguments are essentially dependencies of the migrations.
+// Unfortunately mnemonics are a dependency due to being needed to encrypt address
+// metadata, so migration must absolutely occur after login at the earliest.
+//
+// Future dependencies must be explained as they are added.
+//
+export const migrateUserData = (mnemonic: string) => {
```
polarker, line 30: Do we want to put this in `wallet-sdk`?

Also better to have unit tests cover this function. It can be altered accidentally.

polarker (commented):
I went through it quickly. I think we need to have some practice regarding testing critical path code. cc @mvaivre

src/utils/misc.ts
LeeAlephium: Yeah, we should have some tests for sure...!

src/utils/migration.ts
LeeAlephium: Yeah having tests would've caught the error @nop33 hit right away too. I'll do this next week along with the other tests :ok_hand:

src/contexts/global.tsx
polarker: > To create a binary seed from the mnemonic, we use the PBKDF2 function with a mnemonic sentence (in UTF-8 NFKD) used as the password and the string "mnemonic" + passphrase (again in UTF-8 NFKD) used as the salt.

`passphrase` should be salt, instead of being the actual 13th/25th world

LeeAlephium: I'm going to find some implementations to verify, but I'm pretty sure then all it means is I should remove the space between the mnemonic and the passphrase, example:

* Current: "hello world some word extraword"
* Corrected: "hello wold some wordextraword"

salts are appended to hash inputs i.e. `h(input + salt)` so this should be right... Going to verify though!

src/utils/migration.ts
LeeAlephium: > Do we want to put this in wallet-sdk?

We can move it there at a later time

src/contexts/global.tsx
LeeAlephium: This should change probably be put into `alephium/sdk` instead - I'll go do it there and change the function call here

LeeAlephium: Fixed in alephium/sdk and applied here

LeeAlephium:
This is just waiting on js-sdk release for BIP39 support, and then I'll need to rebase and then I'll request a re-review

LeeAlephium:
The "should lock the wallet when idle for too long after successful login" test fails now for some reason but I find it difficult to figure out why via the test output. Will look further into it tomorrow :mag:

mvaivre:
After discussion (hopefully a last time) with polto, I suggest to change the wording to ensure the plausible deniability. I'll commit my changes directly and link a screenshot.

The main idea is to replace the part `An incorrect passphrase could generate a wallet with zero funds. Do not panic, this is normal! Calmly try again.` by a warning saying that this is an advanced feature + a link to a wiki article explaining how it works.

LeeAlephium:
> I suppose there's nothing I can do to move this PR. Otherwise, let me know!

Giving it another test would be the next step :pick:

mvaivre:
And after taking a step back, discussing again with @nop33 and polto, I think we should start simple, focusing on what's visible (phrasing) to allow for plausible deniability - at least for cases where the attacker is not a super tech saavvy hacker.

I think it would be better to start with @nop33  the "upper bound" approach, which avoids bloating the storage while keeping the code simple. If we take 20 as an upper bounds, it'd also provide 100% deniability for a very large portion of our users. 

In summary, I'd rather start with a simpler approach for now - focusing on plausible deniability through UI - and maybe later iterate and maybe make it more complex if needed. Let's not forget that we'll eventually support Ledger, which will probably be the preferred way of securing the wallet.

LeeAlephium:
* Both approaches are same in complexity
* @nop33 's approach is not UI bound (the actual data is protected as well to a degree)
* The iteration has happened since the beginning of the PR
* Ledger integration will take years

This leads me to ask (but not argue):

* Why will we use the upper bound approach when it has the same complexity and UI consequences as the unbounded approach?
* Why are we ignoring the iteration we've already done?
* Given the immense time requirement to develop a hardware wallet integration why is it relevant to the discussion?

I'll revise the PR to use the upper bound approach

mvaivre:
> Why wasn't I involved in the discussion?

Let's start with this one! Ilias, Polto and me are meeting in Neuchatel IRL, and it's so much easier to find solution / compromise that suits everyone. So we quickly got around the question again, and agreed to split this into multiple iterations: 

**First iteration (now)**

1) We start simple (KISS). When using a "decoy" wallet, we store the metadata the same way as we're doing today. 
2) When a wallet is using a passphrase, we don't store *any* metadata linked to this wallet. We trade a bit of convenience for 100% plausible deniability while keeping our code very simple. 
3) For passphrase wallets, the only thing we will need to implement is a flag (stored in memory, like `passphrase: true`) set at the unlocking step, in order to hide UI elements related to labels / colors and prevent saving a new metadata entry (or overriding an existing one) in the local storage.
4) Addresses will need to be derived every time a "passphrase wallet" is opened.

**Second iteration**

Address discovery will be available through our backend (https://github.com/alephium/desktop-wallet/issues/136), removing the need for step 4, improving UX.

**Third iteration**

/!\ This part may change in the future, no need to spend more time discussing it today :)

- We may allow users to import read-only wallets using an extended public key. By doing this, nobody can prove that local storage entry is actually linked to a "writable" wallet. 
- We may also store everything linked to metadata and wallets in 1 encrypted binary file respectively (instead of multiple entries) making it less obvious the number of wallets available. 
- ... we may have more ideas in the future ;)
- This would allow us to support metadata for Ledger wallets as well, without compromising on plausible deniability.

LeeAlephium:
> Polto and me are meeting in Neuchatel IRL
Nice :sunglasses: 

From my understanding the goal has changed again from "using upper bound method" to "using BIP39 without metadata" - is this right?...

> We may also store everything linked to metadata and wallets in 1 encrypted binary file respectively (instead of multiple entries) making it less obvious the number of wallets available.

There are problems with this (and the fact it won't stop anything) but yeah, the 3rd section I'll leave for another day

nop33:
> From my understanding the goal has changed again from "using upper bound method" to "using BIP39 without metadata" - is this right?

Yes. I'll start working on the wiki page that we will link to from the login screen.

> There are problems with this (and the fact it won't stop anything) but yeah, the 3rd section I'll leave for another day

The key here is that the moment we allow extended public key import, we can then use that as the key to store metadata (instead of the wallet name). This allows us to store metadata not only for decoy wallets, but for any wallet, because there's no way to prove that metadata found in localStorage are associated with a passphrase wallet. Let's leave this discussion for another day, however!

nop33:
@LeeAlephium are you working on the failing tests?

LeeAlephium:
Hm, I guess it's not pulling a later sdk version. Will be easy to fix quickly. Might have appeared because of the rebasing I had to do

Sakrecoer:
i appreciate the heads-up and salut the effort! It's a small change for software, but a giant leap in user-understanding! 🧑‍🚀

LeeAlephium:
>  I'm working on it together with disabling labels, colors, main address

I completely forgot to do this :smiling_face_with_tear: I don't mind doing it!

nop33 (approved):
Guys, I think we are ready to merge this! All manual tests on my side look good. Unit tests also look good.

PR #255, opened 2022-06-07
Rename references of "account" to "wallet"

### LINKED ISSUE #198: Clear up wallet / account terminology confusion
As discussed with @nop33 on Slack, the use of wallet vs. account is not consistent right now.
We should have : 1 seed = 1 wallet.

Let's replace "account" with "wallet".

nop33:
Closes #198


LeeAlephium:
Did a quick `git grep account` and I gotta say you really nuked "account"! :)

Issue #257, opened 2022-06-08
Wallet crashes after opening it more than 60 times in an hour


LeeAlephium:
The wallet will crash if you open it more than 60 times in an hour 🙂

```
{"message":"API rate limit exceeded for 76.67.201.44. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting%22%7D
```

GitHub limits to 60 requests per hour for unauthenticated requests

Request in question: https://api.github.com/repos/alephium/desktop-wallet/releases/latest

You don't even have to open the wallet 60 times: visit that link more than 60 times and then open the wallet. Crash :D

```
for i in $(seq 61); do curl https://api.github.com/repos/alephium/desktop-wallet/releases/latest > /dev/null; sleep 1; done
```

Then open the wallet, should crash

(Until the next hour, where the limit  will be reset :p )

LeeAlephium:
Yeah, that's the quick fix for sure

PR #260, opened 2022-06-17
Check version once each hour

LeeAlephium:
Resolves https://github.com/alephium/desktop-wallet/issues/257

* On startup, if an hour has passed, check the version, and schedule a timeout for another hour
* If an hour has not passed, schedule a timeout for the time remaining until the next hour
* The version message will only appear once during runtime if there's the same version mismatch (rather than remind the user each hour)
* Thus the version message will appear for _each_ unique version change during the course of the application's lifetime

A new "app data" was created to track this new property, and any future "application lifetime" properties. It's possible there are other properties like this that exist already that should live there but that's out the scope of the issue

This is easier to test if you set `ONE_HOUR` to something like 20 seconds

### DIFF
--- a/src/utils/app-data.ts
+++ b/src/utils/app-data.ts
@@ -1,0 +1,35 @@
+/*
+Copyright 2018 - 2022 The Alephium Authors
+This file is part of the alephium project.
+
+The library is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+export const KEY_APPDATA = 'alephium/desktop-wallet/appdata'
+
+export interface AppData {
+  lastVersionCheckedAt: Date
+}
+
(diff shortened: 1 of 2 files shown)


nop33 (commented):
I checked out the branch, changed the `"version"` property of the package.json to `"1.1.0"` ( so that I force to get an update message), run the app, but I did not get the update message. I checked the localStorage and I did not find an entry with the name `alephium/desktop-wallet/appdata`.

LeeAlephium:
I did consider the first run for what it's worth, but I guess while focusing on other aspects of the problem the first run stopped working; it's all good now (tested with a fresh electron-pack binary even to increase confidence)

src/utils/app-data.ts
```diff
@@ -0,0 +1,35 @@
...
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+export const KEY_APPDATA = 'alephium/desktop-wallet/appdata'
```
mvaivre, line 19: Nit: not a fan of the `AppData` name which is a bit generic, and could be confused with the data used to run the app.
`AppMetaData` may be a bit better?

mvaivre:
Mhhh, I had the message while testing the passphrase PR (v1.2.1) 

image

LeeAlephium:
> Mhhh, I had the message while testing the passphrase PR (v1.2.1).

I'm not sure what to make of this

nop33:
I reported the bug: https://github.com/alephium/desktop-wallet/issues/273

Since it is not related to this PR, I recommend we proceed and merge this PR. @mvaivre any objections?

mvaivre:
@nop33 Nope, sounds good to me.

Issue #253, opened 2022-06-01
Network switch always displays "custom"


nop33:
https://user-images.githubusercontent.com/1579899/171364377-3e8113c8-4a15-4c57-9ab6-e572bbaa261f.mp4

LeeAlephium:
I think this issue is causing a secondary issue of no addresses showing up on network change... I'll bisect on Monday to see when it was introduced and what we can do

PR #263, opened 2022-06-20
Get the correct network name when switching networks

LeeAlephium:
Resolves https://github.com/alephium/desktop-wallet/issues/253

This was an interesting issue! Initially I was going to use `git bisect` to find where the bug was introduced, but as I was trying to find a good commit, it was evident that there was a data problem at hand. No matter how far back I went, the issue was still present.

After some deeper inspection the cultprit was found: "networkId: 4"

But how come this causes the network name to be custom? It's because of how network settings are compared to the settings in localStorage. 

```js
export const getNetworkName = (settings: Settings['network']) => {
  return (Object.entries(networkEndpoints).find(([, presetSettings]) => {
    return isEqual(presetSettings, settings)
  })?.[0] || 'custom') as NetworkName | 'custom'
}
```

If the objects are equal (by value / deep comparison), then it returns the key as the name.

Since "networkId: 4" is wedged in that comparison it'll always return "custom".

When was `networkId` introduced? Or should we ask the opposite question: when was it removed? The answer seems to be 10a5631909b3bf42d48464638e57013fb1e11ed3.

I don't think though this is the source of the problem because it seems so far - it's pre 1.1.0 release. I'm unable though to find any other mention of `networkId`. `git log -S networkId` was used to find that commit and nothing later than that moment mentions it. I quickly checked `js-sdk` and nothing seems relevant there either. Any theories? Did @nop33 and I just so happen to checkout an older version, run it, and forget about it?

This presents a larger problem at hand here: we can't presume the data coming from localStorage is valid or structured as expected.

Creating a Network class with an isEqual method or a function isEqualNetwork (I opted for this) which compares only its properties should adequately fix the problem.

### DIFF
--- a/src/utils/settings.ts
+++ b/src/utils/settings.ts
@@ -16,7 +16,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { clone, isEqual, merge } from 'lodash'
+import { clone, merge } from 'lodash'
 
 import { ThemeType } from '../style/themes'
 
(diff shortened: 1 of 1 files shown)


nop33 (approved):

mvaivre (approved):

LeeAlephium:
"it [networkId] was introduced when Cheng was working on the wallet connect feature in combination with the docker devnet setup, I am quite sure about that. I remember fixing a bug related to that. " -- nop33

So `git log -S` didn't find it because it was actually on cheng-walletconnect branch. That explains how it happened to us both :slightly_smiling_face:

Issue #259, opened 2022-06-16
Error while building the transaction: Invalid value for: body (missing keys in dictionary: attoAlphAmount at index 196: decoding failure)


LeeAlephium:

LeeAlephium:
Resolved by https://github.com/alephium/js-sdk/pull/112 when updating the version in package.json

LeeAlephium:
Updating the node and explorer apis in js-sdk is all that's needed to correct this.

nop33:
Is this something that is still happening?

LeeAlephium:
Once we merge the api updates, no

nop33:
Not relevant anymore. Closing.

PR #264, opened 2022-06-21
Update js-sdk to correct API usage error

LeeAlephium:
Resolves https://github.com/alephium/desktop-wallet/issues/259

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "electron-is-dev": "^2.0.0"
   },
   "devDependencies": {
-    "@alephium/sdk": "0.0.5",
+    "@alephium/sdk": "0.0.11",
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^12.1.0",
     "@types/jest": "^27.0.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "electron-is-dev": "^2.0.0"
       },
       "devDependencies": {
-        "@alephium/sdk": "0.0.5",
+        "@alephium/sdk": "^0.0.11",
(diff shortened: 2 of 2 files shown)


LeeAlephium:
Very weird "dmg-license not accessible from dmg-builder" error has appeared and I've been trying to resolve it. Just recording that I have been poking at it...

LeeAlephium:
We need to be very careful to not destroy the package-lock... Seems electron-builder has updated things and it will break building for now.

I can focus on updating our electron-builder version in another PR (maybe update all our dependencies)?

nop33:
That means that the currently installed wallets (1.2.0) will break once the hardfork (node 1.4.0) is released to mainnet, just like it did with our previous release (1.3.0), correct?

LeeAlephium:
Yep, so it'll be a necessary update

nop33:
> The old wallet is easily detectable via no Accept header, and newer wallets are by seeing Accept header and replying from the appropriate server instance.

This would require changes on our server to read the `Accept` header and accordingly route the request, correct?

package.json
```diff
@@ -38,7 +38,7 @@
     "electron-is-dev": "^2.0.0"
   },
   "devDependencies": {
-    "@alephium/sdk": "0.0.5",
+    "@alephium/sdk": "0.0.11",
```
nop33, line 41: We probably want `0.0.13-rc.0` here.

LeeAlephium:
This PR is in a weird state because it refers to an older js-sdk version with the fix but using a newer one actually doesn't include the fix. For sanity I'm going to close it and open a new one when js-sdk has the updated node and explorer APIs. I would also like to give it a better branch name like "1.4.0-compat".

PR #265, opened 2022-06-22
Safer code when group === 0

nop33:
It didn't cause any problems so far, just some extra calculations which can be avoided. I was implementing this in the mobile wallet and I thought it's better to reflect this code in the desktop wallet as well.


LeeAlephium:
LGTM :+1: Nice catch. `??` is a god send...!

LeeAlephium (approved):

PR #278, opened 2022-07-07
Remove metadata feature for passphrase-enabled wallets

nop33:
~~The @alephium/sdk version needs to be updated as part of this PR to remove the breaking changes (see https://github.com/alephium/js-sdk/pull/114)~~ Done.

Here's a description (with audio 🔊) of what was disabled and changed:

https://user-images.githubusercontent.com/1579899/177808127-82662359-5ae1-4c8f-818c-d3b25ee13927.mp4

### DIFF
--- a/src/components/Inputs/WalletPassphrase.tsx
+++ b/src/components/Inputs/WalletPassphrase.tsx
@@ -33,7 +33,7 @@ const passphraseLink = 'https://wiki.alephium.org/wallet/Desktop-Wallet-Guide#pa
 
 const WalletPassphrase = ({ value, onChange }: Props) => (
   <ExpandableSection sectionTitleClosed="Optional passphrase (advanced)" centered>
-    <InfoBox noMarginBottom importance="alert">
+    <InfoBox importance="alert">
       <p>
         <WarningEmphasis>This is an advanced feature! </WarningEmphasis>
         <br />
--- a/src/components/Inputs/WalletPassphrase.tsx
+++ b/src/components/Inputs/WalletPassphrase.tsx
@@ -16,6 +16,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
+import { FC } from 'react'
 import styled from 'styled-components'
 
 import { openInWebBrowser } from '../../utils/misc'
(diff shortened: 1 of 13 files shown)


src/modals/NewAddressModal.tsx
```diff
@@ -103,22 +103,30 @@ const NewAddressModal = ({ title, onClose, singleAddress }: NewAddressModalProps
...
+          {!singleAddress && (
+            <InfoBox Icon={Info} contrast noBorders>
+              The group number will be automatically be appended to the addresses’ label.
+            </InfoBox>
+          )}
+        </Section>
+      )}
+      {passphraseHash && singleAddress && (
+        <InfoBox contrast noBorders>
+          If you don&apos;t mind which group the address will be in, go ahead and hit that &quot;Generate&quot; button.
```
LeeAlephium, line 126: This makes me wonder as a user why I'd mind

LeeAlephium (approved):
Other than the one wording comment, the code looks good to me, I'll functionally test in the BIP39 PR

src/modals/NewAddressModal.tsx
nop33: Did you watch the video where I explain why I put this there? 😅  (audio on, plz!)

LeeAlephium: I watched it with no audio, yep :V _one moment later_ Ok so the message still itself doesn't explain why they'd mind, it could be written in a way that explains what will happen if generate is pressed instead making it more clear and not make them wonder what it really means: `"By default an address of any group will be generated. This is normal behavior. If you need to generate an address of a specific group please expand the section below. "` Here the user is assured that it's normal behavior and doesn't have to think "do I really mind?" - y'know?

nop33: OK, good suggestion, I'll rephrase, thanks a lot! <3

nop33: How'bout this? 
image

LeeAlephium: Yeah that's good

src/components/Inputs/WalletPassphrase.tsx
```diff
@@ -27,19 +28,20 @@ import Input from './Input'
...
         <WarningEmphasis>This is an advanced feature! </WarningEmphasis>
         <br />
-        <span>Use it only if you know what you are doing. Please, consult our </span>
-        <ActionLink onClick={() => openInWebBrowser(passphraseLink)}>documentation</ActionLink>
-        <span> to learn about it.</span>
+        <span>Use it only if you know what you are doing.</span>
+        <br />
+        <span>
+          Please, consult our <Link>documentation</Link> to learn about it.
+        </span>
```
nop33, line 44: I created the local `Link` component so that once we integrate i18n it will be easier to reason about. Lemme know what you think otherwise.

mvaivre (approved):

PR #281, opened 2022-07-08
Show upgrade msg only when version is newer

### LINKED ISSUE #273: Show update message only when GitHub release version is newer than current version
**Describe the bug**
As @mvaivre reported in https://github.com/alephium/desktop-wallet/pull/260#issuecomment-1170990495 and https://github.com/alephium/desktop-wallet/pull/260#issuecomment-1172147183, the code that checks for a **new** version on GitHub does not really check if the version it finds is newer, it only checks if it is different:

```js
currentVersion !== latestVersion
```

https://github.com/alephium/desktop-wallet/blob/master/src/hooks/useLatestGitHubRelease.ts#L38

**To Reproduce**
1. Change the `version` number in package.json to a higher one than the one from the latest release on GitHub
2. Run the app
3. You will see a message to upgrade to a previous version (which is the latest release on GitHub)

**Expected behavior**
The message should not appear if the latest release on GitHub is before the version of the app

nop33:
Closes #273


src/hooks/useLatestGitHubRelease.ts
```diff
@@ -38,7 +39,7 @@ const useLatestGitHubRelease = () => {
     const data = await response.json()
     const latestVersion = data.tag_name.replace('v', '')
 
-    if (semverRegex.test(latestVersion) && currentVersion !== latestVersion) {
+    if (semverRegex.test(latestVersion) && currentVersion && semverCompare(latestVersion, currentVersion) > 0) {
```
LeeAlephium, line 42: Nit: This check `semverRegex.test(latestVersion) ` should probably happen just before and if it fails, give a console.warn but it's not vital.

nop33: The regex is there to ensure that we don't check for release candidates. I don't see the reason to have a warning for that.

LeeAlephium: The warning would be "Does not follow standard semver syntax" in the cases we're run running a non-standard version (like you said) which makes it clear when developing (again it's just a Nit, I'd find this useful but maybe you not)

nop33: hm, why would you care whether there's a non-standard version available on GitHub while you're developing on your local wallet?

LeeAlephium: It's more independent of that: there may be a scenario when someone is running a dev version and has forgotten (maybe they self-built and it's been months) - otherwise I agree when developing you know X) I'm coming from a place of "If the semver string is not standard, show something" where I'm not assuming why, but the fact it can happen... Reporting it's not standard sounds useful to me because it becomes clear immediately this is not a standard release, *for whatever reason*

nop33: > Reporting that it's not standard sounds useful to me because it becomes clear immediately this is not a standard release

I don't really see any value as to why there should be a warning in the console when the latest release on GitHub is not one that the user should be concerned about. I'll bypass this nit, if that's ok, and merge :)

PR #282, opened 2022-07-11
Add missing translations

nop33:


src/components/AddressMetadataForm.tsx
```diff
@@ -51,7 +51,7 @@ const AddressMetadataForm = ({
         <>
           <HorizontalDivider narrow />
           <InlineLabelValueInput
-            label={t`★ Main address`}
+            label={`★ ${t`Main address`}`}
```
nop33, line 54: Since we already had the `Main address` keyword, I removed the `★ Main address` one.

src/components/Inputs/AmountInput.tsx
```diff
@@ -60,7 +60,7 @@ const AmountInput = ({ className, availableAmount, ...props }: AmountInputProps)
       amountValueAsFloat > parseFloat(availableAmountInAlph)
         ? t`Amount exceeds available balance`
         : amountValueAsFloat < parseFloat(minAmountInAlph)
-        ? t('Amount must be greater than {{minAmountInAlph}}', { minAmountInAlph })
+        ? t('Amount must be greater than {{ minAmountInAlph }}', { minAmountInAlph })
```
nop33, line 63: Sorry, I couldn't resist 😅

src/modals/AddressSweepModal.tsx
```diff
@@ -139,10 +139,14 @@ const AddressSweepModal = ({ sweepAddress, onClose, onSuccessfulSweep }: Address
...
-          {t`and transfer them to`}{' '}
-          <ColoredWord color={sweepAddresses.to.settings.color}>{sweepAddresses.to.getName()}</ColoredWord>.
+          <Trans t={t} i18nKey="sweepOperationFromTo">
+            This operation will sweep all funds from
+            <ColoredWord color={sweepAddresses.from.settings.color}>
+              {{ from: sweepAddresses.from.getName() }}
+            </ColoredWord>
+            and transfer them to
+            <ColoredWord color={sweepAddresses.to.settings.color}>{{ to: sweepAddresses.to.getName() }}</ColoredWord>.
+          </Trans>
```
nop33, line 149: I think this solution is more appropriate because there might be some languages where the sentences break at different points than in English, as the previous solution was implementing.

mvaivre: Nice one.

mvaivre (approved):

src/modals/AddressSweepModal.tsx
LeeAlephium: Very nice point :D

PR #283, opened 2022-07-11
Remove passphrase argument from data migration

nop33:
While removing a single arg seemed very trivial, I discovered some more things that had to be done:
- Removed unused passphrase state from `src/contexts/wallet.tsx`
- Removed hashed passphrase from global context and instead store a boolean flag: `isPassphraseUsed`
- Instead of passing the `isPassphraseUsed` arg to the `loadStoredAddressesMetadataOfWallet` and `storeAddressMetadataOfWallet`, just don't call them if passphrase is used
- Keep an optional `isPassphraseUsed` arg in the aforementioned functions for safety
- Rename `accountName` to `walletName` in migration file
- Clean-up component state after logging in

I have tested my changes with the following method:
1. Checked out `master`
2. Created a new wallet, added some metadata (generated a few addresses, added labels and colors)
3. Closed the app
4. Checked out this branch
5. Opened the same wallet to verify the data migration worked. I could still see the addresses/labels ✅ 
6. Opened the same wallet, but with a passphrase. I could not see any addresses/labels ✅ 
7. Repeated the process by swapping steps 5 and 6, just to check ✅ 

@LeeAlephium @mvaivre can you think of another test I could run? Please, feel free to test on your side as well, since this is a very delicate feature and we don't have space for error. Thanks!

### DIFF
--- a/src/modals/TransactionDetailsModal.tsx
+++ b/src/modals/TransactionDetailsModal.tsx
@@ -57,7 +57,7 @@ const TransactionDetailsModal = ({ transaction, address, onClose }: TransactionD
   const isOutgoingTx = amount < 0
   amount = isOutgoingTx ? amount * -1n : amount
   const addressName = address.getLabelName(!isPassphraseUsed)
-  const color = address.settings.color
+  const addressColor = address.settings.color
 
   const handleShowTxInExplorer = () => {
     openInWebBrowser(`${explorerUrl}/#/transactions/${transaction.hash}`)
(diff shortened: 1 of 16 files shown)


LeeAlephium:
LGTM, I did an additional test of having multiple non-passphrase wallets and switching between them (just to add one more bit of coverage to our manual testing) :ok_hand:

src/modals/TransactionDetailsModal.tsx
```diff
@@ -50,12 +50,14 @@ const TransactionDetailsModal = ({ transaction, address, onClose }: TransactionD
...
     },
-    passphraseHash
+    isPassphraseUsed
   } = useGlobalContext()
   const theme = useTheme()
   let amount = calAmountDelta(transaction, address.hash)
   const isOutgoingTx = amount < 0
   amount = isOutgoingTx ? amount * -1n : amount
+  const addressName = address.getLabelName(!isPassphraseUsed)
+  const color = address.settings.color
```
mvaivre, line 60: Nit: `addressName` <=> `addressColor` ?

nop33: Fixed!

mvaivre (approved):

PR #286, opened 2022-07-11
Make the application completely keyboard accessible

LeeAlephium:
Making my way through a list I've made:

* [x] Wallet selection modal cannot be triggered
* [x] Create / import wallet cannot be highlighted
* [x] Wallet words cannot be selected
* [x] Settings tabs cannot be changed
* [x] Any toggle cannot be highlighted
* [x] Expanding sections cannot be highlighted
* [x] Wallets cannot be deleted
* [x] Menu cannot be navigated to
* [x] Link buttons are not highlighted (but can be triggered)
* [x] Transaction history rows cannot be highlighted
* [x] Addresses rows cannot be highlighted
* [x] More info links cannot be highlighted
* [x] Address color cannot be highlighted
* [x] Address color selector cannot be closed without closing the entire modal
* [x] Address selector cannot be highlighted
* [x] Clipboard in address details cannot be highlighted
* [x] Back button in address details cannot be highlighted
* [x] Label in address details cannot be highlighted
* [x] Number of transactions cannot be highlighted
* [x] Total ALPH balance cannot be highlighted
* [x] Nothing in the transaction side pull-out can be highlighted
* [x] Nothing in the Security Check can be highlighted
* [x] Back button in wallet creation can't be highlighted

### DIFF
--- a/src/components/Buttons/OpenInExplorerButton.tsx
+++ b/src/components/Buttons/OpenInExplorerButton.tsx
@@ -42,7 +42,17 @@ const OpenInExplorerButton = ({ address, className }: OpenInExplorerButtonProps)
     openInWebBrowser(`${explorerUrl}/#/addresses/${address}`)
   }
 
-  return <ExternalLink className={className} data-tip={t`Open in explorer`} size={15} onClick={handleShowInExplorer} />
+  return (
+    <ExternalLink
+      className={className}
+      data-tip={t`Open in explorer`}
+      size={15}
+      onClick={handleShowInExplorer}
+      onKeyPress={handleShowInExplorer}
+      role="link"
+      tabIndex={0}
+    />
+  )
 }
 
 export default styled(OpenInExplorerButton)`
--- a/src/components/ExpandableSection.tsx
+++ b/src/components/ExpandableSection.tsx
@@ -18,9 +18,10 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { motion } from 'framer-motion'
 import { ChevronDown } from 'lucide-react'
-import { FC, useEffect, useState } from 'react'
+import { FC, useCallback, useEffect, useState } from 'react'
 import styled, { css } from 'styled-components'
 
+import InputArea from './Inputs/InputArea'
 import { Section } from './PageComponents/PageContainers'
 
 interface ExpandableSectionProps {
@@ -46,35 +47,56 @@ const ExpandableSection: FC<ExpandableSectionProps> = ({
   className
 }) => {
   const [isExpanded, setIsExpanded] = useState(open)
+  const [isVisible, setIsVisible] = useState(open)
 
   useEffect(() => {
-    setIsExpanded(open)
+    if (open) {
+      setIsVisible(open)
+      setIsExpanded(open)
+    } else {
+      setIsExpanded(open)
+    }
   }, [open])
 
-  const handleTitleClick = () => {
+  const handleTitleExpansion = () => {
     const newState = !isExpanded
     onOpenChange && onOpenChange(newState)
+    if (newState) {
+      setIsVisible(newState)
+    }
     setIsExpanded(newState)
   }
 
+  const onAnimationComplete = useCallback(() => {
+    setIsVisible(isExpanded)
+  }, [isExpanded])
+
+  const titleText = isExpanded && sectionTitleOpen ? sectionTitleOpen : sectionTitleClosed
+
   return (
     <ExpandableSectionContainer className={className} shrinkWhenOpen={shrinkWhenOpen} isOpen={isExpanded}>
-      <Title onClick={handleTitleClick}>
+      <Title onInput={handleTitleExpansion} aria-expanded={isExpanded}>
         {centered && <LeftDivider />}
         {isCheckbox ? (
-          <input type="checkbox" checked={isExpanded} onChange={() => setIsExpanded(!isExpanded)} />
+          <input type="checkbox" tabIndex={-1} checked={isExpanded} onChange={() => setIsExpanded(!isExpanded)} />
         ) : (
           <Chevron animate={{ rotate: isExpanded ? 180 : 0 }} />
         )}
 
-        <TitleText>{isExpanded && sectionTitleOpen ? sectionTitleOpen : sectionTitleClosed}</TitleText>
+        <TitleText>{titleText}</TitleText>
         <Divider />
       </Title>
-      <ContentWrapper animate={{ height: isExpanded ? 'auto' : 0 }} transition={{ duration: 0.2 }}>
-        <Content>
-          <Section align="stretch">{children}</Section>
-        </Content>
-      </ContentWrapper>
+      {isVisible && (
+        <ContentWrapper
+          onAnimationComplete={onAnimationComplete}
+          animate={{ height: isExpanded ? 'auto' : 0 }}
+          transition={{ duration: 0.2 }}
+        >
+          <Content>
+            <Section align="stretch">{children}</Section>
+          </Content>
+        </ContentWrapper>
+      )}
     </ExpandableSectionContainer>
   )
 }
@@ -95,9 +117,8 @@ const ExpandableSectionContainer = styled.div<{ shrinkWhenOpen: boolean; isOpen?
     `}
 `
 
-const Title = styled.div`
+const Title = styled(InputArea)`
   display: flex;
-  cursor: pointer;
   align-items: center;
   color: ${({ theme }) => theme.global.accent};
 `
--- a/src/components/ExpandableSection.tsx
+++ b/src/components/ExpandableSection.tsx
@@ -18,7 +18,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { motion } from 'framer-motion'
 import { ChevronDown } from 'lucide-react'
-import { FC, useCallback, useEffect, useState } from 'react'
+import { FC, useEffect, useState } from 'react'
 import styled, { css } from 'styled-components'
 
 import InputArea from './Inputs/InputArea'
@@ -67,10 +67,6 @@ const ExpandableSection: FC<ExpandableSectionProps> = ({
     setIsExpanded(newState)
   }
 
-  const onAnimationComplete = useCallback(() => {
-    setIsVisible(isExpanded)
-  }, [isExpanded])
-
   const titleText = isExpanded && sectionTitleOpen ? sectionTitleOpen : sectionTitleClosed
 
   return (
@@ -88,7 +84,7 @@ const ExpandableSection: FC<ExpandableSectionProps> = ({
       </Title>
       {isVisible && (
         <ContentWrapper
-          onAnimationComplete={onAnimationComplete}
+          onAnimationComplete={() => setIsVisible(isExpanded)}
           animate={{ height: isExpanded ? 'auto' : 0 }}
           transition={{ duration: 0.2 }}
         >
--- a/src/components/Inputs/ColorPicker.tsx
+++ b/src/components/Inputs/ColorPicker.tsx
@@ -20,9 +20,11 @@ import { AnimatePresence, motion } from 'framer-motion'
 import { useState } from 'react'
 import { TwitterPicker } from 'react-color'
 import { useDetectClickOutside } from 'react-detect-click-outside'
+import { useTranslation } from 'react-i18next'
 import styled from 'styled-components'
 
 import { getRandomLabelColor, labelColorPalette } from '../../utils/colors'
+import InputArea from '../Inputs/InputArea'
 import { inputDefaultStyle, InputProps } from '.'
 
 interface ColorPickerProps {
@@ -31,21 +33,28 @@ interface ColorPickerProps {
 }
 
 const ColorPicker = ({ value, onChange }: ColorPickerProps) => {
+  const { t } = useTranslation('App')
   const color = value?.toString() || getRandomLabelColor()
   const [isPopupOpen, setIsPopupOpen] = useState(false)
   const ref = useDetectClickOutside({ onTriggered: () => setIsPopupOpen(false) })
 
+  const handlePopupOpen = () => setIsPopupOpen(!isPopupOpen)
+  const onChangeComplete = (newColor: { hex: string }) => {
+    onChange(newColor.hex)
+    handlePopupOpen()
+  }
+
   return (
     <ColorPickerContainer ref={ref}>
-      <Input onClick={() => setIsPopupOpen(!isPopupOpen)}>
+      <InputAreaStyled aria-label={t`Pick a color`} onInput={handlePopupOpen}>
         <Circle color={color} />
-      </Input>
+      </InputAreaStyled>
       <AnimatePresence exitBeforeEnter initial={true}>
         {isPopupOpen && (
           <Popup initial={{ opacity: 0 }} animate={{ opacity: 1 }} exit={{ opacity: 0 }}>
             <TwitterPickerStyled
               color={color}
-              onChangeComplete={(newColor) => onChange(newColor.hex)}
+              onChangeComplete={onChangeComplete}
               colors={labelColorPalette}
               triangle="top-right"
             />
@@ -65,9 +74,8 @@ const ColorPickerContainer = styled.div<InputProps>`
   flex-direction: column;
 `
 
-const Input = styled.div<InputProps>`
+const InputAreaStyled = styled(InputArea)<InputProps>`
   ${({ isValid }) => inputDefaultStyle(isValid)}
-  cursor: pointer;
   position: relative;
   display: inline-flex;
   align-items: center;
--- a/src/components/Inputs/Select.tsx
+++ b/src/components/Inputs/Select.tsx
@@ -16,7 +16,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { AnimatePresence, motion } from 'framer-motion'
+import { AnimatePresence } from 'framer-motion'
 import { isEqual } from 'lodash'
 import { MoreVertical } from 'lucide-react'
 import { useCallback, useEffect, useState } from 'react'
@@ -25,6 +25,7 @@ import styled, { css } from 'styled-components'
 import { sectionChildrenVariants } from '../PageComponents/PageContainers'
 import Popup from '../Popup'
 import { inputDefaultStyle, InputLabel, InputProps } from './'
+import InputArea from './InputArea'
 
 export interface SelectOption<T> {
   value: T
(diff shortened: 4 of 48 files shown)


LeeAlephium:
Tested with Orca screen reader in GNOME: a bunch of components need names but very cool to see it more or less working just fine :D

LeeAlephium:
Demonstration:

https://user-images.githubusercontent.com/105806603/178786905-8c40b987-df60-47ee-a1a3-c8008aeeb0a9.mp4

LeeAlephium:
Please note this is the *first iteration*. It's large so I wanted to get the groundwork done before re-iterating

LeeAlephium:
The second iteration will focus on making sure roles are properly applied and tab indexable elements have appropriate labels

LeeAlephium:
@nop33 could you try to fix the test? I've been trying for almost an hour now and no matter what I do, I can't seem to get the `select` to trigger... I've tried the below:

* userEvent.selectOptions
* fireEvent.change
* Used getByTestId to be more precise about element selected

Maybe I'm using them incorrectly but based on what I'm reading and other uses I'm seeing, it makes no sense why it's not working...

nop33:
@LeeAlephium with `DEBUG_PRINT_LIMIT=100000 npm test` I can see that the Login button is disabled, so clicking on it does not get you to the dashboard page, so the `Transaction history` text is not there, and that's why the tests are failing.

src/components/Buttons/OpenInExplorerButton.tsx
```diff
@@ -42,7 +42,19 @@ const OpenInExplorerButton = ({ address, className }: OpenInExplorerButtonProps)
     openInWebBrowser(`${explorerUrl}/#/addresses/${address}`)
   }
 
-  return <ExternalLink className={className} data-tip={t`Open in explorer`} size={15} onClick={handleShowInExplorer} />
+  return (
+    <>
```
nop33, line 46: Why the Fragment?

src/components/ExpandableSection.tsx
```diff
@@ -47,15 +47,21 @@ const ExpandableSection: FC<ExpandableSectionProps> = ({
...
 
   return (
     <div className={className}>
-      <Title onClick={handleTitleClick}>
+      <Title
+        onClick={handleTitleExpansion}
+        onKeyPress={handleTitleExpansion}
+        role="button"
+        tabIndex={0}
+        aria-expanded={isExpanded}
```
nop33, line 63: Nice job 👏

src/components/Inputs/Select.tsx
```diff
@@ -89,14 +106,18 @@ function Select<T>({
...
+        onClick={onContainerClick}
       >
         <InputLabel inputHasValue={!!value} htmlFor={id}>
           {label}
         </InputLabel>
-        <MoreIcon>
-          <MoreVertical />
-        </MoreIcon>
+        {options.length > 1 && (
+          <>
```
nop33, line 115: Why the Fragment?

src/components/Popup.tsx
```diff
@@ -16,40 +16,60 @@ You should have received a copy of the GNU Lesser General Public License
...
+
+  useEffect(() => {
+    document.addEventListener('keydown', handleEscapeKeyPress, false)
+    return () => {
+      document.removeEventListener('keydown', handleEscapeKeyPress, false)
+    }
+  }, [handleEscapeKeyPress, onBackgroundClick])
+
+  return (
+    <>
```
nop33, line 45: Why the Fragment?

src/components/TabBar.tsx
```diff
@@ -32,16 +32,22 @@ interface TabBarProps {
...
-    <TabBarContainer>
+    <TabBarContainer role="navigation">
       <TabSelector
         animate={{ x: `${tabItems.findIndex((t) => t.value === activeTab.value) * 100}%` }}
         transition={{ duration: 0.2 }}
         style={{ width: `${100 / tabItems.length}%` }}
       />
       <TabBarContent>
-        {tabItems.map((i) => (
+        {tabItems.map((i, index) => (
```
nop33, line 42: Why the `index`?

src/components/Truncate.tsx
```diff
@@ -20,7 +20,13 @@ import { FC } from 'react'
 import styled from 'styled-components'
 
 const Truncate: FC<{ className?: string }> = ({ className, children }) => {
-  return <div className={className}>{children}</div>
+  return (
+    <>
```
nop33, line 24: Why the Fragment?

src/modals/SettingsModal/WalletsSettingsSection.tsx
```diff
@@ -116,15 +116,25 @@ const WalletItem = ({ walletName, isCurrent, onWalletDelete }: WalletItemProps)
...
+      role="row"
+      tabIndex={0}
       onMouseEnter={() => setIsShowingDeleteButton(true)}
       onMouseLeave={() => setIsShowingDeleteButton(false)}
     >
-      <WalletName>
+      <WalletName role="cell" tabIndex={0} onFocus={() => setIsShowingDeleteButton(true)}>
         {walletName}
-        {isCurrent && <CurrentWalletLabel> {t`(current)`}</CurrentWalletLabel>}
+        {isCurrent && <CurrentWalletLabel>{t`(current)`}</CurrentWalletLabel>}
```
nop33, line 126: Why remove the space?

nop33 (commented):
Awesome job @LeeAlephium! It feels so good to press tab and navigate the UI, so smooth.

One remark: I would appreciate it if when you open a PR you would first go through the PR diff line by line and do a code review on your own code, before assigning it for someone else to do a review. Many small things can be spotted this way and the reviewer won't have to spend time pointing them out (like the unnecessary use of Fragments, or the removal of the space). Thanks!

nop33:
> I've been trying for almost an hour now and no matter what I do, I can't seem to get the select to trigger

@LeeAlephium I also tried to figure it out by debugging for an hour, but nothing worked so far...

LeeAlephium:
Our app is now accessible to the blind as well


https://user-images.githubusercontent.com/105806603/179052441-bf20eafd-d2c3-44a8-bd8b-46e36630463b.mp4

src/components/Inputs/ColorPicker.tsx
```diff
@@ -31,21 +32,34 @@ interface ColorPickerProps {
...
+
   return (
     <ColorPickerContainer ref={ref}>
-      <Input onClick={() => setIsPopupOpen(!isPopupOpen)}>
+      <Input
+        role="button"
+        tabIndex={0}
+        aria-label={t`Pick a color`}
+        onClick={handlePopupOpen}
+        onKeyPress={handlePopupOpen}
```
nop33, line 53: Maybe we could have a component like `ClickableDiv` or sth like that where by default it has all these properties set. I see them being repeated quite often in this PR ( ex:  `AddressOption`, `ExpandableSection Title`, `ActionButtonContainer`, etc). @LeeAlephium WDYT?

src/components/Popup.tsx
```diff
@@ -16,40 +16,58 @@ You should have received a copy of the GNU Lesser General Public License
...
+    }
+  }, [handleEscapeKeyPress, onBackgroundClick])
+
+  return (
+    <PopupContainer
+      initial={{ opacity: 0 }}
+      animate={{ opacity: 1 }}
+      exit={{ opacity: 0 }}
+      onClick={() => {
+        onBackgroundClick()
```
nop33, line 50: I know it's not your code, but this could be written as
```ts
onClick={onBackgroundClick}
```

src/modals/CenteredModal.tsx
```diff
@@ -44,39 +45,50 @@ const CenteredModal: FC<CenteredModalProps> = ({
...
-    </CenteredBox>
-  </ModalContainer>
-)
+}) => {
+  const { t } = useTranslation('App')
+  const divRef = useRef<HTMLDivElement>(null)
+
+  useEffect(() => {
+    divRef?.current?.focus()
+  }, [divRef])
```
nop33, line 54: I've seen this 3 times in the code, how about a hook?
```ts
const divRef = useFocusOnDivOnLoad()
```
or sth like that

src/modals/SettingsModal/NetworkSettingsSection.tsx
```diff
@@ -72,8 +72,10 @@ const NetworkSettingsSection = () => {
   }
 
   const handleNetworkPresetChange = useCallback(
-    (option: typeof networkSelectOptions[number] | undefined) => {
-      if (option && option.value !== selectedNetwork) {
+    (option?: SelectOption<NetworkName>) => {
+      if (option === undefined) return
+
+      if (option.value !== selectedNetwork) {
```
nop33, line 78: Why splitting it in 2 conditionals and not just one?
```ts
if (option === undefined || option.value === selectedNetwork) return
```

src/modals/SettingsModal/WalletsSettingsSection.tsx
```diff
@@ -116,15 +116,25 @@ const WalletItem = ({ walletName, isCurrent, onWalletDelete }: WalletItemProps)
...
-      <WalletName>
+      <WalletName role="cell" tabIndex={0} onFocus={() => setIsShowingDeleteButton(true)}>
         {walletName}
         {isCurrent && <CurrentWalletLabel> {t`(current)`}</CurrentWalletLabel>}
       </WalletName>
       {isShowingDeleteButton && (
-        <WalletDeleteButton squared transparent onClick={() => onWalletDelete(walletName)}>
+        <WalletDeleteButton
+          role="button"
+          aria-label="Delete"
```
nop33, line 131: Needs to be translated

src/modals/SettingsModal/WalletsSettingsSection.tsx
```diff
@@ -116,15 +116,25 @@ const WalletItem = ({ walletName, isCurrent, onWalletDelete }: WalletItemProps)
...
     >
-      <WalletName>
+      <WalletName role="cell" tabIndex={0} onFocus={() => setIsShowingDeleteButton(true)}>
         {walletName}
         {isCurrent && <CurrentWalletLabel> {t`(current)`}</CurrentWalletLabel>}
       </WalletName>
       {isShowingDeleteButton && (
-        <WalletDeleteButton squared transparent onClick={() => onWalletDelete(walletName)}>
+        <WalletDeleteButton
+          role="button"
```
nop33, line 130: Shouldn't the `Button` itself always have this role, instead of adding it everytime we style it differently?

src/pages/HomePage.tsx
```diff
@@ -140,16 +140,14 @@ const Login = ({ walletNames, onLinkClick }: LoginProps) => {
...
-      </SwitchLink>
+      <SwitchLink onClick={onLinkClick}>{t`Create / import a new wallet`}</SwitchLink>
     </>
   )
 }
 
 const InitialActions = ({
   showLinkToExistingWallets,
-  onLinkClick
+  onLinkClick = () => ({})
```
nop33, line 150: Why this default?

src/pages/Wallet/AddressesPage.tsx
```diff
@@ -93,11 +93,13 @@ const AddressesPage = () => {
         {sortAddressList(addresses).map((address) => {
           return (
             <TableRow
+              role="row"
```
nop33, line 96: I find some `TableRow` instances without the `role="row"`. Is this expected?

It'd be nice if we could define styled-components components with some default attrs, but that would require creating an actual React component... :/

For example, look at how repeated the `role="cell" tabIndex={0}` are in this screenshot...

<details><summary>Click to see scrshot</summary>

image

</details>

nop33 (commented):
@LeeAlephium I suppose that you must have learned so many things regarding accessibility while working on this PR! What would your recommendations be for @mvaivre and I for any future code that we will write in this project (and others)? What do we need to pay attention to? When should we use `role`, `tabIndex`, and all that? Shall we also include accesibility tests in every PR review from now on? How shall we do that?

I would appreciate sharing some of the new knowledge you aquired! Thanks! <3

src/pages/Wallet/WalletLayout.tsx
```diff
@@ -127,7 +127,7 @@ const WalletLayout: FC = ({ children }) => {
           <InfoBox text={activeWalletName} label={t`WALLET`} />
         ) : (
           <Select
-            label={t`WALLET`}
+            label={t`SWITCH WALLET`}
```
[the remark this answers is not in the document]
mvaivre: I also get where it comes from, but IMO this label shouldn't be an action.
However, if we believe that "Wallet" may be confusing, we may use "Current wallet" instead.

src/modals/CenteredModal.tsx
LeeAlephium: Yeah good idea

src/modals/SettingsModal/NetworkSettingsSection.tsx
LeeAlephium: Was focused on satisfying the typechecker - yeah one is better

src/modals/SettingsModal/WalletsSettingsSection.tsx
LeeAlephium: Ah yeah, since the "root" element / component is `motion.button` there's no need

src/pages/HomePage.tsx
LeeAlephium: So it's a no-op if nothing is passed and `SwitchLink` will still work if `showLinkToExistingWallets` is true

src/pages/Wallet/AddressesPage.tsx
LeeAlephium: When running the screenreader it was re-reading a lot of text when a role was applied on top of other roles

We should maybe peek at those large react component libraries which can be extended...

src/components/Inputs/ColorPicker.tsx
LeeAlephium: Explained offline: it's pretty situational but there are probably ways to make it DRYer, we can iterate

nop33:
@LeeAlephium nothing happens when I try to switch a network

https://user-images.githubusercontent.com/1579899/180176346-9c58f5d3-ead3-4c55-a410-753fb36f01dc.mp4

LeeAlephium:
> @LeeAlephium nothing happens when I try to switch a network

Corrected in the latest changes

LeeAlephium:
Converted back to draft since there are components which have been added that will need to be accounted for

nop33:
Not sure this is related to this PR, but the star icon looks quite big in comparison to the dot one in the transaction details modal, but not in the transactions list 😅 

image

src/components/Inputs/Select.tsx
```diff
@@ -136,29 +157,44 @@ function SelectOptionsPopup<T>({
...
-        </OptionItem>
-      ))}
+      <OptionSelect autoFocus size={options.length} onKeyPress={(e) => handleEvent(e.currentTarget)}>
+        {options.map((o) => (
+          <OptionItem
+            key={o.label}
+            value={o.value}
+            onClick={() => {
+              handleOptionSelect(o)
+            }}
```
nop33, line 180: nit: Could be written in one line

src/components/TabBar.tsx
```diff
@@ -81,7 +90,21 @@ const TabContainer = styled.div`
...
+  flex: 1;
+  background-color: ${({ theme }) => theme.global.accent};
+  z-index: -1;
+`
+
+const Tab = styled(({ isActive, onClick, onKeyPress, children, className }) => (
+  <div className={className} onClick={onClick} onKeyPress={onKeyPress} role="tab" tabIndex={0} aria-selected={isActive}>
+    {children}
+  </div>
+))`
```
nop33, line 107: That's an interesting approach..!

The alternative would take more lines to implement, so I kinda like this one. @mvaivre WDYT?

nop33:
I did some functional testing, but I didn't try to perform our whole test suite with the keyboard 😅

LeeAlephium:
> but the star icon looks quite big in comparison

Hm...! Could you open an issue please?

nop33:
This bug doesn't appear on `v1.4.0` branch, so it must be related to this PR.

LeeAlephium:
> This bug doesn't appear on v1.4.0 branch, so it must be related to this PR.

The bug is present there, it just doesn't manifest

This PR doesn't touch `AddressBadge` - after a quick peek the issue seems to be the use of `rem` instead of `px`. Using `px` the sizes are consistent across the application (which I find weird in itself and will look further into it)

src/components/TabBar.tsx
mvaivre: I personally prefer the usual separation of concerns :

```
// Components & logic first
<Components />
...
// Styles at the bottom
const styled(...)
```

Even though this solution is shorter, it may make the code less readable in the end IMHO. 
I've seen that you used a similar technique in `HiddenLabel` and `Truncate`, but without naming the components. So if I don't look at the file name, it's not easy to understand what the component do. And I find it less easy to parse the props they take, as the code is more cluttered, with the styling glued to the logic.

LeeAlephium: The only way to know what a component does is to read its code - take `AddressBadge` for example. It's not possible to know what it does just based on its name... Rather it identifies what behavior its associated with, so regardless, you will have to read the documentation (or code if lack thereof)...

Yep the code is more cluttered because our format rules enforce those props to be all on one line. :shrug: 

I don't mind either way. The complexity of the code is so minimal that written either way could be interpreted as easy to understand. Just let me know what you both would like. :+1:

mvaivre: >  Take `AddressBadge` for example. It's not possible to know what it does just based on its name

Well, I don't agree with that: if that constant is called `AddressBadge`, I know that it's supposed to be use as an address badge.

Now, when you read the following, how do you know what it's supposed to be used for ? 

```ts
export default styled(({ text, ...props }) => <div {...props}>{text}</div>)`
  position: absolute;
  left: -10000px;
  top: auto;
  width: 1px;
  height: 1px;
  overflow: hidden;
`
```

----------

> Yep the code is more cluttered because our format rules enforce those props to be all on one line.

It's not the one-liner that's annoying me, it's the lack of blank lines. Maybe it's just me but when I read & write some code, I use blank line to give the file some structure. Ie. I assume that two pieces of code with different goals are gonna be split by one or multiple blank lines. Here, logic and style are sticked together, and that's what I don't really like. 

But let's avoid wasting too much time on those things. @nop33, what's you final opinion on this? I honestly don't really care if some components are written that way, but in the long run, I'd prefer if we would stick to our previous ways which were working well for us. In the end, all I want is for us to write consistent, readable code :)

nop33: I get the arguments from both of you and I don't have a super strong preference, but I tend to go with @mvaivre's view.

LeeAlephium:
https://github.com/alephium/desktop-wallet/issues/342 for the address badge bug

LeeAlephium:
@mvaivre @nop33 was there something that wasn't addressed?

mvaivre (approved):
LGTM! 👏

Issue #303, opened 2022-07-18
Help > About menu is empty


Sakrecoer:
**Describe the bug**
Trying to access the "about" entre in the "help" menu has no effect

**To Reproduce**
1. Go to "help" in the top part of the window
2. Click on 'About'
3. Nothing happens

**Expected behavior**
Some pop-up with information about the version and maybe license

**Screenshots**
![image](https://user-images.githubusercontent.com/3484593/179549364-2b6f5db6-61f9-4094-aea0-e06d229bd091.png)

**Environment**
 - OS: Ubuntu 22.04
 - Version 1.4.0-rc4

PR #305, opened 2022-07-19
Show about dialog on Linux

LeeAlephium:
Resolves https://github.com/alephium/desktop-wallet/issues/303

### DIFF
--- a/public/electron.js
+++ b/public/electron.js
@@ -17,7 +17,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
 // Modules to control application life and create native browser window
-const { app, BrowserWindow, Menu, shell } = require('electron')
+const { app, BrowserWindow, dialog, Menu, shell } = require('electron')
 const path = require('path')
 const isDev = require('electron-is-dev')
 const contextMenu = require('electron-context-menu')
@@ -33,6 +33,10 @@ const gotTheLock = app.requestSingleInstanceLock()
 // Build menu
 
 const isMac = process.platform === 'darwin'
+const isWindows = process.platform === 'win32'
+const isLinux = !isMac && !isWindows
+
+console.log(isMac ? 'mac' : isWindows ? 'windows' : 'linux')
 
 const template = [
   ...(isMac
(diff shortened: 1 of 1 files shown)


public/electron.js
```diff
@@ -33,6 +33,10 @@ const gotTheLock = app.requestSingleInstanceLock()
 // Build menu
 
 const isMac = process.platform === 'darwin'
+const isWindows = process.platform === 'win32'
+const isLinux = !isMac && !isWindows
+
+console.log(isMac ? 'mac' : isWindows ? 'windows' : 'linux')
```
mvaivre, line 39: To be removed

nop33 (commented):
Can you show with a screenshot how this looks like? Thanks!

public/electron.js
```diff
@@ -33,6 +33,10 @@ const gotTheLock = app.requestSingleInstanceLock()
 // Build menu
 
 const isMac = process.platform === 'darwin'
+const isWindows = process.platform === 'win32'
+const isLinux = !isMac && !isWindows
```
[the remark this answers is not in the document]
LeeAlephium: Ah yea, the name `isLinux` is probably terrible. The same issue will most likely manifest on all the following platforms listed by nodejs process.platform (https://nodejs.org/api/process.html#processplatform):

* 'aix'
* 'freebsd'
* 'linux'
* 'openbsd'
* 'sunos'

Which all use "unix" window managers / desktop environments

LeeAlephium:
> Can you show with a screenshot how this looks like? Thanks!

Very basic:

public/electron.js
```diff
@@ -33,6 +33,8 @@ const gotTheLock = app.requestSingleInstanceLock()
 // Build menu
 
 const isMac = process.platform === 'darwin'
+const isWindows = process.platform === 'win32'
+const isNotMacOrWindows = !isMac && !isWindows
```
nop33, line 37: Is this var rly needed at the end? It's only used once. Why not just use `!isMac && !isWindows` directly in the condition?

LeeAlephium: I found it useful to abstract what "Is not mac or windows" means since the underlying check could become more complicated or possibly re-used in the future

mvaivre (approved):

nop33 (approved):

PR #310, opened 2022-07-21
DRYer accessibility code

nop33:


src/components/Buttons/ClickableArea.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const ClickableArea: FC<ClickableAreaProps> = ({ className, onClick, children }) => (
+  <div className={className} role="button" tabIndex={0} onClick={onClick} onKeyPress={onClick}>
+    {children}
+  </div>
+)
+
+export default styled(ClickableArea)`
+  cursor: pointer;
+`
```
LeeAlephium, line 35: Since this higher order component does more than listen for click events it would make sense to rename it to something which reflects what it does more closely... Considering it handles multiple inputs, it could be called `InputArea`?

src/components/Inputs/ColorPicker.tsx
```diff
@@ -45,13 +46,7 @@ const ColorPicker = ({ value, onChange }: ColorPickerProps) => {
...
   return (
     <ColorPickerContainer ref={ref}>
-      <Input
-        role="button"
-        tabIndex={0}
-        aria-label={t`Pick a color`}
-        onClick={handlePopupOpen}
-        onKeyPress={handlePopupOpen}
-      >
+      <Input aria-label={t`Pick a color`} onClick={handlePopupOpen}>
```
LeeAlephium, line 49: The screenreader (Orca) no longer picks up the label with this change, I'm investigating now

LeeAlephium: Issue is https://github.com/alephium/desktop-wallet/pull/310/files#diff-3ac83cbff04686c428c1043abdb1aa3819c25290ed4aec26be1becc172b0f168R27 does not accept all props (and so `aria-label` is ignored)

LeeAlephium (approved):

PR #316, opened 2022-08-08
Update address badge across the application

LeeAlephium:
Refreshes the address badge design and applies it to the whole application.

## Demonstration

https://user-images.githubusercontent.com/105806603/183493588-1c598c8b-af66-478f-a0e5-a67eace85456.mp4


src/utils/addresses.ts
```diff
@@ -30,6 +30,28 @@ export type AddressSettings = {
   color?: string
 }
 
+export type AddressVariant = Address | AddressSettings
```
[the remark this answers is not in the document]
nop33: Agreed with the proposal!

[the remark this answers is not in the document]
LeeAlephium: Rainbow.me seems to be the closest in terms of code-style goals so I'm using it as an example here (I looked at Metamask and Blue wallet also but they use mostly JS with a mix of TS).

Using `src/types` as the central place for types in a TypeScript project really removes the context for a lot of types and is a bit of an anti-pattern... Usually `src/typings` is used to type packages which have no types and that's about it. The `AddressVariant` could be seen as a "utility type" for addresses, if it helps justify why it should live there. I don't believe anything is "hidden" or "less visible"... In fact they may be less visible and consequently less used if they're put in another place!

Here is an example from rainbow.me's code, where you see types are sprinkled throughout the entire code base for many reasons:



So I don't agree with the proposal but if 2/3 people want to then ok :+1:

src/components/AddressBadge.tsx
```diff
@@ -17,42 +17,77 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+  display: flex;
+
+  ${({ truncate }) =>
+    truncate &&
     `
-  }}
+      white-space: nowrap;
+      overflow: hidden;
+      text-overflow: ellipsis;
+  `}
```
nop33, line 92: Can be rewritten as:
```suggestion
export default styled(AddressBadge)`
  padding: 6px 0 6px 0;
  display: flex;
  
  ${({ truncate }) =>
    truncate &&
    `
      white-space: nowrap;
      overflow: hidden;
      text-overflow: ellipsis;
  `}
```

src/contexts/addresses.tsx
```diff
@@ -353,7 +353,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
       if (addressesMetadata.length === 0) {
         saveNewAddress(
           new Address(wallet.address, wallet.publicKey, wallet.privateKey, 0, {
-            isMain: true,
+            isMain: isPassphraseUsed ? false : true,
```
nop33, line 356: I am not so sure about this change. Even though visually we don't have the concept of "main" address for passphrase-enabled wallets, I am afraid that our business logic might break when there is no single main address (all references to `mainAddress` will be broken). I would not make this change.

src/utils/addresses.ts
```diff
@@ -30,6 +30,28 @@ export type AddressSettings = {
...
+
+export function isAddressSettings(address: AddressVariant): address is AddressSettings {
+  return (address as AddressSettings).isMain !== undefined
+}
```
nop33, line 37: nit: No need for return if written as arrow function

src/contexts/addresses.tsx
LeeAlephium: When I was testing it everything seemed ok? I can revert it for now but it might be a good idea to investigate removing  the dependency of requiring a main address...

src/utils/addresses.ts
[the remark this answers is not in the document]
nop33: > we could maybe put all the address related types and type-checks (like isAddress, isAddressSettings) in src/types/addresses.d.ts?

FWIT, we already have `/src/types/addresses.ts` in the mobile wallet: https://github.com/alephium/mobile-wallet/blob/master/src/types/addresses.ts

So far I haven't encountered a problem with having types of the same context in one file, so I wouldn't change our pattern now until we encounter a problem.

Yes for `/src/types/addresses.ts`.

src/components/AddressBadge.tsx
```diff
@@ -17,42 +17,75 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+  } else {
+    data = {
+      color: theme.font.primary,
+      isMain: false,
+      label: ''
+    }
+  }
+
+  return (
+    <div className={className} {...props}>
```
nop33, line 53: hm, are `{...props}` really needed there, now that we don't use `Badge` anymore?

nop33 (commented):
REally good job 👏

src/components/AddressBadge.tsx
LeeAlephium: I think I thought it was necessary to pass the props so the styled component could see it, but no it's already getting them / typed I realize now

src/components/AddressBadge.tsx
```diff
@@ -17,42 +17,75 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+  ${({ color, isMain, theme }) => `
+    &::before {
+      width: 1rem;
+      display: block;
+      text-align: center;
+      margin-right: 0.2rem;
+      line-height: 1rem;
+      font-size: ${isMain ? '1.4em' : '1em'};
+      vertical-align: middle;
+      content: '${isMain ? '★' : '●'}';
```
nop33, line 86: The surounding `'` seems unnecessary, am I wrong?

src/components/Inputs/ColoredLabelInput.tsx
```diff
@@ -81,7 +83,7 @@ const InputStyled = styled(Input)`
     theme.name === 'dark' ? color : tinycolor(color).isLight() ? theme.font.primary : theme.font.contrastPrimary};
 
   &:not([value='']) {
-    padding-left: 20px;
+    padding-left: calc(1rem + 0.2rem + 1em);
```
nop33, line 86: Is this a magic number? 😅  I guess this is the summary of paddings and spacings of other elements, right? Which means if they get changed elsewhere they'd need to be changed here as well... Maybe exported vars would help here?

src/components/AddressBadge.tsx
LeeAlephium: Hehe, yeah this one "got me": it's necessary because `content`  takes a string, but I need to also return a string. The result css would be `content: ★` vs `content: '★'`

src/components/Inputs/ColoredLabelInput.tsx
LeeAlephium: It's other element widths yeah - I figured it's at the very least better than the old value. I think it's unnecessary "busy work" to make this more dynamic but sure if you think otherwise :+1:

nop33: Well, it's true that we don't rly have a fixed system of spacings with shared variables. But I still think it's worth to create vars when component spacings are so interdependent.

src/components/AddressBadge.tsx
nop33: Ah, I see the issue! The whole block is written as a string! HAve a look at this, I find it much better:

```ts
const Icon = styled.span<{ isMain: boolean; color?: string }>`
  &::before {
    width: 1rem;
    display: block;
    text-align: center;
    margin-right: 0.2rem;
    line-height: 1rem;
    font-size: ${({ isMain }) => (isMain ? '1.4em' : '1em')};
    vertical-align: middle;
    content: ${({ isMain }) => (isMain ? '★' : '●')};
    border-radius: 100%;
    color: ${({ color, theme }) => color ?? theme.font.primary};
  }
`
```

nop33: No wait, it doesn't work, sec.

nop33: ok yeah, you're totally right. However, I'd still use my recommendation above (after wrapping it with `'` of course)

LeeAlephium: :thinking: the suggestion just adds function noise?...

nop33: It's more consistent to how we write styled components. It also gives me syntax highlighting.

nop33: Anyway, it's a nit.

LeeAlephium: I respectfully ignore the nit :pray:

nop33 (approved):

PR #317, opened 2022-08-08
Update transaction history listings

### LINKED ISSUE #301: Address from/to not clear on Overview
On an address's transaction overview it's clear where funds are traveling because it shows "To/From"

But it's very unclear on the Overview

LeeAlephium:
Closes #301

Refreshes the transaction history to match this design with some slight changes:

# Overview view

Reference design:

(https://www.figma.com/file/wDWcIxJIq5aBTLwMPGb1Js/Alephium-Desktop-Wallet---DeFi-evolution?node-id=0%3A1 for full reference)

## Demonstration

https://user-images.githubusercontent.com/105806603/183493714-742190dc-2d57-4e11-89a9-ced22cb70aca.mp4

# Address details view

## Demonstration

https://user-images.githubusercontent.com/105806603/183493784-e7e57061-a5d5-4a5a-800c-30b1107e3ba0.mp4

Main difference is no headers and colors, with the goal being consistency. I apologize if the original intent of these was misunderstood (maybe the color and headings are 100% necessary here?) and please tell me if you'd like to see them match the design exactly!


src/components/DirectionalArrow.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { TransactionDirection } from '../utils/transactions'
+
+interface Props {
+  direction: TransactionDirection
+}
+
+const DirectionalArrow = ({ direction }: Props) => {
+  const isReceiving = direction === 'in'
+  const color = {
+    circle: isReceiving ? 'rgba(108, 217, 158, 0.11)' : 'rgba(153, 153, 153, 0.11)',
```
mvaivre, line 31: We should base those colors on theme values here (font secondary, "valid" color). 
We can use the `tinycolor` library to tweak the opacity of those values.

LeeAlephium:
Minor detail comments by @mvaivre on Slack:

* The number of additional addresses in parenthesis (+2) should be next to the address / label and not in a separated column

I tried this in an earlier version and I found it hard to follow compared to the current one:




* It should be in font secondary color

🫡

* The from/to text should also use the secondary color

These are actually different in your design file and where I pulled the color from:



"from" is bg-accent, "to" is bg-secondary

Which is precisely what I'm using:

```js
const DirectionBadgeOut = styled(BadgeStyled)`
  background-color: ${({ theme }) => theme.bg.secondary};
`

const DirectionBadgeIn = styled(BadgeStyled)`
  background-color: ${({ theme }) => theme.bg.accent};
`
```

Thoughts?

mvaivre:
> "from" is bg-accent, "to" is bg-secondary

`bg-accent` for the background. But I was talking about the *text* color, which should be `font.secondary` instead of `font.primary` (which I think it is, based on the screenshots you shared. I haven't yet checked the code properly!)

> I tried this in an earlier version and I found it hard to follow compared to the current one:

The main issue IMO is the various address lengths. Those should be cut using ellipsis.
And as mentioned, the additional number of address should be shown in `font.secondary` color, distinguishing it from the address string.

If this is not satisfactory, I'm planning to do a UI refresh round later anyway in #320  :)

LeeAlephium:
All good! I just wanted to make sure everything is understood :star: 

Here's a shot with the adjustments:



It certainly looks better than the staggering from before! I still wonder if aligned is better, but not much of a concern to me now since it's more or less "slightly" aligned.

On another note our current valid color isn't the nicest to use for the arrows (this is one reason why I use colors directly from your mocks vs what we have in the theme) - should I adjust the theme coloring to match or should we defer it?

mvaivre:
Thanks :) 

You can go ahead and adjust the "valid" color.

LeeAlephium:
It's all done, let me know if you'd like any more adjustments :triangular_ruler:

src/components/OverviewPage/TransactionList.tsx
```diff
@@ -46,69 +35,29 @@ const OverviewPageTransactionList = ({ className, onTransactionClick }: Overview
...
+  const allConfirmedTxs = getTransactionsForAddresses('confirmed', addresses)
+  const allPendingTxs = getTransactionsForAddresses('pending', addresses)
 
   const loadNextTransactionsPage = async () => {
     addresses.forEach((address) => fetchAddressTransactionsNextPage(address))
   }
 
   const showSkeletonLoading = isLoadingData && !allConfirmedTxs.length && !allPendingTxs.length
 
-  const transactionsTableHeadersI18n = transactionsTableHeaders.map((el) => ({ ...el, title: t(el.title) }))
```
nop33, line 65: Does that mean we need to clean up our App.json from unused keys?

src/components/Table.tsx
```diff
@@ -41,13 +41,15 @@ interface TableCellProps {
...
-          <TableHeaderCell key={title} align={align}>
-            {title}
-          </TableHeaderCell>
-        ))}
-      </TableHeaderRow>
+      {headers && headers.length > 0 && (
+        <TableHeaderRow columnWidths={headers.map(({ width }) => width)}>
+          {headers.map(({ title, align }) => (
+            <TableHeaderCell key={title} align={align}>
+              {title}
```
nop33, line 48: Is this translated?

src/components/Table.tsx
```diff
@@ -104,9 +106,8 @@ const TableColumns = styled.div<{ columnWidths?: (string | undefined)[] }>`
           grid-auto-flow: column;
         `};
 
-  padding: 0 var(--spacing-5);
   align-items: center;
-  height: var(--tableCellHeight);
+  padding: 17.5px 20px;
```
nop33, line 110: `.5px`? 😅

src/components/TimeSince.tsx
```diff
@@ -0,0 +1,36 @@
...
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import dayjs from 'dayjs'
+import styled from 'styled-components'
+
+interface TimeSinceProps {
+  timestamp: number
+  className?: string
+  faded?: boolean
```
nop33, line 25: super nit: I tend to always put `className` the last :)

src/components/TimeSince.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const TimeSince = ({ timestamp, className }: TimeSinceProps) => {
+  return <div className={className}>{dayjs(timestamp).fromNow()}</div>
+}
+
+const TimeSinceStyled = styled(TimeSince)`
+  ${({ faded, theme }) => (faded ? `color: ${theme.font.secondary}` : '')}
+`
+
+export default TimeSinceStyled
```
nop33, line 36: Can be written simpler using our convention:

```suggestion
export default styled(TimeSince)`
  ${({ faded, theme }) => (faded ? `color: ${theme.font.secondary}` : '')}
`

File: src/components/TimeSince.tsx [hunk 7]
```diff
@@ -0,0 +1,36 @@
...
+
+interface TimeSinceProps {
+  timestamp: number
+  className?: string
+  faded?: boolean
+}
+
+const TimeSince = ({ timestamp, className }: TimeSinceProps) => {
+  return <div className={className}>{dayjs(timestamp).fromNow()}</div>
+}
```
nop33, line 30: nit: It's a dump component, no need for the `return` statement! :)

```suggestion
const TimeSince = ({ timestamp, className }: TimeSinceProps) => (
  <div className={className}>{dayjs(timestamp).fromNow()}</div>
)
```

src/components/Token.tsx
```diff
@@ -0,0 +1,30 @@
...
+export type TokenType = 'alph'
+
+interface TokenProps {
+  type: TokenType
+  className?: string
+}
+
+const Token = ({ type, className }: TokenProps) => {
+  return <div className={className}>{type.toUpperCase()}</div>
+}
```
nop33, line 28:
```suggestion
const Token = ({ type, className }: TokenProps) => <div className={className}>{type.toUpperCase()}</div>
```

src/components/TransactionalInfo.tsx
```diff
@@ -16,35 +16,185 @@ You should have received a copy of the GNU Lesser General Public License
...
+import Badge from './Badge'
+import DirectionalArrow from './DirectionalArrow'
+import IOList from './IOList'
+import TimeSince from './TimeSince'
+import Token from './Token'
 
 interface TransactionalInfoProps {
-  type: 'out' | 'in' | 'pending'
-  content: string | bigint
+  transaction: Transaction | SimpleTx
```
nop33, line 36: How about renaming `SimpleTx` to `PendingTx`?

/cc @mvaivre

src/components/TransactionalInfo.tsx
```diff
@@ -16,35 +16,185 @@ You should have received a copy of the GNU Lesser General Public License
...
+  addressHash?: string
   className?: string
-  amount?: boolean
-  prefix?: string
+  hideLabel?: boolean
```
nop33, line 39: nit: put above `className` 😇

src/components/TransactionalInfo.tsx
```diff
@@ -16,35 +16,185 @@ You should have received a copy of the GNU Lesser General Public License
...
+import DirectionalArrow from './DirectionalArrow'
+import IOList from './IOList'
+import TimeSince from './TimeSince'
+import Token from './Token'
 
 interface TransactionalInfoProps {
-  type: 'out' | 'in' | 'pending'
-  content: string | bigint
+  transaction: Transaction | SimpleTx
+  addressHash?: string
```
nop33, line 37: shouldn't this be of type `AddressHash`?

src/components/TransactionalInfo.tsx
```diff
@@ -16,35 +16,185 @@ You should have received a copy of the GNU Lesser General Public License
...
+  } else {
+    throw new Error('Could not determine transaction type, all transactions should have a type')
+  }
 
   return (
-    <div className={className} color={color}>
-      {prefix && <span>{prefix}</span>}
-      {amount && content ? <Amount value={BigInt(content)} fadeDecimals color={color} /> : content}
-    </div>
+    <Container className={className} theme={theme}>
```
nop33, line 76: To avoid passing `theme`, this component could be replaced by a `div`. Then there is no need for the `Container` component. Instead, you can replace it with `export default styled(TransactionalInfo)`... and this styled component will already have access to the props and theme.

Edit: Every styled component has access to theme actually

src/components/TransactionalInfo.tsx
```diff
@@ -16,35 +16,185 @@ You should have received a copy of the GNU Lesser General Public License
...
+        <CellArrow>
+          <DirectionalArrow direction={type} />
+        </CellArrow>
+        <TokenTimeInner>
+          <TokenStyled type={token} />
+          <TimeSince timestamp={timestamp} faded />
+        </TokenTimeInner>
+      </CellAmountTokenTime>
+      {address && (
+        <>
```
nop33, line 87: No need for fragment

src/utils/transactions.ts
```diff
@@ -17,7 +17,71 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+
+export type TransactionDirection = 'out' | 'in' | 'pending'
+export type TransactionType = 'consolidation' | 'transfer' | 'sweep'
+export type TransactionStatus = 'pending' | 'confirmed'
+
+export function getTransactionsForAddresses(
+  txStatus: TransactionStatus,
+  addresses: Address[]
+): BelongingToAddress<TransactionVariant>[] {
+  return addresses
```
nop33, line 45: nit: no need for return if converted to arrow function (which we tend to prefer in our code base)

src/pages/Wallet/AddressDetailsPage.tsx
```diff
@@ -140,63 +128,19 @@ const AddressDetailsPage = () => {
...
       <PageH2>{t`Transaction history`}</PageH2>
-      <Table headers={transactionsTableHeaders} minWidth="500px">
+      <Table minWidth="500px">
         {address.transactions.pending
           .slice(0)
           .reverse()
-          .map(({ txId, timestamp, toAddress, amount, type }) => (
-            <TableRow key={txId} columnWidths={tableColumnWidths} blinking>
-              <TableCell>
-                <TransactionalInfo content={t`Pending`} type="pending" />
```
nop33, line 150: Shouldn't the App.json be cleaned after this removal?

src/components/TransactionalInfo.tsx
```diff
@@ -16,35 +16,185 @@ You should have received a copy of the GNU Lesser General Public License
...
+          <CellAddress>
+            {!hideLabel && (
+              <CellAddressBadge>
+                <AddressBadge address={address} truncate />
+              </CellAddressBadge>
+            )}
+            <DirectionalAddress>
+              {type === 'out' && <DirectionBadgeOut>{t`to`}</DirectionBadgeOut>}
+              {type !== 'out' && <DirectionBadgeIn>{t`from`}</DirectionBadgeIn>}
```
nop33, line 96: Add both in App.json

src/pages/Wallet/AddressDetailsPage.tsx
```diff
@@ -140,63 +128,19 @@ const AddressDetailsPage = () => {
...
-              columnWidths={tableColumnWidths}
-              onClick={() => onTransactionClick(transaction)}
-            >
-              <TableCell>
-                <TransactionalInfo content={isOut ? '↑ ' + t`Sent` : '↓ ' + t`Received`} type={isOut ? 'out' : 'in'} />
-              </TableCell>
-              <TableCell>{dayjs(transaction.timestamp).fromNow()}</TableCell>
-              <TableCell>
-                <DirectionalAddress>
-                  <DirectionBadge>{isOut ? t`To` : t`From`}</DirectionBadge>
```
nop33, line 181: Remove both from App.json

src/components/OverviewPage/TransactionList.tsx
```diff
@@ -46,69 +35,29 @@ const OverviewPageTransactionList = ({ className, onTransactionClick }: Overview
...
-
+      {allConfirmedTxs.map(({ data: tx, address }: BelongingToAddress<Transaction>) => {
         return (
-          <TableRow
-            key={`${transaction.hash}-${transaction.address.hash}`}
-            columnWidths={tableColumnWidths}
-            onClick={() => onTransactionClick(transaction)}
-          >
-            <TableCell>
-              <TransactionalInfo content={isOut ? '↑ ' + t`Sent` : '↓ ' + t`Received`} type={isOut ? 'out' : 'in'} />
```
nop33, line 98: Remove both from App.json

nop33 (commented):
Good job in this PR @LeeAlephium !

src/components/Table.tsx
LeeAlephium: Any titles that are passed in will have to be. The component itself shouldn't be responsible for this, but rather the user that is passing in the title (like you commented above, where I'll now need to remove translations since I created them in that particular file to be passed into this component).

src/components/Table.tsx
LeeAlephium: It was the distance in the mock :shrug:

src/components/OverviewPage/TransactionList.tsx
LeeAlephium: Ah yeah, I missed one: `Direction`. The rest are still used by other components.

src/components/TransactionalInfo.tsx
LeeAlephium: Just "to" was needed

src/pages/Wallet/AddressDetailsPage.tsx
LeeAlephium: These are still used in the transaction details modal

src/components/OverviewPage/TransactionList.tsx
LeeAlephium: These are still used in the transaction details modal

src/components/TransactionalInfo.tsx
LeeAlephium: I'm not a fan of the suggestion to remove the locality of `Container` since it's used by the component for its specific properties, but it's done :sparkles: 

I'm not sure why theme was being passed in the first place, it was never used :shrug:

src/components/TransactionalInfo.tsx
LeeAlephium: Seems out of scope to do it in this PR but it was quick to do, so it's done

src/components/TimeSince.tsx
LeeAlephium: Note this immediate exporting only works for this particular set of components (`styled(Component)`) this will fail the linter rules if trying to immediately export any arrow-function components, i.e. `export default () => ...`

src/components/TimeSince.tsx
LeeAlephium: We use `const A = () => { return ... }` 18 other times in the project, could you explain how they're different than this? Like this little one:

```
d823a185 src/pages/WalletManagement/CheckWordsPage.tsx (Ilias Trichopoulos 2022-01-04 17:02:25 +0100 245) const getAlphabeticallyOrderedList = (arr: string[]) => {
d823a185 src/pages/WalletManagement/CheckWordsPage.tsx (Ilias Trichopoulos 2022-01-04 17:02:25 +0100 246)   return arr.slice().sort()
d823a185 src/pages/WalletManagement/CheckWordsPage.tsx (Ilias Trichopoulos 2022-01-04 17:02:25 +0100 247) }
```

I'm building up my external eslintrc file so we can reduce these inconsistencies, I'm just trying to understand the particularities so we can avoid any discussion about code style like the Go people :muscle:

src/components/OverviewPage/TransactionList.tsx
nop33: Aren't they with lower case there? I think it's important to distringuish.

src/pages/Wallet/AddressDetailsPage.tsx
nop33: Aren't they with lower case there? I think it's important to distringuish.

src/components/TimeSince.tsx
nop33: yes, of course

LeeAlephium:
Are there any more points I need to address?

src/components/Table.tsx
mvaivre: Haha, as I said, the mockups are kind of alive and not 100% pixel perfect. I would spend much more time and be much less creative if I wanted them to be absolutely perfect all the time.

Let's round values when this happens.

mvaivre:
Great work! As I'm going on holidays, I'll let @nop33 be the judge for the approval!

src/components/Table.tsx
nop33: ping @LeeAlephium

src/components/TransactionalInfo.tsx
```diff
@@ -16,35 +16,180 @@ You should have received a copy of the GNU Lesser General Public License
...
+  const { t } = useTranslation('App')
+
+  const address = getAddress(_addressHash)
+
+  let amount: bigint | undefined = BigInt(0)
+  let timestamp = 0
+  let type: TransactionDirection
+  let outputs: Output[] = []
+
+  const token = 'alph'
```
nop33, line 56: I see an export of this in `src/components/Token.tsx`. Maybe we can reuse it?

nop33 (commented):
Really good job 👏 

I run the app and saw this in the console:

```
Warning: Received `true` for a non-boolean attribute `truncate`.

If you want to write it to the DOM, pass a string instead: truncate="true" or truncate={value.toString()}.
    at div
    at AddressBadge (http://localhost:3000/static/js/main.chunk.js:1572:5)
```

---

Moreover, I have a question. Look at this particular example. I see the `(+4)` next to the address that the funds were sent from, but when I see the details of the transaction I see that all the inputs are from the same address. Wouldn't it make more sense to show the `(+X)` only when the inputs are from different addresses?

image

src/components/Table.tsx
LeeAlephium: Yep, all good! It was an answer to the "why" - There was a PR when I first started where the mocks were to be followed very strictly, and there were a few times when it wasn't, so I've learned from then to be more strict first and then re-adjust as needed in the PR review

LeeAlephium:
> Moreover, I have a question. Look at this particular example. I see the (+4) next to the address that the funds were sent from, but when I see the details of the transaction I see that all the inputs are from the same address. Wouldn't it make more sense to show the (+X) only when the inputs are from different addresses?

Good question... This is related to the change address stuff also. I guess the + should reflect a more high level of addresses vs every individual utxo address. I'll adjust how it's calculated then

src/components/TransactionalInfo.tsx
LeeAlephium: That's a type, not a value

LeeAlephium:
https://github.com/alephium/desktop-wallet/issues/331 is resolved now as well

nop33 (approved):
Amazing job, well done 👏

LeeAlephium:
> Warning: Received `true` for a non-boolean attribute `truncate`.

I forgot to fix this :facepalm:

LeeAlephium:
(The remaining issue doesn't exist anymore for whatever reason)

PR #318, opened 2022-08-08
Update transaction history for address details

LeeAlephium:
Refreshes the transaction history in the address details to match this design of the new one on the overview:

Main difference is no headers and colors, with the goal being consistency. I apologize if the original intent of these was misunderstood (maybe the color and headings are 100% necessary here?) and please tell me if you'd like to see them match the design exactly!

## Demonstration

https://user-images.githubusercontent.com/105806603/183493784-e7e57061-a5d5-4a5a-800c-30b1107e3ba0.mp4


LeeAlephium:
I hit "close" and I guess because I had merged this on the client-side, it's showing merged (and even emitted an event to our Slack)... I only meant to close because those two PRs are too coupled to deal with separately.

Issue #313, opened 2022-08-02
Use custom CSS scrollbars for consistent look across OSes


mvaivre:
The appearance of the scrollbars on MacOS and most Linux distros is 🤢, especially with the dark theme on.

![image](https://user-images.githubusercontent.com/6715943/182387257-874c5f01-4e10-4079-b83e-3bc5232fa33b.png)

Good news: we know for sure that we're running on webkit! Thus, we don't have to use JS to create pseudo-scrollbars, we can leverage this: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

LeeAlephium:
@mvaivre or @nop33 what does macos do when the scrollbar is on a black background? Does it turn light grey?

nop33:
yes

<img width="87" alt="image" src="https://user-images.githubusercontent.com/1579899/183706843-5cb438c4-a522-4367-883b-a682b139e2d1.png">

mvaivre:
As a reference, here's a demo of scrollbar interactions on macos. I doubt we can do exactly that with CSS only, but let's try to follow this minimalistic approach!

https://user-images.githubusercontent.com/6715943/183846304-2b728f82-af31-4211-acfd-a93fb995e6ba.mov

LeeAlephium:
@mvaivre to be sure, you saw https://github.com/alephium/desktop-wallet/pull/324#issuecomment-1209775360, and we want something that is much closer to the video demo, right?

mvaivre:
I had not see it nope! It's a very good start, thanks. Ideally yes, we'd like it to be closer to what's shown above:
- With a little padding so that it doesn't touch the side of the window
- A bit less contrasty (less dark / light, depending on the active theme)
- Let's make the scrollbars appear only when hovering the scrollable area (a la ledger live, or Slack)

And no track (as you made it).

LeeAlephium:
Adding padding and having nice fade in/out is going to make this way less trivial - Slack for example creates a scrollbar component and uses this. I'll do the same then if we want these... At least we'll have very configurable scrollbars then :laughing:

LeeAlephium:
https://github.com/malte-wessel/react-custom-scrollbars/issues/406 :facepalm:

LeeAlephium:
https://github.com/xobotyi/react-scrollbars-custom is working perfectly, I just have to adjust a few other places in the application where scrollbars appear and it's all good

PR #178, opened 2022-03-23
Allow horizontal scrolling with wheel

nop33:
Closes #174

I would actually need people with mise and wheels to test this for me.

Issue #174, opened 2022-03-23
Not able to scroll through addresses with the mouse wheel


mvaivre:
![image](https://user-images.githubusercontent.com/6715943/159703503-d64bf065-81c9-4548-b1be-d5ced6568518.png)

Horizontal scrolling doesn't seem to work with a real mouse wheel (no trackpad).

nop33:
This might help

https://alvarotrigo.com/blog/scroll-horizontally-with-mouse-wheel-vanilla-java/

PR #324, opened 2022-08-09
Use a consistent scrollbar styling across platforms

LeeAlephium:
Resolves https://github.com/alephium/desktop-wallet/issues/313

https://user-images.githubusercontent.com/105806603/184899277-812232fb-94e2-4683-af09-5cd700e444a5.mp4

### DIFF
--- a/src/components/Scrollbar.tsx
+++ b/src/components/Scrollbar.tsx
@@ -20,8 +20,10 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import React, { CSSProperties, useState } from 'react'
 import Scrollbar, { ScrollbarProps } from 'react-scrollbars-custom'
-import { ElementPropsWithElementRef } from 'react-scrollbars-custom/dist/types/types'
+import { ElementPropsWithElementRef, ScrollState } from 'react-scrollbars-custom/dist/types/types'
 import { useTheme } from 'styled-components'
+
+import { ScrollContextProvider } from '../contexts/scroll'
 
 const paddingRight = '6px'
 const width = `calc(6px + ${paddingRight})`
@@ -42,6 +44,7 @@ interface ScrollbarCustomProps extends ScrollbarProps {
 
 const ScrollbarCustom = (props: ScrollbarCustomProps) => {
   const theme = useTheme()
+  const [scroll, setScroll] = useState<ScrollState>()
   const [isScrolling, setIsScrolling] = useState(false)
   const [isMouseOver, setIsMouseOver] = useState(false)
   const isShow = isScrolling || isMouseOver
@@ -101,6 +104,7 @@ const ScrollbarCustom = (props: ScrollbarCustomProps) => {
       thumbYProps={thumbProps}
       onScrollStart={onScrollStart}
       onScrollStop={onScrollStop}
+      onUpdate={(s: ScrollState) => setScroll(s)}
       scrollDetectionThreshold={500} // ms
       onMouseEnter={onMouseEnter}
       onMouseLeave={onMouseLeave}
@@ -108,7 +112,7 @@ const ScrollbarCustom = (props: ScrollbarCustomProps) => {
       noScrollY={props.noScrollY}
       translateContentSizeYToHolder={props.translateContentSizeYToHolder}
     >
-      {props.children}
+      <ScrollContextProvider value={{ scroll }}>{props.children}</ScrollContextProvider>
     </Scrollbar>
   )
 }
--- a/src/components/Scrollbar.tsx
+++ b/src/components/Scrollbar.tsx
@@ -18,7 +18,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 // Used as reference: https://github.com/xobotyi/react-scrollbars-custom/issues/46#issuecomment-897506147
 
-import React, { CSSProperties, useState } from 'react'
+import React, { CSSProperties, useCallback, useRef, useState, WheelEvent } from 'react'
 import Scrollbar, { ScrollbarProps } from 'react-scrollbars-custom'
 import { ElementPropsWithElementRef, ScrollState } from 'react-scrollbars-custom/dist/types/types'
 import { useTheme } from 'styled-components'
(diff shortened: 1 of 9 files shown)


LeeAlephium:
@mvaivre too subtle? Should it include a track?


https://user-images.githubusercontent.com/105806603/183742199-7b60b771-a7ef-4156-ac68-2c3665f17956.mp4

mvaivre:
https://github.com/alephium/desktop-wallet/issues/313#issuecomment-1211610712

nop33:
Pretty nice demo man ✨ I'll review it soon!

nop33 (commented):
Nice job! 👏 It looks indeed much sleeker now!

Here's my functional review:

A few differences I noticed with the standard mac scrollbars:
1. Scrollbar is now always visible when the cursor is on the scrollable surface. On mac it should appear only when scrolling inside the scrollable surface (see video).
2. Scrollbar does not become wider on hover. On mac it does (see video).
3. Scrolling does not work when the cursor is on top of the scrollbar. On mac it does (see video).
4. This feature is not working anymore: https://github.com/alephium/desktop-wallet/issues/174

https://user-images.githubusercontent.com/1579899/185056022-9fe6929f-eb39-42c6-b083-2d614150e859.mp4

LeeAlephium:
1 & 2: From what I understood from @mvaivre it doesn't have to be *exactly* like how macos does its scrollbars (I asked about the widening, for example) - see https://github.com/alephium/desktop-wallet/issues/313#issuecomment-1211610712 - otherwise I would've done it that way

3: This may be difficult to fix but I'll give it a go!

4: Horizontal scrolling is working for me - tilting left/right on my scroll wheel is scrolling as expected (scrolling up/down on a horizontal scroll should not work)

nop33:
> 1 & 2: From what I understood from @mvaivre it doesn't have to be exactly like how macos does its scrollbars (I asked about the widening, for example) - see https://github.com/alephium/desktop-wallet/issues/313#issuecomment-1211610712 - otherwise I would've done it that way

Hm, I thought it was fine to not replicate the mac functionality in other systems, but I don't think that meant that we should change the default mac functionality on mac.

> 4: Horizontal scrolling is working for me - tilting left/right on my scroll wheel is scrolling as expected (scrolling up/down on a horizontal scroll should not work)

This feature allowed scrolling vertically with the mouse while the cursor was in the addresses bar, and it would scroll horizontally. Check out my PR with the solution.

nop33:
I noticed that you actually implemented it as @mvaivre suggested (a la Slack, Ledger live). So, if he's OK with overriding the mac default behavior in order to fix the scroll bars in other platforms then I'm fine to proceed. I'd personally rather only apply the scrollbars on windows and linux and leave mac as they are. If it's even feasible.

LeeAlephium:
3 & 4: Fixed, but we can see fixing 4 was non-trivial. I would prefer removing this commit in favor of code simplicity.

nop33:
> fixing 4 was non-trivial (pretty sure I spent ~2 hours on it). I would prefer removing https://github.com/alephium/desktop-wallet/commit/4928a9d949a37e3ecb8806527184f8c910a6da81 in favor of code simplicity.

@LeeAlephium did you have a look at my original solution to this? https://github.com/alephium/desktop-wallet/pull/178/files

This was a request from the community that we implement it, it wouldn't be very nice to take it back...

LeeAlephium:
> did you have a look at my original solution to this?

Yes, you'll see it's essentially the exact same https://github.com/alephium/desktop-wallet/pull/324/commits/5c933da40dcc37371eff2d2f34228c91f9f99ff5 . I mistakenly wrote 4 when I meant point 3 (but I linked the correct commit)

nop33:
I noticed that the cards are not aligned in the center (vertically) in respect to the overview card. I also noticed that the bottom corner is nor rounded anymore

image

LeeAlephium:
Nice catch! Yea the way they layout is done for cards and the way the scrollbar operates is not very compatible with each other...

LeeAlephium:
@nop33 fixed

mvaivre:
Functional test: great work mate, it looks great! ✨ 
One weird artefact I've discovered though (white line on top of the scroll bar):

https://user-images.githubusercontent.com/6715943/186190687-60c993a1-61fa-4a71-9421-28b130b6cc19.mov

LeeAlephium:
Ah, that's the Mac scrollbar peeking through the top... 15px was enough to hide the default styled one on Linux (and presumably Windows since they use the same style). I'll increase this

LeeAlephium:
@mvaivre or @nop33 it'd be best actually if one of you adjusted this accordingly, otherwise this may take a couple of guesses to get right.

src/components/Scrollbar.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+const paddingRight = '6px'
+const width = `calc(6px + ${paddingRight})`
+
+const paddingTop = '6px'
+const height = `calc(6px + ${paddingTop})`
+
+const createScrollbarPiece = (regular: CSSProperties, additional?: CSSProperties, useAdditionalStyling?: boolean) => ({
+  renderer: ({ elementRef, style, ...restProps }: ElementPropsWithElementRef) => (
+    <div {...restProps} ref={elementRef} style={{ ...style, ...regular, ...(useAdditionalStyling && additional) }} />
```
nop33, line 36: Hm, how come `...(useAdditionalStyling && additional)` is not throwing any errors? If `useAdditionalStyling` is `false`, then this becomes `...false` which is not valid.

src/components/Scrollbar.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const scrollerElemRef = useRef<HTMLDivElement | null>(null)
+
+  const isShow = isScrolling || isMouseOver
+
+  const onScrollStart = () => setIsScrolling(true)
+  const onScrollStop = () => setIsScrolling(false)
+  const onMouseEnter = () => setIsMouseOver(true)
+  const onMouseLeave = () => setIsMouseOver(false)
+
+  const { isDynamic, noScrollX } = props
```
nop33, line 59: nit: Usually we deconstruct props next to the component definition, although it's not a hard rule, so all good here. Later on, however, I see that you are using props like `props.elementRef` or `props.noScrollY`, but here you're deconstructing only some of them (`isDynamic, noScrollX`). How about we either deconstruct all that we need here, or better, at the component definition? 😇

src/components/Scrollbar.tsx
LeeAlephium: It's perfectly valid:

```js
> { ...false }
{}
```

src/components/Scrollbar.tsx
LeeAlephium: I wouldn't deconstructed `elementRef` or `noScrollY` because it's just adding duplicate text since they're only used once to pass through, but absolutely this could move to being deconstructed in the function arguments

nop33 (approved):
👏

Issue #322, opened 2022-08-09
Remove tests for extension pack


nop33:
We are not using the app to build an extension so we could improve the test time by removing the tests for packing an extension.

PR #332, opened 2022-08-16
Remove building and testing as an extension

LeeAlephium:
The application is not designed to be an extension so this removes the building and testing of it as an extension.

Resolves https://github.com/alephium/desktop-wallet/issues/322


LeeAlephium:
We don't know to what extent the scripts even work so it follows to not maintain them or keep them around if we're not building an extension...

I'll checkout how the other files are used, I assumed Electron still uses these for whatever reason

LeeAlephium:
It (and `manifest.json`) wasn't use anywhere so it's been removed (I tested a build as well to be sure it wasn't something hidden either)

nop33 (approved):

PR #335, opened 2022-08-17
Ilias reflect wallet delete

nop33:
@LeeAlephium what do you think of these changes? Instead of exporting `set` methods in the global context, I replaced them with `saveWallet` and `deleteWallet`. It feels cleaner to me that way because we can reuse `saveWallet` instead of basically duplicating its code in the import and create wallet procedure. It also fits well imo next to the other methods of the global context (`lockWallet`, `unlockWallet` - I took the opportunity to rename `login` to `unlockWallet`). This way we also limit the places where we use the `Storage` object from the `getStorage()` SDK method. Lines of code also reduced.


LeeAlephium:
Nice improvement, I appreciate it!

PR #339, opened 2022-08-23
Synchronize theme between application and OS

LeeAlephium:
Special thanks to diomark and @sven-hash for help test

### DIFF
--- a/public/preload.js
+++ b/public/preload.js
@@ -1,3 +1,21 @@
+/*
+Copyright 2018 - 2022 The Alephium Authors
+This file is part of the alephium project.
+
+The library is free software: you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+The library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
 const { contextBridge, ipcRenderer } = require('electron')
 
 contextBridge.exposeInMainWorld('electron', {
--- a/public/preload.js
+++ b/public/preload.js
@@ -1,0 +1,13 @@
+const { contextBridge, ipcRenderer } = require('electron')
+
+contextBridge.exposeInMainWorld('electron', {
+  changeTheme: (theme) => ipcRenderer.invoke('changeTheme', theme),
+  onUpdateThemeDark: (cb) => {
+    ipcRenderer.on('update:theme:dark', cb)
+    return () => ipcRenderer.removeListener('update:theme:dark', cb)
+  },
+  onUpdateThemeLight: (cb) => {
+    ipcRenderer.on('update:theme:light', cb)
+    return () => ipcRenderer.removeListener('update:theme:light', cb)
+  }
+})
--- a/public/preload.js
+++ b/public/preload.js
@@ -19,7 +19,6 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 const { contextBridge, ipcRenderer } = require('electron')
 
 contextBridge.exposeInMainWorld('electron', {
-  changeTheme: (theme) => ipcRenderer.invoke('changeTheme', theme),
   onUpdateThemeDark: (cb) => {
     ipcRenderer.on('update:theme:dark', cb)
     return () => ipcRenderer.removeListener('update:theme:dark', cb)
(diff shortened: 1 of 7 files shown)


LeeAlephium:
It was a pain but I've done it!

https://user-images.githubusercontent.com/105806603/186521333-a4c2c273-f073-46be-a638-7f1ffe6f918e.mp4

public/preload.js
```diff
@@ -0,0 +1,13 @@
...
+
+contextBridge.exposeInMainWorld('electron', {
+  changeTheme: (theme) => ipcRenderer.invoke('changeTheme', theme),
+  onUpdateThemeDark: (cb) => {
+    ipcRenderer.on('update:theme:dark', cb)
+    return () => ipcRenderer.removeListener('update:theme:dark', cb)
+  },
+  onUpdateThemeLight: (cb) => {
+    ipcRenderer.on('update:theme:light', cb)
+    return () => ipcRenderer.removeListener('update:theme:light', cb)
```
LeeAlephium, line 11: I had to split these because it appears that passing data and not immediately binding callbacks causes weird behavior.

src/contexts/global.tsx
```diff
@@ -228,6 +230,25 @@ export const GlobalContextProvider: FC<{ overrideContextValue?: PartialDeep<Glob
...
+        /* do nothing */
+      }
+      const removeListeners: (() => void)[] = []
+      removeListeners.push(
+        _window.electron?.onUpdateThemeDark(() => updateSettings('general', { theme: 'dark' })) || noop
+      )
+      removeListeners.push(
+        _window.electron?.onUpdateThemeLight(() => updateSettings('general', { theme: 'light' })) || noop
+      )
+      return () => removeListeners.forEach((r) => r())
```
nop33, line 246: Instead of having this noop/doNothing function, wouldn't it work the same if we just check first if `_window.electron` is defined and in this case push the function call to the listeners array?

```ts
  useEffect(() => {
    const removeListeners: (() => void)[] = []
    const electron = _window.electron

    if (electron) {
      electron.changeTheme(settings.general.theme)

      removeListeners.push(electron.onUpdateThemeDark(() => updateSettings('general', { theme: 'dark' })))
      removeListeners.push(electron.onUpdateThemeLight(() => updateSettings('general', { theme: 'light' })))
    }

    return () => removeListeners.forEach((r) => r())
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [])
```

nop33 (commented):
Thanks @LeeAlephium for tackling this! 🙏 It seems to have been quite complicated...

FYI, functional tests are not working:

https://user-images.githubusercontent.com/1579899/186611233-2a67e821-6d6e-4ef2-b313-5967fab01a2c.mp4

src/contexts/global.tsx
```diff
@@ -228,6 +230,25 @@ export const GlobalContextProvider: FC<{ overrideContextValue?: PartialDeep<Glob
     }
   }, [currentNetwork, networkStatus, t])
 
+  useEffect(
+    () => {
+      _window.electron?.changeTheme(settings.general.theme)
+      const noop = () => {
+        /* do nothing */
+      }
```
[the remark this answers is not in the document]
LeeAlephium: The linter will complain otherwise and adding a comment is what's recommended

src/contexts/global.tsx
LeeAlephium: Yeah, looks good :slightly_smiling_face:

src/contexts/global.tsx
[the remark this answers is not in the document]
LeeAlephium: I'm going to follow your refactor below anyway :ok_hand:

LeeAlephium:
@nop33 suggestions applied, but either you or @mvaivre will need to debug that theme switching issue on Mac. I have zero idea as to why that's happening, and I can't dive deeper without a Mac. I've gone over the code a multitude of times and just can't figure why it's different for only Mac.

LeeAlephium:
@mvaivre your changes break it on Linux and Windows - the window theme no longer follows the internal app theme:

https://user-images.githubusercontent.com/105806603/186903159-e3cd1e4b-6436-423a-a449-2a56387264f3.mp4

The code you removed was for that

LeeAlephium:
There's going to be new behavior since Mac is acting wonky:

@mvaivre on Slack:

* When opening the app the first time, we use the OS theme (`system`)
* If the user manually switch the theme using the toggle, we change it to either lightor dark
* If the user changes the OS’s theme, we don’t override the app’s theme. After all, it’s only a click away.

src/contexts/global.tsx
```diff
@@ -232,16 +233,18 @@ export const GlobalContextProvider: FC<{ overrideContextValue?: PartialDeep<Glob
...
-
-      if (electron) {
-        electron.changeTheme(settings.general.theme)
-        removeListeners.push(electron.onUpdateThemeDark(() => updateSettings('general', { theme: 'dark' })))
-        removeListeners.push(electron.onUpdateThemeLight(() => updateSettings('general', { theme: 'light' })))
-      }
-
-      return () => removeListeners.forEach((r) => r())
+      const removeListener = _window.electron?.onGetNativeTheme((nativeTheme) => {
+        console.log(nativeTheme)
```
mvaivre, line 237: Console.log :)

mvaivre:
Seems to work well on MacOS 👍

mvaivre (approved):

PR #341, opened 2022-08-24
nit: Add padding lines

nop33:
@LeeAlephium I just added some padding lines based on my preference. It's a nit PR. I think @mvaivre also prefers to have empty lines to separate var declarations, and different statements. How about we use this eslint rule from now on? https://eslint.org/docs/latest/rules/padding-line-between-statements


LeeAlephium:
Yes, let's add the rule

Issue #342, opened 2022-08-25
Address Badge sizes are inconsistent


LeeAlephium:
Sometimes a star is larger larger a dot, and then those are sometimes different sizes between modals.

mvaivre:
Thanks for this issue. Could we take this opportunity to make sure that the padding between the start & label is also consistent with the padding between the dot and the label? We may need to reduce the start size a bit.
Thanks!

LeeAlephium:
It should have been consistent but I'll make sure they are :+1:

Issue #334, opened 2022-08-16
Copying/pasting wallet address from top bar should copy the full address


diomark:
If you try to copy the text field that shows the address name/label at the top, it only copies the first part -

i.e. -
![image](https://user-images.githubusercontent.com/20622085/184937574-2de08534-010f-4880-8cc4-59d4908b7666.png)
...simply copies..
★ 18i3yJjaXZ...

I understand we can get the address in other places.. but if a text field is copyable then it should be the full address.. (this is on Windows using wallet 1.3.0)

LeeAlephium:
@diomark would you expect all labels, when copied, to copy the full address?

diomark:
@LeeAlephium I think that'd be a good way to handle it. (when label is copied, copy full address)

LeeAlephium:
@mvaivre seems custom context menus are non-trivial. Are we ok with a little copy button popup beside badges?

![Untitled](https://user-images.githubusercontent.com/105806603/186211944-1ff77fbe-d89e-4c25-acb6-568fda968a28.png)

LeeAlephium:
Related https://github.com/alephium/desktop-wallet/issues/302

mvaivre:
We could do something like this:

<img width="782" alt="image" src="https://user-images.githubusercontent.com/6715943/186439468-6d1b862d-85c3-4ece-93dd-221baf6d57db.png">

Same for labels. For extra clarity, let's add a "Copy address" tooltip when hovering the copy icon.

BTW, motice the different "copy" icon, that we should update in the address detail view :)

Issue #302, opened 2022-07-18
Easy way to copy From or To address in transaction overview


LeeAlephium:
Right now you have to click on the transaction row and copy the address from the side panel

PR #344, opened 2022-08-26
Address badge labels copy full address

LeeAlephium:
Resolves https://github.com/alephium/desktop-wallet/issues/342, https://github.com/alephium/desktop-wallet/issues/334, https://github.com/alephium/desktop-wallet/issues/302


src/components/Buttons/ClipboardButton.tsx
```diff
@@ -92,3 +111,17 @@ export default styled(ClipboardButton)`
...
+  display: flex;
+  align-items: center;
+  overflow: hidden;
+
+  &:hover > ${CellClipboard} {
+    opacity: 1;
+  }
+
+  &:hover > ${CellChildren} {
+    -webkit-mask-image: linear-gradient(to right, rgba(0, 0, 0, 1) 50%, rgba(0, 0, 0, 0) calc(100% - 15px));
```
LeeAlephium, line 125: `CellClipboard` and `CellChildren` must be defined before these template interpolations, otherwise an error occurs, so they appear before the export.

mvaivre:
image

Could we change the tooltip text to "Copy address"? Otherwise we may think that the label text is going to be copied. Thanks!

mvaivre:
In the Address details view, this is now overflowing:

image

(Not sure it's linked to this PR, but it can be the occasion to fix it anyway :))

mvaivre:
image

We want to see the actual address instead of "address".

mvaivre:
Could you change the "copy icon" based on [this](https://github.com/alephium/desktop-wallet/issues/334#issuecomment-1225776605) comment? It seems to be more commonly used across today's apps. Thanks!

mvaivre:
image

The Addresses' table is broken, because the group isn't shown anymore.

LeeAlephium:
> The Addresses' table is broken, because the group isn't shown anymore.

It was actually the address hash missing :DD

Everything's fixed up!

LeeAlephium:
@mvaivre is there anything left I should address?

LeeAlephium:
Certain this also resolves https://github.com/alephium/desktop-wallet/issues/347

LeeAlephium:
@mvaivre rebase done

mvaivre:
Thanks for the fixes. 

One remaining issue (appart from the table alignment thingy mentioned in #347), the action icons in the address details view become smaller as the window width decreases:

image

LeeAlephium:
@mvaivre ready for more comments

mvaivre:
There are still some things itching my eyes a bit :P

Row text baseline isn't consistent :
image

Also, after taking a step back, I think that showing the main address "star icon" doesn't work that well visually and doesn't bring vital info to the table (we don't need to know which is the current main address in TX history tables..). Let's use a "dot" for all addresses.

mvaivre:
Also, could we align the "from / to" adresses and badge so that the spacing is consistent?

image


image

This would make the data easier to parse. Thanks!

mvaivre:
And last little thing while I'm at it. To respect the design evolution steps and keep consistency throughout versions, let's add a border to tables and remove the box shadow.

image

I'll work on an overall UI update taking inspiration from the mobile UI for 1.5.0. 

Thanks a lot! 🙏

LeeAlephium:
@mvaivre review

mvaivre:
Thanks, very nice.

2 little things :

1) Do you think that we could give a bit more space to addresses & labels if we have some?
image

Let's still set a max-width, but bigger than what it is today to show a bit more info on wide screens.
PS: Those columns will be way more useful with middle ellipsis (https://github.com/alephium/desktop-wallet/issues/350)

2) Could you show ellipsis here as well?
image

LeeAlephium:
> PS: Those columns will be way more useful with middle ellipsis

I'll do that then first and use it here

LeeAlephium:
@mvaivre review

I put the space you requested on Slack in the fix-address-cut branch

mvaivre:
Alrighty! I really like the middle ellipsis.

image

Could you try to get those two spaces (1 and 2) equivalent? If not possible, I'd rather have 1 being smaller, as I want to put more emphasis (ie. free space around) the amount. 

Also, as I said in my previous message :

> Let's still set a max-width, but bigger than what it is today to show a bit more info on wide screens.

Let's limit the addresses/labels width to around 200px :

image

LeeAlephium:
> Could you try to get those two spaces (1 and 2) equivalent? 

Just to be sure, then the other line with "Second" should have very large spaces on the sides (and cut off a large portion of the other label)?

LeeAlephium:
Yeah, looks quite nice to me:

src/components/Buttons/ClipboardButton.tsx
```diff
@@ -58,40 +64,55 @@ const ClipboardButton = ({ textToCopy, className }: ClipboardButtonProps) => {
...
+  )
+
+  return children || (Array.isArray(children) && children.length > 0) ? (
+    <div className={className}>
+      <CellChildren>{children}</CellChildren>
+      <CellClipboard>{clipboard}</CellClipboard>
+    </div>
+  ) : (
+    clipboard
+  )
```
nop33, line 99: In which case will `children` be falsy AND `Array.isArray(children) && children.length > 0` be true? In other words, when will the condition be met?

LeeAlephium: Never since `children.length > 0` excludes `[]` being a possibility

LeeAlephium: I'm also confused by the question, the code is "is children truthy OR (is children a non-empty list)"...

nop33: I changed it. Resolving.

nop33 (approved):
Code review and functional review is done, all good on my side. I would not merge it until @mvaivre also gives his approval.

Well done! 👏

mvaivre (approved):

Issue #350, opened 2022-09-06
Use middle ellipsis for hashes


mvaivre:
Let's use middle ellipsis for addresses and ids. Ideally, the length should be dynamic to adapt to any parent component.

<img width="134" alt="image" src="https://user-images.githubusercontent.com/6715943/188588393-0e6a9ac2-5702-4f81-adf9-7963f2ae4607.png">

mvaivre:
Depending on backend progress, this could eventually be shipped in 1.4.0.

LeeAlephium:
https://jsfiddle.net/cLweop1n/3/ I've made it as fast as I think is possible

nop33:
This can be closed, right?

PR #360, opened 2022-09-13
Add a component that creates middle ellipsis

LeeAlephium:
Supports https://github.com/alephium/desktop-wallet/issues/350 but will be resolved in https://github.com/alephium/desktop-wallet/pull/344

It seems to run very smoothly.

I achieve this by:

1. Having little to no re-allocations (useRef, useEffect)
2. Having little to no re-interpretations (hence functions declared outside of component)

I tried to get rid of useState and directly set `innerText` but React seems to have tight control of this. It was to avoid causing any React internals from computing, but it seems it's still fast regardless.

### DIFF
--- a/src/components/Ellipsed.tsx
+++ b/src/components/Ellipsed.tsx
@@ -1,0 +1,68 @@
+/*
+Copyright 2018 - 2022 The Alephium Authors
+This file is part of the alephium project.
+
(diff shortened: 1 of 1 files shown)


LeeAlephium:
Many changes were needed for integrating this component; I've instead included it in https://github.com/alephium/desktop-wallet/pull/344

Issue #358, opened 2022-09-12
Support setting lock time when sending a transaction


mvaivre:
Users should be able to set a lock time when sending some coins. 

**TODO:** 

- [x] Check address group requirements
- [x] Finish design

PR #361, opened 2022-09-15
Initial support for lock times in transactions

LeeAlephium:
Resolves https://github.com/alephium/desktop-wallet/issues/358


src/components/ToggleSection.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+import { motion } from 'framer-motion'
+import { ReactNode, useState } from 'react'
+import styled from 'styled-components'
+
+import Toggle from './Inputs/Toggle'
+
+interface ToggleSectionProps {
+  title: string
+  children: ReactNode | ReactNode[]
```
mvaivre, line 27: `ReactNode` encompasses the use of single of multiple children (no need for `ReactNode[]` afaik)

LeeAlephium: That's what I thought too when I originally started using this pattern (and that it sucked to have to do this lol) - but it seems I didn't look further, and that `ReactFragment` is the type in the `ReactNode` type alias that enables multiple children :D 

`type ReactFragment = {} | (boolean | any[] | ReactChild)[]`

I'll remove the list variants :+1:

src/components/Inputs/Select.tsx
```diff
@@ -160,11 +160,12 @@ function SelectOptionsPopup<T extends OptionValue>({
   onBackgroundClick: () => void
   title?: string
 }) {
-  const handleEvent = (el: HTMLSelectElement) =>
+  const handleEvent = (el: HTMLSelectElement) => {
     handleOptionSelect({
       label: options[el.selectedIndex]?.label,
       value: el.value as T
     })
+  }
```
nop33, line 168: I don't see why this change was necessary, but ok :)

nop33 (commented):
👍 

Spacing between form fields is larger than between the last form field and the toggle section 

image

src/components/Inputs/Select.tsx
LeeAlephium: I remember seeing this during rebase; yeah it isn't

nop33 (commented):
Last thing, could you also ensure the toggle section in the login page is extending to the full width (like the input fields)? In English it's shorter. Thanks!

image

locales/en-US/App.json
```diff
@@ -232,5 +232,7 @@
   "into": "into",
   "out from": "out from",
   "Transaction details": "Transaction details",
-  "Copy address": "Copy address"
+  "Copy address": "Copy address",
+  "Unlocks at": "Unlocks at",
+  "Unlocked at": "Unlocked at"
```
nop33, line 237: A couple more that we need in French :) @mvaivre 

I think that once your #371 is merged, we should include both translations in every PR that changes something in English.

LeeAlephium: As discussed, will happen in the other branch #371

LeeAlephium:
@nop33 is there anything else?

nop33 (approved):

PR #362, opened 2022-09-20
First translation (french) + language switch

mvaivre:


LeeAlephium:
Some tests are failing, but code-wise things look good

Only thing is an empty file: [locales/index.ts](https://github.com/alephium/desktop-wallet/pull/362/files#diff-36d4dd8c8f46a5d72aedc7e2f6bf23a09fb6b1ba2087ff15ccdb2b481c1b96f8)

LeeAlephium:
When first selecting a language, it's empty (when it's clear the default is English):



Switching between languages also causes a noticeable freeze - adding a loading spinner would be good to show (even though it won't spin, it'll at least signal there is some sort of loading happening...)

LeeAlephium (approved):
I see it's still draft but after communicating in a side-channel I think the PR is good

PR #353, opened 2022-09-06
Fix address card being cutoff

LeeAlephium:
Resolves https://github.com/alephium/desktop-wallet/issues/348

PR #364, opened 2022-09-22
Labels ilias

nop33:
Instead of commenting all these improvements/bugs to @LeeAlephium 's PR, I went ahead and implemented the things I noticed in my functional tests. 😇  I would appreciate some quite tests from your side since you have different transaction data than I do. I might have missed something.

https://user-images.githubusercontent.com/1579899/191739167-886b3bb3-8c1b-48c7-af54-b10ff0283f3a.mp4

After recording the video I noticed that the pending txs do not get displayed at all (no data). This was fixed as well.
image


LeeAlephium:
Quick comment

> [Bring back main address indication](https://github.com/alephium/desktop-wallet/pull/364/commits/b9f4f43fcdf2b3bdce9ffcb341ac9c638e3f9eca)

@mvaivre asked to remove

src/components/Ellipsed.tsx
```diff
@@ -57,10 +57,10 @@ const Ellipsed = ({ text, className }: EllipsedProps) => {
 
   const handleResize = createHandleResize(el, charWidth, text, setText)
 
-  useLayoutEffect(() => {
+  useEffect(() => {
```
LeeAlephium, line 60: Why this change?

src/modals/AddressOptionsModal.tsx
```diff
@@ -74,7 +72,7 @@ const AddressOptionsModal = ({ address, onClose }: AddressOptionsModal) => {
       : t`To remove this address from being the main address, you must set another one as main first.`
 
   return (
-    <div ref={divRef} tabIndex={0}>
```
LeeAlephium, line 77: Could you show me the broken scrolling?

This div is needed for accessibility purposes

src/components/AddressBadge.tsx
```diff
@@ -34,18 +35,19 @@ type AddressBadgeProps = ComponentPropsWithoutRef<typeof Badge> & {
...
 const AddressBadge = ({ address, showHashWhenNoLabel, className, ...props }: AddressBadgeProps) => {
   const { t } = useTranslation('App')
+  const { isPassphraseUsed } = useGlobalContext()
 
   if (!address) return null
 
   return showHashWhenNoLabel && !address.settings.label ? (
     <Hash className={className}>
-      {address.settings.isMain && '★'}
+      {!isPassphraseUsed && address.settings.isMain && '★'}
```
LeeAlephium, line 44: As mentioned main addresses no longer have a star

src/pages/Wallet/AddressesPage.tsx
```diff
@@ -107,7 +107,7 @@ const AddressesPage = () => {
               </ClipboardButton>
             </TableCell>
             <TableCell role="cell" tabIndex={0}>
-              <AddressBadge address={address} truncate />
+              {address.settings.label ? <AddressBadge address={address} truncate /> : '-'}
```
LeeAlephium, line 109: I would add this logic inside the `AddressBadge`

LeeAlephium:
At a glance it's no longer possible to tell if an address belongs to you or not - the dot was a nice indicator

Example:

src/pages/Wallet/AddressesPage.tsx
LeeAlephium: Hm, it's pretty specific to this context; all good

LeeAlephium:
The non-ellipsed / address-as-label is still shown in address summary cards:

LeeAlephium:
Nice job, just those small things and we'll merge it in :sparkles:

src/modals/AddressOptionsModal.tsx
LeeAlephium: It's possible it was addressed in https://github.com/alephium/desktop-wallet/pull/353

src/components/AddressBadge.tsx
nop33: Where was it mentioned, I am sorry if I missed it.

src/modals/AddressOptionsModal.tsx
nop33: It is shown in the video. The unexpected scrolling happens when opening the address options modal. Removing the div does not affect accessibility, I can still navigate with the Tab key in the modal. Am I missing sth?

src/components/AddressBadge.tsx
LeeAlephium: https://github.com/alephium/desktop-wallet/pull/364#issuecomment-1254982029

src/components/Ellipsed.tsx
nop33: To solve the issue where the hash of the selected address was not changing in the Send modal (AddressSelect) (demonstrated in the video). I had to put `text` into the depedencies of a `useEffect` and since the body of it was the same as the one in `useLayoutEffect`, I merged them into one `useEffect`. I read the docs for `useLayoutEffect` and it says to always prefer `useEffect` unless you really need to use `useLayoutEffect`. Since everything still works, I went for the simpler solution. Did I miss sth?

LeeAlephium: `useLayoutEffect` was purposefully used because the measurements needed for the ellipsis must occur after DOM updates - it may appear to work in this instance but the problem could manifest at another time - this happened to me while I was working on it

The dependency is ok :ok_hand:

nop33:
> At a glance it's no longer possible to tell if an address belongs to you or not - the dot was a nice indicator

I am of the opinion that there shouldn't be a distinction between not-labeled addresses and any other external address. If you want a distinction, add a label. I feel like it complicates the UI.

@mvaivre has some ideas on adding a "tag" icon on hover to encourage users to add labels to their own addresses. I propose that we re-iterate on this later, I am personally satisfied with the current state of the UI.

nop33:
> The non-ellipsed / address-as-label is still shown in address summary cards:

That's fine by me, it's on purpose!

LeeAlephium:
Okie dokie :+1: 

I noticed there are several places where address badges are missing the new borders (Addresses screen, address selecting when sending, as examples)

nop33:
> I noticed there are several places where address badges are missing the new borders (Addresses screen, address selecting when sending, as examples)

That's also intentional. I'll leave it up to @mvaivre to fine-tune the rest, for me it looks great as it is now!

src/components/AddressBadge.tsx
nop33: Now that the "dot" is moved on the right side, I think it's ok to have the star back!

src/modals/AddressOptionsModal.tsx
nop33: The focus moves on the title of the modal, is that not enough?

LeeAlephium: Ah yeah, it is - forgot `CenteredModal` takes care of this now

mvaivre:
Ah, I was going to do some functional tests (and possible tweaks) as soon as I'm done with my current task. Not a big deal, I'll open another PR if necessary.

PR #365, opened 2022-09-22
Lock time ilias

nop33:
1. I changed the bg of the toggle section to match the designs. The light theme however doesn't look nice and needs adjustment.
2. Even though I added a min date for the datepicker field, it seems that I can still select a past hour (date works fine)

### DIFF
--- a/src/components/Inputs/AmountInput.tsx
+++ b/src/components/Inputs/AmountInput.tsx
@@ -81,22 +81,32 @@ const AmountInput = ({ className, availableAmount, ...props }: AmountInputProps)
         }
         {...restProps}
         error={error}
-      />
-      {availableAmount && (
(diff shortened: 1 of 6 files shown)


LeeAlephium:
Code :+1: 

Just one thing: in a side-channel discussion with @mvaivre we agreed it makes more sense to show local time, not UTC time. UTC time is just used internally - if you could change that back it'd be great :ok_hand: 

> Even though I added a min date for the datepicker field, it seems that I can still select a past hour (date works fine)

Let's log an issue https://github.com/alephium/desktop-wallet/issues/366

LeeAlephium:
You can merge after that small fix

LeeAlephium (approved):

PR #372, opened 2022-09-26
Update @alephium/sdk to support network upgrades

LeeAlephium:
Based on lock-time since there are changes to the APIs regarding this too

### DIFF
--- a/src/components/IOList.tsx
+++ b/src/components/IOList.tsx
@@ -51,7 +51,10 @@ const IOList = ({ currentAddress, isOut, outputs, inputs, timestamp, linkToExplo
       .map((v) => v.address)
       .uniq()
       .value()
+
     const addressHash = isAllCurrentAddress ? currentAddress : notCurrentAddresses[0]
+    if (!addressHash) return null
+
     const extraAddressesText = notCurrentAddresses.length > 1 ? `(+${notCurrentAddresses.length - 1})` : ''
 
     // There may be a case where a wallet sends funds to the same address, which doesn't
@@ -60,7 +63,6 @@ const IOList = ({ currentAddress, isOut, outputs, inputs, timestamp, linkToExplo
 
     const getAddressComponent = (addressHash: AddressHash) => {
       const address = getAddress(addressHash)
-      if (!address) return null
 
       return address ? (
         <AddressBadge truncate address={address} showHashWhenNoLabel withBorders />
@@ -69,8 +71,6 @@ const IOList = ({ currentAddress, isOut, outputs, inputs, timestamp, linkToExplo
       )
     }
 
-    if (!addressHash) return null
-
     return truncate ? (
       <TruncateWrap>
         {getAddressComponent(addressHash)}
@@ -78,10 +78,10 @@ const IOList = ({ currentAddress, isOut, outputs, inputs, timestamp, linkToExplo
       </TruncateWrap>
     ) : (
       <Addresses>
-        {addressesToShow.map((address) => {
-          if (!address) return null
+        {addressesToShow.map((addressHash) => {
+          if (!addressHash) return null
 
-          const addressComponent = getAddressComponent(address)
+          const addressComponent = getAddressComponent(addressHash)
           return linkToExplorer ? (
             <ActionLinkStyled onClick={() => handleShowAddress(addressHash)} key={addressHash}>
               {addressComponent}
--- a/src/components/IOList.tsx
+++ b/src/components/IOList.tsx
@@ -60,6 +60,7 @@ const IOList = ({ currentAddress, isOut, outputs, inputs, timestamp, linkToExplo
 
     const getAddressComponent = (addressHash: AddressHash) => {
       const address = getAddress(addressHash)
+      if (!address) return null
 
       return address ? (
         <AddressBadge truncate address={address} showHashWhenNoLabel withBorders />
@@ -68,6 +69,8 @@ const IOList = ({ currentAddress, isOut, outputs, inputs, timestamp, linkToExplo
       )
     }
 
+    if (!addressHash) return null
+
     return truncate ? (
       <TruncateWrap>
         {getAddressComponent(addressHash)}
@@ -75,9 +78,10 @@ const IOList = ({ currentAddress, isOut, outputs, inputs, timestamp, linkToExplo
       </TruncateWrap>
     ) : (
       <Addresses>
-        {addressesToShow.map((addressHash) => {
-          const addressComponent = getAddressComponent(addressHash)
+        {addressesToShow.map((address) => {
+          if (!address) return null
 
+          const addressComponent = getAddressComponent(address)
           return linkToExplorer ? (
             <ActionLinkStyled onClick={() => handleShowAddress(addressHash)} key={addressHash}>
               {addressComponent}
--- a/src/components/TransactionalInfo.tsx
+++ b/src/components/TransactionalInfo.tsx
@@ -18,6 +18,7 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
 
 import { calAmountDelta, formatAmountForDisplay } from '@alephium/sdk'
 import { Output, Transaction } from '@alephium/sdk/api/explorer'
+import { AssetOutput, Output, Transaction } from '@alephium/sdk/api/explorer'
 import { colord } from 'colord'
 import { ArrowRight as ArrowRightIcon } from 'lucide-react'
 import { useTranslation } from 'react-i18next'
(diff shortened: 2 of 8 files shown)


nop33 (commented):
App doesn't work

```
Uncaught TypeError: Cannot convert undefined to a BigInt
```

src/components/IOList.tsx
```diff
@@ -60,6 +60,7 @@ const IOList = ({ currentAddress, isOut, outputs, inputs, timestamp, linkToExplo
 
     const getAddressComponent = (addressHash: AddressHash) => {
       const address = getAddress(addressHash)
+      if (!address) return null
```
nop33, line 63: That means that the following condition will never evaluate to display the component

```ts
<AddressEllipsed addressHash={addressHash} />
```

...

src/components/IOList.tsx
```diff
@@ -68,16 +69,19 @@ const IOList = ({ currentAddress, isOut, outputs, inputs, timestamp, linkToExplo
...
         {extraAddressesText && <AddressesHidden>{extraAddressesText}</AddressesHidden>}
       </TruncateWrap>
     ) : (
       <Addresses>
-        {addressesToShow.map((addressHash) => {
-          const addressComponent = getAddressComponent(addressHash)
+        {addressesToShow.map((address) => {
+          if (!address) return null
 
+          const addressComponent = getAddressComponent(address)
```
nop33, line 84: Why the parameter rename? I find it clearer the way it was before.

src/components/IOList.tsx
nop33: If the hash is not one of the internal addresses, your change returns `null` instead of the ellipsed hash.

LeeAlephium:
@nop33 rebased, and fixed those other 2 comments

src/components/TransactionalInfo.tsx
```diff
@@ -70,6 +70,10 @@ const TransactionalInfo = ({ transaction: tx, addressHash, className, hideLeftAd
   let pendingToAddressComponent
   let lockTime: Date | undefined
 
+  const token = 'alph'
```
nop33, line 73: Why moving it up again? Was there an issue with having it close to the render function? It serves no purpose to have it so high up, it's only used in the render function.

src/modals/TransactionDetailsModal.tsx
```diff
@@ -153,16 +153,16 @@ const TransactionDetailsModal = ({ transaction, address, onClose }: TransactionD
           <DetailsRow label={t`Inputs`}>
             <AddressList>
               {transaction.inputs?.map((input) => (
-                <ActionLinkStyled key={`${input.outputRef.key}`} onClick={() => handleShowAddress(input.address)}>
-                  <AddressEllipsed key={`${input.outputRef.key}`} addressHash={input.address} />
+                <ActionLinkStyled key={`${input.outputRef.key}`} onClick={() => handleShowAddress(input.address ?? '')}>
+                  <AddressEllipsed key={`${input.outputRef.key}`} addressHash={input.address ?? ''} />
```
nop33, line 157: Does it really make sense to render `ActionLinkStyled` and `AddressEllipsed` if `input.address` is `undefined? Instead of passing an empty string to a function and component that both expect an address hash, I would just not render the components at all.

```ts
              {transaction.inputs?.map(
                (input) =>
                  input.address && (
                    <ActionLinkStyled
                      key={`${input.outputRef.key}`}
                      onClick={() => handleShowAddress(input.address as string)}
                    >
                      <AddressEllipsed key={`${input.outputRef.key}`} addressHash={input.address} />
                    </ActionLinkStyled>
                  )
              )}
```

src/modals/TransactionDetailsModal.tsx
```diff
@@ -153,16 +153,16 @@ const TransactionDetailsModal = ({ transaction, address, onClose }: TransactionD
...
                 </ActionLinkStyled>
               ))}
             </AddressList>
           </DetailsRow>
           <DetailsRow label={t`Outputs`}>
             <AddressList>
               {transaction.outputs?.map((output) => (
-                <ActionLinkStyled key={`${output.key}`} onClick={() => handleShowAddress(output.address)}>
+                <ActionLinkStyled key={`${output.key}`} onClick={() => handleShowAddress(output.address ?? '')}>
```
nop33, line 165: This change doesn't seem to be necessary

src/utils/transactions.ts
```diff
@@ -86,10 +86,10 @@ export function sortTransactions(a: HasTimestamp, b: HasTimestamp): number {
 }
 
 export const hasOnlyInputsWith = (inputs: Input[], addresses: Address[]): boolean =>
-  inputs.every((i) => addresses.map((a) => a.hash).indexOf(i.address) >= 0)
+  inputs.every((i) => addresses.map((a) => a.hash).indexOf(i?.address ?? '') >= 0)
```
nop33, line 89: This looks cleaner to me

```suggestion
  inputs.every((i) => i?.address && addresses.map((a) => a.hash).indexOf(i.address) >= 0)
```

nop33: Same for below

src/components/TransactionalInfo.tsx
LeeAlephium: My reason is in the future this will be replaced with proper code which I think will live up there - otherwise, there's no reason and is highly preferential

src/modals/TransactionDetailsModal.tsx
LeeAlephium: They never will be undefined, but the type-checker was not happy, so this satisfied it - sure, not rendering them at all is an option too

src/modals/TransactionDetailsModal.tsx
LeeAlephium: Same reason as input?.address

src/modals/TransactionDetailsModal.tsx
LeeAlephium: It seems the type-checker is really unhappy with a wrapping `input.address !== undefined && ...` ... It's been resolved using both ways

src/modals/TransactionDetailsModal.tsx
nop33: I think you're wrong, `output.address` is always a `string`.

src/modals/TransactionDetailsModal.tsx
```diff
@@ -152,20 +152,29 @@ const TransactionDetailsModal = ({ transaction, address, onClose }: TransactionD
...
+                (input) =>
+                  input.address !== undefined && (
+                    <ActionLinkStyled
+                      key={`${input.outputRef.key}`}
+                      onClick={() => handleShowAddress(input.address ?? '')}
+                    >
+                      <AddressEllipsed key={`${input.outputRef.key}`} addressHash={input.address} />
+                    </ActionLinkStyled>
+                  )
+              )}
```
nop33, line 165: `input.address` is of type `string | undefined`. We don't specifically have to check for `!== undefined`. We also don't need to pass an empty string to a function that expects an addressHash (`?? ''`). The compiler does not complain with what I proposed earlier:


```ts
{transaction.inputs?.map(
                (input) =>
                  input.address && (
                    <ActionLinkStyled
                      key={`${input.outputRef.key}`}
                      onClick={() => handleShowAddress(input.address as string)}
                    >
                      <AddressEllipsed key={`${input.outputRef.key}`} addressHash={input.address} />
                    </ActionLinkStyled>
                  )
              )}
```

src/modals/TransactionDetailsModal.tsx
```diff
@@ -152,20 +152,29 @@ const TransactionDetailsModal = ({ transaction, address, onClose }: TransactionD
...
-                </ActionLinkStyled>
-              ))}
+              {transaction.outputs?.map(
+                (output) =>
+                  output.address && (
+                    <ActionLinkStyled key={`${output.key}`} onClick={() => handleShowAddress(output.address)}>
+                      <AddressEllipsed key={`${output.key}`} addressHash={output.address} />
+                    </ActionLinkStyled>
+                  )
+              )}
```
nop33, line 177: There is no need to check for `output.address`. It's never `undefined`. The following is enough:

```ts
{transaction.outputs?.map((output) => (
                <ActionLinkStyled key={`${output.key}`} onClick={() => handleShowAddress(output.address)}>
                  <AddressEllipsed key={`${output.key}`} addressHash={output.address} />
                </ActionLinkStyled>
              ))}
```

src/components/TransactionalInfo.tsx
nop33: sounds good 👍

src/modals/TransactionDetailsModal.tsx
LeeAlephium: The issue is `output` can be undefined - I'll do the same suggestion as with `input`

LeeAlephium:
Everything should be good now; I had missed "as string" in your suggestion and it caused the trouble

nop33 (approved):
👍 👏

PR #374, opened 2022-09-26
Detect and display pending TX found in mempool

LeeAlephium:
Requires new API, so based on update-schemas


LeeAlephium:
Due to https://github.com/alephium/explorer-backend/issues/360, incoming pending won't work, but pending outgoing will, and eventually when that issue is resolved incoming should "just work".

nop33 (commented):
Thanks for this PR Lee 🙏 

I've created #377 to address some of the issues I found in this PR.

nop33:
I rebased since the base branch had diversed.

src/hooks/useTransactionInfo.tsx
```diff
@@ -0,0 +1,115 @@
...
+    lockTime = lockTime.toISOString() === new Date(0).toISOString() ? undefined : lockTime
+  }
+
+  return {
+    amount,
+    direction,
+    infoType,
+    outputs,
+    lockTime,
+    label: {
```
mvaivre, line 78: Mhh, I'm not sure that I like adding static data to a hook.
For me, hook's data should depend on a state. 

I'd create a new util method `getTransactionUI(type: 'in' | 'out' | 'move' | 'pending)` in  `utils/transactions` which return the needed labels, colors and components... WDYT?

nop33: ok! Yeah, I also wansn't 100% satisfied, but I put it in the hook to reduce lines of code. Happy to move it out.

nop33: It has to be a hook however, because it's using `t` and `useTheme`

mvaivre: Okay yeah. Or we add `t` and `theme` as required arguments...

nop33: Fixed!

nop33: > Or we add t and theme as required arguments

I prefer the hook option, makes the code cleaner imo.

mvaivre (approved):

PR #377, opened 2022-09-28
Fixes and improvements on pending tx

nop33:
This PR fixes:
1. The scroll issue in the home page (see https://github.com/alephium/desktop-wallet/pull/377#discussion_r982459318)

This PR improves:
1. The display of consolidation txs. When sending from address X to address X, the tx tables show nothing as value, and the tx details modal shows only the fee as the sent value, which is confusing. 

|before | after |
| ----- | ----- |
| image  | image|
| image | image |

### DIFF
--- a/locales/fr-FR/App.json
+++ b/locales/fr-FR/App.json
@@ -213,6 +213,7 @@
   "(current)": "(actif)",
   "Default address for sending transactions.": "Adresse par défaut pour effectuer des transactions.",
   "Error while fetching data for address {{ hash }}": "Erreur lors de la récupération des données de l'adresse {{ hash }}",
+  "Error while fetching pending transactions for address {{ hash }}": "Error while fetching pending transactions for address {{ hash }}",
   "Open in explorer": "Ouvrir dans l'explorer",
   "Reduce": "Réduire",
   "Show addresses": "Montrer les adresses",
--- a/locales/fr-FR/App.json
+++ b/locales/fr-FR/App.json
@@ -213,7 +213,7 @@
   "(current)": "(actif)",
   "Default address for sending transactions.": "Adresse par défaut pour effectuer des transactions.",
   "Error while fetching data for address {{ hash }}": "Erreur lors de la récupération des données de l'adresse {{ hash }}",
-  "Error while fetching pending transactions for address {{ hash }}": "Error while fetching pending transactions for address {{ hash }}",
+  "Error while fetching pending transactions for address {{ hash }}": "Erreur durant la récupération de transactions en attente pour l'adresse {{ hash }}",
   "Open in explorer": "Ouvrir dans l'explorer",
   "Reduce": "Réduire",
   "Show addresses": "Montrer les adresses",
--- a/src/components/OverviewPage/TransactionList.tsx
+++ b/src/components/OverviewPage/TransactionList.tsx
@@ -61,14 +61,11 @@ const OverviewPageTransactionList = ({ className, onTransactionClick }: Overview
 
   return (
     <Table isLoading={showSkeletonLoading} className={className} minWidth="500px">
-      {allPendingTxs
-        .slice(0)
-        .reverse()
-        .map(({ data: tx, address }: BelongingToAddress<PendingTx>) => (
-          <TableRow key={tx.txId} blinking role="row" tabIndex={0}>
-            <TransactionalInfo transaction={tx} addressHash={address.hash} />
-          </TableRow>
-        ))}
+      {allPendingTxs.map(({ data: tx, address }: BelongingToAddress<PendingTx>) => (
+        <TableRow key={tx.txId} blinking role="row" tabIndex={0}>
+          <TransactionalInfo transaction={tx} addressHash={address.hash} />
+        </TableRow>
+      ))}
       {allConfirmedTxs.map(({ data: tx, address }: BelongingToAddress<Transaction>) => {
         if (hasOnlyInputsWith((tx as Transaction).inputs ?? [], addresses) && getDirection(tx, address.hash) == 'in')
           return null
--- a/src/contexts/addresses.tsx
+++ b/src/contexts/addresses.tsx
@@ -32,7 +32,7 @@ import { PartialDeep } from 'type-fest'
 import { TimeInMs } from '../types/numbers'
 import { AddressSettings, loadStoredAddressesMetadataOfWallet, storeAddressMetadataOfWallet } from '../utils/addresses'
 import { NetworkName } from '../utils/settings'
-import { fromUnconfirmedTransactionToPendingTx, TransactionType } from '../utils/transactions'
+import { convertUnconfirmedTxToPendingTx, TransactionType } from '../utils/transactions'
 import { useGlobalContext } from './global'
 
 export type PendingTx = {
@@ -253,7 +253,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
               txs.forEach((tx) => {
                 if (tx.type !== 'Unconfirmed') return
                 if (address.transactions.pending.some((t: PendingTx) => t.txId == tx.hash)) return
-                address.addPendingTransaction(fromUnconfirmedTransactionToPendingTx(tx, address.hash, currentNetwork))
+                address.addPendingTransaction(convertUnconfirmedTxToPendingTx(tx, address.hash, currentNetwork))
               })
             })
           ),
--- a/src/contexts/addresses.tsx
+++ b/src/contexts/addresses.tsx
@@ -243,24 +243,35 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
         })
         return
       }
+      setIsLoadingData(true)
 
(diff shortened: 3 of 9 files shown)


locales/fr-FR/App.json
```diff
@@ -213,6 +213,7 @@
   "(current)": "(actif)",
   "Default address for sending transactions.": "Adresse par défaut pour effectuer des transactions.",
   "Error while fetching data for address {{ hash }}": "Erreur lors de la récupération des données de l'adresse {{ hash }}",
+  "Error while fetching pending transactions for address {{ hash }}": "Error while fetching pending transactions for address {{ hash }}",
```
nop33, line 216: @mvaivre I could use your help translating this :)

src/components/OverviewPage/TransactionList.tsx
```diff
@@ -61,14 +61,11 @@ const OverviewPageTransactionList = ({ className, onTransactionClick }: Overview
 
   return (
     <Table isLoading={showSkeletonLoading} className={className} minWidth="500px">
-      {allPendingTxs
-        .slice(0)
-        .reverse()
```
nop33, line 66: I don't know why this was like this, pending addresses should be displayed like any other.

src/contexts/addresses.tsx
```diff
@@ -171,13 +171,10 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
     (addressState) => addressState.network === currentNetwork
   )
 
-  // Used for showing flashing on the large summary card
```
nop33, line 174: This is an inaccurate comment. `addressesWithPendingSentTxs` is used for multiple reasons. It's used for defining the receiving addresses with pending transactions so that their info gets refreshed and also used in the calculation of whether data is loading or not.

src/contexts/addresses.tsx
```diff
@@ -233,44 +228,57 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
...
-              const { data: txs } = response
-              txs.forEach((tx) => {
-                if (tx.type !== 'Unconfirmed') return
-                if (address.transactions.pending.some((t: PendingTx) => t.txId == tx.hash)) return
-                address.addPendingTransaction(fromUnconfirmedTransactionToPendingTx(tx, address.hash, currentNetwork))
-              })
-            })
-          ),
-        Promise.resolve()
-      )
```
nop33, line 261: Looking at the devtools, the requests were not made in a sequence (one after the other resolves), but all together. The updated logic is similar to how `fetchAndStoreAddressesData` is implemented.

src/contexts/addresses.tsx
```diff
@@ -432,10 +439,6 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
   useEffect(() => {
     // In case the "to" address of a pending transaction is an address of this wallet, we need to query the API for this
     // one as well
-    const addressesWithPendingSentTxs = addressesOfCurrentNetwork.filter(
-      (address) => address.transactions.pending.filter((pendingTx) => pendingTx.network === currentNetwork).length > 0
-    )
-
```
nop33, line 438: There's no point having this here, it already exists outside of this `useEffect`. It will evaluate to the same value.

src/contexts/addresses.tsx
```diff
@@ -456,20 +459,18 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
...
-  useEffect(() => {
-    const interval = setInterval(async () => {
-      await fetchPendingTxs(addressesOfCurrentNetwork)
-      setCheckForNewPending({})
-      // MinimumWaitTime + HalfAvgBlockTime
-    }, 1000 + (1000 * 60) / 2)
-    return () => {
-      clearInterval(interval)
-    }
-  }, [addressesOfCurrentNetwork, fetchPendingTxs])
```
nop33, line 472: Since we are not checking for new confirmed transactions unless the user hits the "Refresh" button, I don't see a reason to do that with unconfirmed txs. I have updated the refresh function to also check for unconfirmed.

src/utils/transactions.ts
```diff
@@ -99,35 +99,27 @@ export const getDirection = (tx: Transaction, address: AddressHash): Transaction
...
-
-  for (let index = (tx.outputs ?? []).length - 1; index > 0; index--) {
-    const output = (tx.outputs ?? [])[index]
-    const inputIndex = inputs.indexOf(output.address)
-    if (inputs[inputIndex]) {
-      change += BigInt(output.attoAlphAmount)
-      inputs.splice(inputIndex, 1)
-    }
-    if (inputs.length == 0) return total - change
-  }
```
nop33, line 117: I find this calculation very complex to understand and reason about. I know that my solution might be slightly more computationally heavy, but is very simple.

src/utils/transactions.ts
```diff
@@ -99,35 +99,27 @@ export const getDirection = (tx: Transaction, address: AddressHash): Transaction
...
-export const fromUnconfirmedTransactionToPendingTx = (
+export const convertUnconfirmedTxToPendingTx = (
   tx: UnconfirmedTransaction,
   belongingTo: AddressHash,
   network: NetworkName
 ): PendingTx => {
-  let amount = calculateTotalTxOutput(tx as unknown as Transaction)
-  const amountIsBigInt = typeof amount === 'bigint'
-  const type = amount && amountIsBigInt && amount < 0 ? 'out' : 'in'
-  amount = amount && (type === 'out' ? amount * BigInt(-1) : amount)
```
nop33, line 130: As I mentioned in the other PR, `amount` will always be a `bigint` here. My changes reflect that.

locales/fr-FR/App.json
mvaivre: "Erreur durant la récupération de transactions en attente pour l'adresse {{ hash }}"

src/contexts/addresses.tsx
LeeAlephium: How is that, when it's a chain of `.then()` ? (Haven't verified but it seemed sequential to me)

src/pages/HomePage.tsx
```diff
@@ -53,40 +52,38 @@ const HomePage = () => {
...
-                <InitialActions showLinkToExistingWallets onLinkClick={hideInitialActions} />
-              </>
-            )}
-          </FloatingPanel>
-        </InteractionArea>
-        {isUpdateWalletModalVisible && (
-          <UpdateWalletModal newVersion={newLatestVersion} onClose={() => setUpdateWalletModalVisible(false)} />
-        )}
-      </HomeContainer>
-    </Scrollbar>
```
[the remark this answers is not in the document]
LeeAlephium: On the home screen there was just no scrollbar at all.

src/utils/transactions.ts
LeeAlephium: It is an inherently complex calculation. We will need tests to make sure your alternative is equivalent

src/contexts/addresses.tsx
nop33: It didn't look sequential to me when I tested.

In any case, I think that the change I made makes the code similar to existing code (using await syntax)

src/contexts/addresses.tsx
nop33: But they should give the same result.

src/pages/HomePage.tsx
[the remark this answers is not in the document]
nop33: Without your change, it seems that everything works fine on my side. With your change there are 2 issues introduced:
1. I cannot see the top of the screen, when the window is very small
2. The bottom left side is empty

In the demo I first show your branch, and then I show how it works with this changed deleted (as in my branch)


https://user-images.githubusercontent.com/1579899/192801496-554e3fcd-f6f2-407a-81d6-1a2707948f29.mp4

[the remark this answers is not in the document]
LeeAlephium: :thinking: huh, okie dokie then! I'd be more than glad for us to revert the change

src/contexts/addresses.tsx
LeeAlephium: If you've found this to be the case then ok - I did this under the assumption it's possible (and it is possible) that the statement inside the useEffect can evaluate to something else because of the nature of when useEffect fires.

src/contexts/addresses.tsx
LeeAlephium: Doesn't matter much to me which way we encode the "sequencing" of the promises - I do it the other way because it's a common idiom (plus I find the Promise API superior in pretty much every way to async/await, but that's another conversation to have elsewhere :p)

As long as the new code is equiv. all good :+1:

src/pages/HomePage.tsx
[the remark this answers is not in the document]
nop33: This PR does that, so no need :) We just need to merge it!

LeeAlephium:
No, it removes the original code, the tests should belong here first to verify the code written is equivalent and it costs nothing to have a PR open

nop33:
There were no tests in the original branch either.

mvaivre (approved):

PR #380, opened 2022-09-28
Fixes and improvements on pending amount calculations and amounts display

nop33:
This PR:
1. Reuses the sdk function `calAmountDelta` (which already has tests for) and only uses the new function `calculateUnconfirmedTxSentAmount` for unconfirmed transactions
2. Reuses new transaction info colors in the transaction details modal
1. Ensures that the calculation of the unconfirmed tx sent amount is only taking into account the specified address
1. Cleans up and fixes logic in unconfirmed transaction to pending transaction conversion function and adds explanation comments 
1. Improves the display of transaction details for "moved" funds (see screenshots below)
2. 🆕  Makes code DRYer with use of hooks
3. 🆕  Simplifies identification of pending txs
4. 🆕 Shows "moved" also in the address detail page (see screenshots below)

| before | after |
| ----- | ----- |
| image | image |
| image | image|

Reason why the amount of unconfirmed transactions cannot be calculated the same way as confirmed transactions:
> Thomas said:
> unfortunatelly Input.attoAlphAmount isn't available right away on an Input .we need to find the corresponding Output in order to get the amount. We do this on a Confirmed tx, but we can't afford doing this (for now) on the unconfirmed tx, as it's an heavy task it would be too costly while the tx is in the mempool for very little time. We only display the max we have, but what you could do is to offer a link to the output, using outputRef.key and /transaction-by-output-ref-ke/<outputRef-key>  endoint.

Feedback is welcomed!

~~PS: Some cleanup might be needed.~~ Added some 🆕 bullet points to cover this 😄


LeeAlephium (approved):

src/utils/transactions.ts
```diff
@@ -93,45 +93,42 @@ export const hasOnlyInputsWith = (inputs: Input[], addresses: Address[]): boolea
...
-    (acc, output) => (inputAddresses.includes(output.address) ? acc + BigInt(output.attoAlphAmount) : acc),
+  if (isConsolidationTx(tx)) return totalOutputAmount
+
+  const totalOutputAmountOfAddress = tx.outputs.reduce(
+    (acc, output) => (output.address === address ? acc + BigInt(output.attoAlphAmount) : acc),
     BigInt(0)
   )
 
-  return total - change
+  return totalOutputAmount - totalOutputAmountOfAddress
```
nop33, line 111: So, basically, since we cannot know the amounts of the input UTXOs, we cannot calculate the delta by substracting the sum of inputs from the sum of outputs of the specified address. Second best thing we can do is to substract the total outputs of the address (aka, change) from the total tx output amount.

This is pretty much what you did as well @LeeAlephium, right (reference to [Lee's code](https://github.com/alephium/desktop-wallet/pull/374/commits/f514957b15a9b9cfe906b16a56403703d5944ff5))? Please, let me know if you think this logic is missing something.

I believe that both of our approaches will produce wrong results in the special case of a transaction where there are more than 1 unique input address.

src/utils/transactions.ts
```diff
@@ -93,45 +93,42 @@ export const hasOnlyInputsWith = (inputs: Input[], addresses: Address[]): boolea
...
+  fromAddress: AddressHash,
   network: NetworkName
 ): PendingTx => {
-  let amount = calculateAmountSent(tx)
-  const type = amount < 0 ? 'out' : 'in'
-  amount = type === 'out' ? amount * BigInt(-1) : amount
+  if (!tx.outputs) throw 'Missing transaction details'
 
-  const fromAddress = type === 'out' ? belongingTo : ((tx.inputs ?? [])[0] ?? {}).address
-  const toAddress = type === 'in' ? ((tx.outputs ?? [])[0] ?? {}).address : belongingTo
```
nop33, line 134: In the base branch, as well as in this branch, the `amount` will always be a positive number. Moreover, we know (due to the issue I mentioned in the comment) that we can only fetch **sending** unconfirmed txs, not receiving nes. Thus, I simplified the code.

src/utils/transactions.ts
LeeAlephium: If it appears to work then I guess it works :shrug:

src/components/TransactionalInfo.tsx
```diff
@@ -76,10 +75,17 @@ const TransactionalInfo = ({ transaction: tx, addressHash, className, hideLeftAd
...
+      amount = tx.outputs.reduce((acc, output) => acc + BigInt(output.attoAlphAmount), BigInt(0))
+      direction = 'out'
+      infoType = 'move'
+    } else {
+      amount = calAmountDelta(tx, _addressHash)
+      direction = amount < 0 ? 'out' : 'in'
+      amount = amount < 0 ? amount * BigInt(-1) : amount
+      infoType =
+        !hideLeftAddress && direction === 'out' && hasOnlyOutputsWith(tx.outputs ?? [], addresses) ? 'move' : direction
+    }
```
mvaivre, line 88: Could we extract this as a function? It's basically reused in `src/modals/TransactionDetailsModal.tsx`

nop33: Already done!

LeeAlephium (approved):

mvaivre (approved):

Issue #390, opened 2022-10-03
Password check modal expands to the whole height of the screen


nop33:
<img width="1191" alt="image" src="https://user-images.githubusercontent.com/1579899/193612732-8753b55d-b7f2-4780-bebc-056662dd1b1c.png">

nop33:
@LeeAlephium would u like to work on this? It seems to be related to the custom scrollbars, I thought you might be able to debug it faster than me :)

nop33:
Fixed by #392

PR #392, opened 2022-10-03
Dont use the same centered modal for the password confirmation

LeeAlephium:
Resolves https://github.com/alephium/desktop-wallet/issues/390

### DIFF
--- a/src/modals/SendModal/index.tsx
+++ b/src/modals/SendModal/index.tsx
@@ -226,6 +226,7 @@ const SendModal = ({ onClose }: SendModalProps) => {
   return (
     <>
       <CenteredModal
+        key={step}
         title={title}
         onClose={onClose}
         isLoading={isLoading}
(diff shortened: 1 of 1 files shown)


nop33:
@LeeAlephium thanks for the quick fix! I noticed that the same problem occurs on the second screen (`SendModalCheckTransaction`) , it's just less visible because the heights of the first and second screen are similar.

nop33:
@LeeAlephium so maybe it'd be better for every step to have its own `CenteredModal`? I guess we can keep the same component logic, except the render method that can be modified to reflect something like that:
```
{ step === 1 && <CenteredModal ...><SendModalTransactionForm ...>...</SendModalTransactionForm></CenteredModal>
{ step === 2 && <CenteredModal ...><SendModalCheckTransaction ...>...</SendModalCheckTransaction></CenteredModal>
{ step === 3 && <CenteredModal ...><PasswordConfirmation ...>...</PasswordConfirmation></CenteredModal>
```

WDYT?

nop33:
@LeeAlephium will you update the PR to remove the first commit since it's not needed?

LeeAlephium:
The first commit also resolves a bug where the user closes the password modal and it closes and resets the whole Send process - not sure if that's intentional, if feels wrong for it to happen

nop33 (approved):

PR #400, opened 2022-10-04
Pre-release fixes

nop33:


src/utils/transactions.ts
```diff
@@ -99,6 +100,13 @@ export const isConsolidationTx = (tx: Transaction | UnconfirmedTransaction): boo
   return inputAddresses.length === 1 && outputAddresses.length === 1 && inputAddresses[0] === outputAddresses[0]
 }
 
+export const removeConsolidationChangeAmount = (totalOutputAmount: bigint, outputs: AssetOutput[] | Output[]) =>
+  outputs.length > 1
+    ? // If there are multiple outputs, the last one must be the change amount
```
LeeAlephium, line 105: The last one is not guaranteed to be the last one, it's just a heuristic. Might be good to change this comment for your future self.

LeeAlephium (approved):

src/utils/transactions.ts
nop33: Thanks @LeeAlephium !

PR #405, opened 2022-10-06
Remove gradient to reduce CPU usage

### LINKED ISSUE #404: Reduce CPU usage
On our Discord it has been reported that our app uses too much CPU:

image

This is what I get on my M1 MacBook.

image

PS: I am getting very high values once I log in. I suspect that it might be the background animation that causes this CPU usage. Every other page except the Overview page shows normal CPU usage levels.

nop33:
Closes #404

@mvaivre I know this is something you really like and are really proud of, that's why I'll leave this as an open PR until you're back from holidays. Check the issue that this refers to.


nop33:
I received confirmation through Slack :)

Issue #141, opened 2022-03-03
Automatic update system


mvaivre:
If we want our users to always stay up to date, we could implement auto-updating. 
I've already deployed that for another electron project in the past, and it's quite straightforward : the app calls a file server where the latest binaries are stored, it checks if the version number is higher, if yes, it downloads and allows the user to install the update in a click, automagically. 

@polarker WDYT?

nop33:
https://www.electronjs.org/docs/latest/tutorial/updates

polarker:
In general, I feel it's risky to update automatically for sensitive applications

nop33:
@polarker what are the security implications that you are referring to?

@mvaivre which crypto wallet desktop apps do you recommend checking? I remember you get inspired by the Ledger Live app a lot. Is there any other?

mvaivre:
@nop33 My main 2 daily drivers are indeed Ledger Live and [Firefly Wallet](https://firefly.iota.org/). Both I believe are using electron and offering automatic updates. [Exodus](https://www.exodus.com/bitcoin-wallet) seems to be relying on electron as well, and is also offering updates (to double check though).

I'm looking for other examples.

polarker:
1. It's not my specialized topic, but I think we should only support this if we are sure that it's safe. We need to play safe when it comes to security.
2. It happened in the past that one of a known wallet got hacked due to automatic update. Basically, it's hard to ensure that you are downloading the right package.
3. I have always been suggesting that we should investigate other projects for critical topics. I want to emphasize this again, please do investigate!!

polarker:
Here is a nice list of desktop wallets for ETH: https://docs.ethhub.io/using-ethereum/wallets/desktop/

mvaivre:
> 1. It's not my specialized topic, but I think we should only support this if we are sure that it's safe. We need to play safe when it comes to security.
> 2. It happened in the past that a known wallet (cannot find the reference for now) got hacked due to automatic update. Basically, it's hard to ensure that you are downloading the right package.
> 3. I have always been suggesting that we should investigate other projects for critical topics. I want to emphasize this again, please do investigate!!

1. Of course, hence this investigation issue.
2. Very interested to know more about this story!
3. As always: of course, this is what we've started to do and what we are doing here. 

> Here is a nice list of desktop wallets for ETH: https://docs.ethhub.io/using-ethereum/wallets/desktop/

[...]

PR #433, opened 2022-11-15
Integrate electron-updater

nop33:
Related to #141

## Sound on 🔉 !

https://user-images.githubusercontent.com/1579899/201922192-92a6a8d8-e8eb-486f-88dd-b70287ad5ca9.mp4

Resources that helped me in this PR:
- https://www.electron.build/auto-update
- https://www.electron.build/configuration/publish
- https://mmelikes.medium.com/electron-auto-updater-with-frontend-manipulation-33d3bc5057f3
- https://www.electronjs.org/docs/latest/tutorial/ipc
- https://www.electron.build/auto-update#debugging

### DIFF
--- a/src/modals/UpdateWalletModal.tsx
+++ b/src/modals/UpdateWalletModal.tsx
@@ -51,27 +51,23 @@ const UpdateWalletModal = ({ onClose, newVersion, startDownload }: UpdateWalletM
   }
 
   useEffect(() => {
-    const removeListener = electron?.updater.onUpdateDownloadProgress((_, info) => setPercent(info.percent))
-
-    return () => removeListener
-  }, [])
-
-  useEffect(() => {
-    const removeListener = electron?.updater.onUpdateDownloaded(() => {
+    const removeUpdateDownloadProgressListener = electron?.updater.onUpdateDownloadProgress((_, info) =>
+      setPercent(info.percent)
+    )
+    const removeUpdateDownloadedListener = electron?.updater.onUpdateDownloaded(() => {
       // Delay success message to give time for download validation errors to arise if any
       setTimeout(() => setStatus('download-finished'), 1000)
     })
-
-    return () => removeListener
-  }, [])
-
-  useEffect(() => {
-    const removeListener = electron?.updater.onError((_, error) => {
+    const removeonErrorListener = electron?.updater.onError((_, error) => {
       setStatus('download-failed')
       setError(error.toString())
     })
 
-    return () => removeListener
+    return () => {
+      removeUpdateDownloadProgressListener && removeUpdateDownloadProgressListener()
+      removeUpdateDownloadedListener && removeUpdateDownloadedListener()
+      removeonErrorListener && removeonErrorListener
+    }
   }, [])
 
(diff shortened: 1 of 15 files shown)


src/modals/UpdateWalletModal.tsx
```diff
@@ -17,37 +17,96 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 import CenteredModal, { ModalFooterButton, ModalFooterButtons } from './CenteredModal'
 
 interface UpdateWalletModalProps {
   onClose: () => void
   newVersion: string
+  startDownload?: boolean
 }
 
-const UpdateWalletModal = ({ onClose, newVersion }: UpdateWalletModalProps) => {
+type Status = 'download-available' | 'downloading' | 'download-finished' | 'download-failed'
```
mvaivre, line 35: Nit: `UpdateStatus`?

src/modals/UpdateWalletModal.tsx
```diff
@@ -17,37 +17,96 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+  }, [])
+
+  useEffect(() => {
+    const removeListener = electron?.updater.onError((_, error) => {
+      setStatus('download-failed')
+      setError(error.toString())
+    })
+
+    return () => removeListener
+  }, [])
```
mvaivre, line 75: Nit: could we regroup the 3 listeners setup in one `useEffect` call?

mvaivre (commented):
Excellent. Just a couple of nits.

Also it seems like you didn't push your config changes in package.json, but I guess that's on purpose, as it's not useful until we're pointing to GH?

src/modals/UpdateWalletModal.tsx
```diff
@@ -17,37 +17,96 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
+
+  useEffect(() => {
+    const removeListener = electron?.updater.onUpdateDownloaded(() => {
+      // Delay success message to give time for download validation errors to arise if any
+      setTimeout(() => setStatus('download-finished'), 1000)
+    })
+
+    return () => removeListener
```
mvaivre, line 65: Is this really cleaning the effect up? 

To do so, here we would need to call `ipcRenderer.removeListener('updater:updateDownloaded', callback)`.

nop33:
> Also it seems like you didn't push your config changes in package.json, but I guess that's on purpose, as it's not useful until we're pointing to GH?

This is correct. In fact, we shouldn't need any configs to be added because based on the docs, if the GitHub env var is detected, it uses GitHub as default. electron-updater is pretty smart ;)

> If GH_TOKEN or GITHUB_TOKEN is defined — defaults to [{provider: "github"}].

https://www.electron.build/configuration/publish

src/modals/UpdateWalletModal.tsx
nop33: You are absolutely right. Working on it.

nop33: You were correct, thanks for spotting this! I tested my changes and everything seems to work. Pushed.

PR #434, opened 2022-11-15
Fix theme initialization and respect user's pref

### LINKED ISSUE #343: Listen to OS's theme change and change app's theme accordingly
Following #339, let's get the app's theme to change in real-time whenever the OS's theme is updated. And make sure to override the OS theme whenever the user manually change the theme. Persist that change between multiple sessions.

nop33:
Closes #343


mvaivre (commented):
Thanks for this!

After manual testing, it appears that this PR doesn't offer the desired behaviour. In particular, the theme isn't updated when the OS's theme changes. The desired behaviour is:

- An OS theme change should automatically trigger an app's theme change;
- If the user _then_ changes the app's theme manually in the settings, the app theme is updated;
- The new theme value is persisted, until the OS's theme changes again (or the user manually changes the app's theme again).

nop33:
Ah, thanks for the feedback @mvaivre!

> An OS theme change should automatically trigger an app's theme change;

May I ask, how often is it that a user changes their OS theme settings, and what's the likelihood that they will have our app open at the same time? Am I misunderstanding sth? I'll work on it, I was just curious! 

> The new theme value is persisted, until the OS's theme changes again (or the user manually changes the app's theme again).

Does this mean that an OS theme change should OVERRIDE the user-defined theme settings of our app? :/ I don't think this is good imo. My opinion is that if the user has not given a preference for theme, we default to whatever the user chose for the OS theme, and that's it. The app theme should be independent from the OS theme from that point on. But if you strongly believe that a change in OS theme should override the preference of the user for the app, I can work on that.

mvaivre:
> May I ask, how often is it that a user changes their OS theme settings, and what's the likelihood that they will have our app open at the same time? Am I misunderstanding sth? I'll work on it, I was just curious!

I agree that this is probably not common, BUT I realised that every other apps that I use often (browsers, messengers, OS native apps) are implementing this behaviour... So I believe that this is sort of a guideline, it's expected.

> But if you strongly believe that a change in OS theme should override the preference of the user for the app, I can work on that.

And what about offering 3 choices in the settings modal, like: 
image

By default, we use whatever the system is using. If the user use the quick toggle _in the app's header_, we consider that this is an override and set either `light` or `dark`. 
However, having the "system" option available in the settings modal would let the user go back to the default behaviour. WDYT?

nop33:
> I realised that every other apps that I use often (browsers, messengers, OS native apps) are implementing this behaviour... So I believe that this is sort of a guideline, it's expected.

Interesting, I never noticed cause I never change my theme after setting it to dark I guess! Thanks for letting me know!

I totally agree with your suggestion for a third "System" option!

So, just a summary to ensure I got everything right:

1. In a fresh wallet install, the theme settings is set to `"system"`. This means that every time the app launches, we check the theme of the system and use that. The stored theme settings is the string `"system"`. If the theme changes WHILE the app is open, it should respect the change and also switch the theme.
2. If the user manually changes the theme settings to either `"light"` or `"dark"`, the app no longer needs to respect the OS theme, even if the OS theme changes. This happens again only when the user sets it back to `"system"`.

mvaivre:
You got it perfectly right! 💯

nop33:
There we go! 🚀 

## Sound ON! 🔉 

https://user-images.githubusercontent.com/1579899/203587742-bbca6c18-54cf-41b5-9f70-a550b64d9bab.mp4

mvaivre (approved):
yes.

Issue #438, opened 2022-11-23
[Placeholder] Investigate deep linking


mvaivre:

nop33:
This was implemented in #439 and merged into the `v2.0` branch. I think we can close it.

PR #439, opened 2022-11-24
Implement deep linking

nop33:
Related to #438


public/electron.js
```diff
@@ -220,6 +230,8 @@ if (!gotTheLock) {
 
     ipcMain.handle('updater:quitAndInstallUpdate', () => autoUpdater.quitAndInstall())
 
+    ipcMain.handle('wc:hideApp', () => app.hide())
```
mvaivre, line 233: I wonder if we should associate the `hideApp` action to wc. This might be used differently in the future. Would a `app` namespace make sense? `app:hide`. ?

nop33: It would! But at the moment it's only part of the `wc` workflow, so I'd personally leave it there until the moment other workflows need to hide the app as well. WDYT?

mvaivre: mhh hiding the app is such a generic action that it doesn't really make sense to me to associate it to `wc`.
What would be the drawbacks of using `app:hide`?

nop33: no drawbacks at all, I'll make the change!

nop33: done!

mvaivre (approved):
Super nice!!

PR #444, opened 2022-12-06
Address discovery on wallet import

nop33:
This PR:
- Performs address discovery when importing a wallet (while the user sees the welcome confetti screen)
- Outsources to workers the workload of deriving addresses:
   - based on indexes found in local storage when unlocking a wallet
   - when generating 1 address per group
- Cleans up app loading logic
- Introduces the `activeWalletSlice` store and removes the `wallet` from the global context
- Cleans up global and addresses contexts

Things that still could be improved:
- The loading animation freezes after the worker has finished. The problem is that our method of updating the state is not optimal (we create a `new Map` and replace the existing one). This should be improved when switching to [Redux with normalized data](https://redux.js.org/usage/structuring-reducers/normalizing-state-shape).


nop33:
Thanks to your awesome work with Vite, I could now write the web workers in TypeScript 🎉 

@mvaivre, requesting review :)

mvaivre (approved):
Great work! Nice step forward 👍

PR #446, opened 2022-12-13
Sign and notarize binaries on releases

killerwhile:


mvaivre:
Looks good! 🙏  I believe that we won't release anything before we get signing for windows and linux as well.

Do you want to iterate on this PR directly (it's pretty short for now and adding the other OSes shouldn't make it may bigger). 
That may makes things easier for us to follow :)

nop33 (approved):

PR #450, opened 2022-12-15
Fix build script

nop33:


mvaivre (commented):
Hey! Thanks for pointing this potential issue! But this isn't correct either, the `VITE_VERSION` env variable isn't properly passed to the build (version isn't shown in the settings modal).

I'll check how to properly use `cross-env` to pass the 2 variables before the `build` script, then I'll go ahead and fix my PR :)

nop33:
ah, ok thanks! I just saw the GH action failing and thought to give it a quick try. I'll close it then.

PR #459, opened 2023-01-04
Thin sidebar navigation and initial notifications modal

nop33:
image

image

PR #460, opened 2023-01-04
Cleanup z-index and create better page sections

nop33:
See https://github.com/alephium/desktop-wallet/pull/459#issuecomment-1371248327


mvaivre:
Great initiative! I'm starting with the functional tests. _Developing..._

- Not sure if it's related to this PR, but the settings modal height is "stuck" after you navigate to its "Wallet" tab. As the settings modal will be redesigned, this may not need to be fixed, but I'm signalling anyway ;)

src/components/UnlockedWalletLayout.tsx
```diff
@@ -58,88 +63,105 @@ const WalletLayout: FC = ({ children }) => {
...
+        <WalletSidebar>
+          <CurrentWalletInitials
+            onClick={() => setIsNotificationsModalOpen(true)}
+            style={{ zIndex: isNotificationsModalOpen ? 2 : undefined }}
+          >
+            {activeWalletNameInitials}
+          </CurrentWalletInitials>
+          <SideNavigation>
+            <NavItem Icon={Layers} label={t`Overview`} to="/wallet/overview" />
+            <NavItem Icon={ArrowLeftRight} label={t`Send`} onClick={() => openSendModal(TxType.TRANSFER)} />
```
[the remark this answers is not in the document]
mvaivre: Just a suggestion, not related to a bug: in the navbar it would be great to make the data-tip to appear on the right, sticked to the parent element (anchor on the right hand side) :)

src/routes/index.tsx
```diff
@@ -19,28 +19,21 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
-        <Route path="/create/:step" element={<CreateWalletRoutes />} />
-        <Route path="/import/:step" element={<ImportWalletRoutes />} />
-        <Route path="/wallet/*" element={<UnlockedWalletRoutes />} />
-        <Route path="" element={<HomePage />} />
-      </Routes>
-    </AnimateSharedLayout>
-  )
-}
+const Router = () => (
+  <AnimateSharedLayout type="crossfade">
```
mvaivre, line 29: This has been removed as of framer motion 5. This layout refactor may be a good oppurtinity to upgrade! :)

mvaivre (commented):
Really great job! 👏

nop33:
> Not sure if it's related to this PR, but the settings modal height is "stuck" after you navigate to its "Wallet" tab (or expand any children inducing a change in a modal's height). As the settings modal will be redesigned, this may not need to be fixed, but I'm signalling anyway ;)

👎 It seems to be the case on `master` as well... It also happens in the Send modal... We need to have a look into this, but I'll leave it out of scope for this PR.

> Clicking "Start" in the "Consolidate UTXOs" card doesn't do anything.

🤔 Cannot reproduce...

<details><summary>See screenshot</summary>

image

</details>

> Clicking the settings icon in the Address details page doesn't do anything.

🤔 I cannot reproduce this...

<details><summary>See screenshots</summary>

image

image

</details>

> The theme toggle isn't usable in the "welcome back" screen.

✅ Fixed!

> Pressing escape to close a popup shouldn't close all open popups. This may not need to be fixed as the Select component will be refactored.

🤔  Yeah, this happens also in our current master, when 2 modals are open and pressing ESC, everything closes. I have attemted to solve this with capturing the event and stopping the propagation, but I didn't manage to make it work... I'll give it another try.

> The "Wallet + notif center" popup theme doesn't update properly

✅ Fixed!

src/routes/index.tsx
nop33: Nice, upgraded to Framer 5 and removed this!

nop33:
>> Not sure if it's related to this PR, but the settings modal height is "stuck" after you navigate to its "Wallet" tab (or expand any children inducing a change in a modal's height). As the settings modal will be redesigned, this may not need to be fixed, but I'm signalling anyway ;)
>
> 👎 It seems to be the case on master as well... It also happens in the Send modal... We need to have a look into this, but I'll leave it out of scope for this PR.

Wow, that was much more complicated than thought, but I solved it 😄 When opening a modal, I register a unique `id` and store it in Redux, creating a stack of modal `id`s. When a modal closes, I select the previous modal and `focus` it so that the `ESC` key can be captured. And of course, I stop the bubbling of the `keyDown` event so that only one modal closes. Tricky!

mvaivre:
Awesome 👏  

Regarding the 2 errors you can't reproduce, that's... interesting.
I get them only when using the testnet genesis address 😕 It works perfectly with other wallets.

This is what I get when opening the send modal :
```
Uncaught TypeError: Cannot read properties of undefined (reading 'toString')
    at e2.convertSetToAlph (numbers.js:124:40)
```

Don't hesitate to ping me if you need the wallet mnemonic.

mvaivre:
Now, a little note regarding the tooltip behaviour : I would force the placement to the right only in the nav bar. Also, I would start using the effect "solid" instead of "float" accros the app.
I see that the new version (v5) of `react-tooltip` allows to configure anchor element using the `data-tooltip-*` props. This is great as it would allow us to tweak the tooltip behaviour differently in different parts of the app.

I went ahead and try to upgrade, but I quickly ended up in a dependency hell, as this requires react 18, which in turn forces to upgrade X, which in turn forces to, blablabla...
I'll open an issue ;)

nop33:
> Regarding the 2 errors you can't reproduce, that's... interesting.
I get them only when using the testnet genesis address

This is getting very weird, cause I am also using the testnet genesis wallet (the one where the seed starts with `towards`) and I am connecting to testnet:
- https://wallet-v15.testnet.alephium.org
- https://backend-v110.testnet.alephium.org

I still don't get an error...

Also, the error you get seems quite weird... There is one call to `toString()` inside the `convertSetToAlph` method, and this method is used in the `AmountInput` component that is in turn used by the `SendModalTransfer` component. Basically, it would only fail if `availableAmount` is undefined, but it can never be undefined since its type is `bigint`...

Very weird thing to debug! What else could I try to simulate your environment?

<details><summary>See screenshot</summary>

image

</details>

nop33:
> Now, a little note regarding the tooltip behaviour : I would force the placement to the right only in the nav bar. Also, I would start using the effect "solid" instead of "float" accros the app.
> I see that the new version (v5) of react-tooltip allows to configure anchor element using the data-tooltip-* props. This is great as it would allow us to tweak the tooltip behaviour differently in different parts of the app.

Thanks for these great suggestions! Since you've created the issues, do you mind if I implement them in a separate PR or should they be part of this?

Apart from the tooltips and the strange error you get with the genesis wallet, I don't think there are any other remaining issues with this PR.

mvaivre:
Of course, let's make them part of another PR. Thanks!
And let's see if my issues persist in my upcoming tests. I'll investigate further on my side.

mvaivre (approved):

PR #461, opened 2023-01-05
Addresses page

nop33:
Work in progress of course, but since I'll work on it on Monday again, I thought it's a good chunk of code to review and merge :)

image


mvaivre:
On fire 🔥♨️

mvaivre (commented):
First review done! 👏

src/components/AddressesTabContent.tsx
```diff
@@ -0,0 +1,96 @@
+/*
```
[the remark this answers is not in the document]
nop33: > We would keep only generic / global components (ie. not attached to a specific page) in the components folder.

Sounds like a good idea! How would u like to handle the placement of the `pages/AddressPage.tsx` component? Should it become `pages/AddressPage/index.tsx`?

src/components/AddressesTabContent.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const AddressesTabContent = () => {
+  const navigate = useNavigate()
+  const { addresses } = useAddressesContext()
+  const { t } = useTranslation('App')
+
+  const navigateToAddressDetailsPage = (addressHash: AddressHash) => () => navigate(`/wallet/addresses/${addressHash}`)
+
+  return (
+    <motion.div initial={{ opacity: 0 }} animate={{ opacity: 1 }} transition={{ duration: 0.3 }}>
```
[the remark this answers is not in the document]
nop33: Yes, another great idea 🧠 ⚡

src/pages/Wallet/AddressesPage.tsx
```diff
@@ -92,102 +81,86 @@ const AddressesPage = () => {
...
+                label={t`Show advanced operations`}
+                toggled={isAdvancedSectionOpen}
+                onToggle={() => setIsAdvancedSectionOpen(!isAdvancedSectionOpen)}
+              />
+            }
           />
-        </AdvancedOperations>
-      </ExpandableSection>
+        </AdvancedOperationsHeader>
+        <Collapsable
```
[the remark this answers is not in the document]
nop33: Ah, interesting idea! So basically it would feel like a bottom modal. If this row would be stick'ed to the bottom wouldn't that limit the available vertical space of the "Addresses" tab? Shall we also try the Ctrl+Scroll idea in there as well?

Another clarification: Would the "Advanced operations" row with the toggle button remain at the bottom when activated and only the operation boxes would appear on top? Or would the whole row move up? I think I find the first option better because I don't have to move the mouse to toggle it off. WDYT?

src/pages/Wallet/AddressesPage.tsx
```diff
@@ -92,102 +81,86 @@ const AddressesPage = () => {
...
+            <AnimatePresence>
+              {!infoMessageClosed && (
+                <InfoMessage link={links.faq} onClose={closeInfoMessage}>
+                  {t`Want to know more? Click here to take a look at our FAQ!`}
+                </InfoMessage>
+              )}
+            </AnimatePresence>
+          </div>
+        </Header>
+        <Tabs>
```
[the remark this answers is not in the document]
nop33: > probably what you thought

Exactly :D Since this

src/components/AddressesTabContent.tsx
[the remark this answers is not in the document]
nop33: I updated the PR with the proposed structure :)

src/components/AddressesTabContent.tsx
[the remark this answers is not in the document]
nop33: I implemented this idea, let me know what you think with what i pushed!

src/pages/Wallet/AddressesPage.tsx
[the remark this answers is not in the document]
nop33: Here's what I could come up with! I am not 100% happy with the bottom modal because I need to define its content height. I couldn't yet find a way to have dynamic modal height based on the content height. I would like it to move on, however, since we are still experimenting with this. Here's a working prototype:

https://user-images.githubusercontent.com/1579899/211323385-a29fec28-4df0-4e0b-9f0e-2397ab61c4fd.mp4

mvaivre:
Very very nice improvements based on your comments and videos! I'm still abroad, just back online. I hope to find some time tomorrow to dig into the code. I'll be back in Switzerland on Wed :)

src/pages/Wallet/AddressesPage.tsx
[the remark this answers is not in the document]
mvaivre: That's exactly what I had in mind, and your implementation looks perfect for now. Let's see if we want to expand this to other pages in the future. The height may need to be computed manually.

src/pages/UnlockedWallet/Layout.tsx
```diff
@@ -36,13 +36,13 @@ import { appHeaderHeightPx, walletSidebarWidthPx } from '@/style/globalStyles'
...
-import Scrollbar from './Scrollbar'
-import Spinner from './Spinner'
+import AppHeader from '../../components/AppHeader'
+import Button from '../../components/Button'
+import NavItem from '../../components/NavItem'
+import Scrollbar from '../../components/Scrollbar'
+import Spinner from '../../components/Spinner'
 
-interface WalletLayout {
+interface UnlockedWalletLayout {
```
mvaivre, line 46: `UnlockedWalletLayoutProps`?

src/pages/UnlockedWallet/Layout.tsx
```diff
@@ -36,13 +36,13 @@ import { appHeaderHeightPx, walletSidebarWidthPx } from '@/style/globalStyles'
 import { TxType } from '@/types/transactions'
```
mvaivre, line 37: Let's call this file `UnlockedWalletLayout` for consistency and "searchability" sake? :)

src/pages/UnlockedWallet/OverviewPage/AddressSummaryCard.tsx
```diff
@@ -23,11 +23,11 @@ import styled, { css } from 'styled-components'
...
-import AddressBadge from './AddressBadge'
-import Amount from './Amount'
-import ClipboardButton from './Buttons/ClipboardButton'
-import QRCodeButton from './Buttons/QRCodeButton'
-import InputArea from './Inputs/InputArea'
+import AddressBadge from '../../../components/AddressBadge'
+import Amount from '../../../components/Amount'
+import ClipboardButton from '../../../components/Buttons/ClipboardButton'
+import QRCodeButton from '../../../components/Buttons/QRCodeButton'
+import InputArea from '../../../components/Inputs/InputArea'
```
mvaivre, line 31: Nit: we may use the `@/components` path for those.

src/pages/UnlockedWallet/OverviewPage/WalletSummaryCard.tsx
```diff
@@ -26,7 +26,7 @@ import { useGlobalContext } from '@/contexts/global'
 import { useGetPriceQuery } from '@/store/priceApiSlice'
 import { currencies } from '@/utils/currencies'
 
-import Amount from './Amount'
+import Amount from '../../../components/Amount'
```
mvaivre, line 29: `@/components`

src/animations/index.ts
```diff
@@ -0,0 +1,66 @@
+/*
```
mvaivre, line 1: Very nice! 👏

src/components/SnackbarManager.tsx
```diff
@@ -48,12 +49,7 @@ const SnackbarManager = ({ message }: { message: SnackbarMessage | undefined })
     <SnackbarManagerContainer>
       <AnimatePresence>
         {message && (
-          <SnackbarPopup
-            initial={{ y: 10, opacity: 0 }}
-            animate={{ y: 0, opacity: 1 }}
-            exit={{ opacity: 0 }}
-            className={message?.type}
-          >
+          <SnackbarPopup {...fadeInBottom} {...fadeOut} className={message?.type}>
```
mvaivre, line 52: NIT: We use 2 technics to compose animation props:
1) You merge animation objects in `src/animations/index`, in this case this would lead to `fadeInOutBottom`;
2) You use multiple spreads directly in the component props, like here.

We should probably use only one technique. Either we add all combinations used in the app in `src/animations/index`, or we remove from the file all merged animations (`fadeInOut`, `fadeInSlowly`, `fadeOutFast` etc.).

I think that I'm in favour of the first option, in order to keep our components props cleaner.

WDYT?

src/modals/BottomModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+import ModalContainer, { ModalContainerProps } from './ModalContainer'
+
+interface BottomModalProps extends ModalContainerProps {
+  label: string
+  contentHeight?: number
+  className?: string
+}
+
+const BottomModal = ({ onClose, children, label, contentHeight, className }: BottomModalProps) => {
+  const elRef = useFocusOnMount<HTMLDivElement>()
```
mvaivre, line 35: This is probably a remnant from an attempt to compute the height manually, and could be removed for now :)

nop33: Actually not! This is to move the focus to the modal for 2 reasons:
1. accessibility (screen readers)
2. so that the ESC key press is captured by the modal

src/components/SnackbarManager.tsx
nop33: > You merge animation objects in src/animations/index, in this case this would lead to fadeInOutBottom;

Actually, it wouldn't. It should fade in from bottom, but it shouldn't fade out from bottom, it should just fade out. So, technically, it should be `fadeInBottomFadeOut`... and it's not really reused anywhere else.

> We should probably use only one technique.

I personally like that we have the option of composability in JSX. Forcing one technic might lead us to having ugly names like the one I mentioned above... I think we should keep reusable prop combinations in the animations file, and for special cases like this one use composability. WDYT?

src/modals/BottomModal.tsx
mvaivre: Ah ok! Perfect.

src/components/SnackbarManager.tsx
mvaivre: It was a nit, I'm okay with starting this way and adjusting course if necessary! 🆗

PR #465, opened 2023-01-10
Upgrade React

nop33:
To upgrade React from 17 to 18 I had to upgrade the following libraries as well:
- `react`
- `react-dom`
- `react-is`
- `@testing-library/react`
- `@types/react`
- `@types/react-dom`
- `framer-motion`
- `react-tooltip`
- `wait-on`

To minimize the changes required by the upgrade to React 18 (due to the fact that `children` is now not part of the `FC` type), I've created the following type:

```ts
type FC<P = unknown> = React.FC<React.PropsWithChildren<P>>
```

I also upgraded vite packages:
- `vite`
- `@vitejs/plugin-react`
- `@vitest/web-worker`

I had to remove the `esbuild` polyfills for it to work. The dev server, as well as the production build seem to still work, so I am not 100% sure they were needed. Let's keep an eye on this.

Since I was at it, I upgraded some more packages that did not cause any breaking changes:
- `electron`
- `eslint-config-prettier`
- `eslint-plugin-simple-import-sort`
- `lucide-react`
- `prettier`
- `type-fest`
- `i18next`
- `react-i18next`

The cool thing about upgrading `react-tooltip` from v4 to v5 is that we don't need all these crazy `ReactTooltip.rebuild()` calls, which reduces many of our `useEffects`. The bad thing about v5 `react-tooltip` is that we have to wrap every component that needs a tooltip with `<TooltipWrapper content="...">`, unless we provide a unique `<Tooltip>` instance to each element that needs a tooltip 🤢.


mvaivre:
> The bad thing about v5 react-tooltip is that we have to wrap every component that needs a tooltip with <TooltipWrapper content="...">, unless we provide a unique <Tooltip> instance to each element that needs a tooltip 🤢.

Well... at least we can now say that it's super explicit 🙃

nop33:
One more cool thing in this PR is that our translation strings are not typed :) Typescript can detect it if a key is missing from the English translation file.

src/pages/UnlockedWallet/UnlockedWalletLayout.tsx
```diff
@@ -87,37 +88,34 @@ const UnlockedWalletLayout: FC<UnlockedWalletLayoutProps> = ({ children, classNa
...
-                transparent
-                squared
-                onClick={refreshAddressesData}
-                disabled={isLoadingData}
-                aria-label={t`Refresh`}
-                data-tip={t`Refresh data`}
-              >
-                {isLoadingData ? <Spinner /> : <RefreshCw />}
-              </RefreshButton>
+              <TooltipWrapper content={t`Refresh data`}>
```
mvaivre, line 103: Should it be `t('Refresh data')`?

src/pages/UnlockedWallet/UnlockedWalletLayout.tsx
```diff
@@ -87,37 +88,34 @@ const UnlockedWalletLayout: FC<UnlockedWalletLayoutProps> = ({ children, classNa
...
+                <RefreshButton
+                  transparent
+                  squared
+                  onClick={refreshAddressesData}
+                  disabled={isLoadingData}
+                  aria-label={t`Refresh`}
```
mvaivre, line 109: `t('Refresh')`

mvaivre: And many others across the app. Should we change them all?

mvaivre (approved):

PR #467, opened 2023-01-10
Fix i18n

### LINKED ISSUE #466: Translation strings that include whitespaces are not getting translated
It happens on `master`, but not on `v1.4.0`. Could Vite be the problem?

nop33:
Closes #466


src/i18n.ts
```diff
@@ -17,28 +17,26 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 
 import i18next from 'i18next'
-import resourcesToBackend from 'i18next-resources-to-backend'
 import { initReactI18next } from 'react-i18next'
 
-i18next
-  .use(
-    resourcesToBackend((language, namespace, callback) => {
-      import(`../locales/${language}/${namespace}.json`)
-        .then((resources) => {
```
nop33, line 27: For some reason that I did not manage to identify, `resources` only contains the single-word keys, so all translation keys with more than 1 word are simply not there...

mvaivre (approved):

PR #469, opened 2023-01-11
Implement address cards

nop33:
https://user-images.githubusercontent.com/1579899/211887531-76f1cc3b-bbff-484e-b781-6f53f8009304.mp4


src/components/Button.tsx
```diff
@@ -24,10 +24,9 @@ import styled, { css } from 'styled-components'
 import { sectionChildrenVariants } from './PageComponents/PageContainers'
 
 interface ButtonProps extends HTMLMotionProps<'button'> {
-  secondary?: boolean
-  alert?: boolean
+  mode?: 'primary' | 'secondary' | 'transparent'
```
nop33, line 27: Any better idea for this prop name? `type` is taken by `HTMLMotionProps`...

mvaivre: role?

mvaivre:
Super nice evolution! 

Regarding the layout: I've been trying the UI with many addresses and it's a bit hard to read when the cards are spread over multiple rows. 
I initially intended the address list to be scrollable horizontally : this could be the opportunity to see how it would behave?
If the page has no overflow (should be the case here), then we can allow to scroll with the mouse wheel without pressing CTRL.

src/components/Button.tsx
nop33: done!

nop33:
> If the page has no overflow (should be the case here)

FWIT, the behavior now is that you can scroll vertically:

https://user-images.githubusercontent.com/1579899/212046202-488672e1-6d58-42f2-81eb-536815dcf893.mp4

mvaivre:
> FWIT, the behavior now is that you can scroll vertically:

Yes I know, but I find it hard to parse with many addresses.

image

Let's start tweaking the cards' design before trying horizontal scrolling.

mvaivre (approved):
👍 👏

PR #474, opened 2023-01-17
Contact management

nop33:
ai ai ai, 1003 LOC! 😅  But hey, it's 121 less if you exclude the translation files and the package.lock 😉 

Assumptions that I made:
1. The user should not be able to create 2 contacts with the same address because, well, it doesn't make sense... Do you have 2 contacts with the same phone number on your phone?
2. The user should not be able to create 2 contacts with the same name (case insensitive). I believe this could lead to mistakes. Let's say that Alice wants to send some funds to Bob. Half a year ago, she had added one of Bob's addresses in her contact list. Bob now gives her a new address. She forgot that she already had Bob in her contact list, so she adds a new contact, with the name "Bob" and his new address. When she opens the Send modal to send him some funds, she could accidentally click on the Bob's OLD contact, ending up sending funds to the old address (which who knows, maybe it's compromised or lost).

The next thing I want to work on is updating the "recipient" field of the Send modal so that the user can choose one of their contacts (but they should also be able to just type/paste a different recipient.

https://user-images.githubusercontent.com/1579899/212948398-c27d4229-e510-403b-9192-a90c084164cb.mp4


src/persistent-storage/address-metadata.ts
```diff
@@ -0,0 +1,52 @@
...
+        index,
+        ...settings
+      })
+    } else {
+      Object.assign(existingAddressMetadata, settings)
+    }
+    console.log(`🟠 Storing address index ${index} metadata locally`)
+
+    this.store(JSON.stringify(addressesMetadata), { mnemonic, walletName }, isPassphraseUsed)
+  }
```
nop33, line 47: This is copy/pasta, no new code 🍝 😇

mvaivre:
Your assumptions are good IMO :+1:

src/pages/UnlockedWallet/AddressesPage/index.tsx
```diff
@@ -103,67 +103,69 @@ const AddressesPage = () => {
...
-            >
-              <AdvancedOperations>
-                <OperationBox
-                  title={t`Consolidate UTXOs`}
-                  Icon={<Codesandbox color="#64f6c2" strokeWidth={1} size={46} />}
-                  description={t`Consolidate (merge) your UTXOs into one.`}
-                  buttonText={t`Start`}
-                  onButtonClick={() => setIsConsolidationModalOpen(true)}
-                  infoLink={links.utxoConsolidation}
+      {currentTab.value === 'addresses' && (
```
mvaivre, line 106: Mhh this makes me think about the UI hierarchy again. We don't have any advanced actions linked to contacts, but in theory, changing a tab shouldn't have any impact on anything out of the tab's "folder". Don't know if I'm clear.

Maybe I need to redesign this "folder" a bit. To be continued...

mvaivre: Something like this (quick and dirty) :

image

WDYT?

src/utils/formValidation.ts
```diff
@@ -0,0 +1,58 @@
+/*
```
mvaivre, line 1: NIT: This may become a big mix of many kind of validations. Could we eventually split those in folders like : `utils/form-validation/addresses-validation.ts` ?

mvaivre (commented):
👏  Noice! I love where it's going.

src/contexts/addresses.tsx
```diff
@@ -210,7 +211,7 @@ export const AddressesContextProvider: FC<{ overrideContextValue?: PartialDeep<A
       if (!activeWalletMnemonic) throw new Error('Could not update address settings, mnemonic not found')
 
       if (!isPassphraseUsed)
-        storeAddressMetadataOfWallet(
+        AddressMetadataStorage.storeAddressMetadata(
```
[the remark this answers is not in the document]
nop33: I will reply to EDIT2, because we both got to the same conclusion :D

That's exactly what I wanted to do. I wanted to override basically the `store` method of `PersistentEncryptedStorage` with a `store` method in `AddressMetadataStorage`, which would then call `super.store`. The problem is that when you want to override a class method, the params of that method need to be the same. Which is not the case here :/ Thus, I made the `store` method protected (so only subclasses can access it) and created `storeAddressMetadata`.

I am open to any other ideas... One idea is to rename the `store` method of `PersistentEncryptedStorage` to something like `_store`. Then, we are free to create `store` in `AddressMetadataStorage`. WDYT?

src/modals/ContactFormModal.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { useGlobalContext } from '@/contexts/global'
+import { useAppSelector } from '@/hooks/redux'
+import ContactStorage from '@/persistent-storage/contacts'
+import { Contact } from '@/types/contacts'
+import {
+  requiredErrorMessage,
+  validateIsAddressValid,
+  validateIsContactAddressValid,
+  validateIsContactNameValid
+} from '@/utils/formValidation'
```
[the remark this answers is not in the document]
nop33: Because Ilias' brain doesn't always perform 💩

src/pages/UnlockedWallet/AddressesPage/index.tsx
nop33: I know exactly what you mean. I am not 100% sold on the vertical scrolling, but we could try it. Should it be part of this PR?

src/utils/formValidation.ts
nop33: Sure, once we have more stuff in it we could do that!

src/contexts/addresses.tsx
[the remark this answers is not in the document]
mvaivre: 👍 yes, your idea sounds like a fine compromise to me.

src/pages/UnlockedWallet/AddressesPage/index.tsx
mvaivre: Let's try the vertical scrolling and see how it feels. If we end up going for horizontal scrolling, I may switch from cards to "rows" to ease visual parsing.

We can tweak the layout in another PR!

nop33:
@mvaivre I addressed all your comments in the following PR, check it out! #476 

If you're happy with what you see there, we can merge this one :)

mvaivre (approved):

src/pages/UnlockedWallet/AddressesPage/index.tsx
nop33: So, here's a quick and dirty solution. What I don't like about it is when the height of the page is not big enough and it requires its own scrolling. If we remove vertical scrolling of the page and force a fixed `800px` height, then our wallet is not gonna work well in smaller screens.

I find it very annoying to have both veritical and horizontal scroll in the page, because it forces me to move the mouse to an area outside of the "horizontally-scrollable" area so that I can scroll vertically.

I think horizontal scrolling works great on mobile devices. On desktop, I am not a big fan. I think the mental model is that scrolling is always vertical. 

But we don't have to do what the status quo is, we can break things :) 

Lemme know what you think.


https://user-images.githubusercontent.com/1579899/213237238-ed16f5b1-e097-4f13-8e5b-360a5ebd2fcb.mp4

mvaivre: Thanks for this. Yeah I'm not 100% convinced yet.
What we could do is to have a switch from card view to list view in the header (like I imagined for the mobile wallet as well). Something like: 
I'd like to design an alternative "vertical list" version that is not a cold and dense table... Let's see when I can do that!

nop33: Alright, I'll move on to other screens then and come back to this at a later point! Thanks for the discussion!

PR #478, opened 2023-01-19
Use React portals to fix all modals

nop33:
This is a very simple PR that makes use of React Portals to place the modals always at the bottom of the DOM to avoid stacking issues. It will be very useful in the upcoming layout changes of stacked modals. It's a very clean solution imo!


mvaivre (approved):
That's awesome! Didn't know about portals, very very handy.

PR #479, opened 2023-01-19
Overview page layout and dev tools settings

nop33:
image

image


mvaivre:
That's super nice!

I've found no major issue in my functional tests, just a couple of tiny remarks:

- The `box-shadow` of the centered modal should be updated with the following params : `0 0 50px rgb(0 0 0)`
- The amount quickly wraps if the window is not wide enough. I may need to update the design a bit... Let's not touch the layout in this PR.

EDIT: WIP updates have been added to Figma. Don't hesitate to share your feedback in the Figma file. 

image

Switching to code review now.

src/pages/UnlockedWallet/AddressesPage/index.tsx
```diff
@@ -98,7 +101,14 @@ const AddressesPage = () => {
...
             <TabBar items={tabs} onTabChange={(tab) => setCurrentTab(tab)} activeTab={currentTab} />
-            <TabContent>{tabs.find((t) => t.value === currentTab.value)?.component}</TabContent>
+            <TabContent>
+              {
+                {
+                  addresses: <AddressesTabContent />,
+                  contacts: <ContactsTabContent />
+                }[currentTab.value]
+              }
+            </TabContent>
```
mvaivre, line 111: NIT: Should we extract this? 🙃 

```ts
const currentTabComponent = {
  addresses: <AddressesTabContent />,
  contacts: <ContactsTabContent />
}[currentTab.value]

File: src/pages/UnlockedWallet/OverviewPage/AssetsOverviewPanel.tsx [hunk 2]
```diff
@@ -0,0 +1,203 @@
...
+import SettingsModal from '@/modals/SettingsModal'
+import { useGetPriceQuery } from '@/store/priceApiSlice'
+import { currencies } from '@/utils/currencies'
+
+interface AssetsOverviewProps {
+  isLoading?: boolean
+  className?: string
+}
+
+const AssetsOverview = ({ className, isLoading }: AssetsOverviewProps) => {
```
mvaivre, line 42: Nit: based on the name, I thought this would be the "assets / tokens" card. What about `WalletAmountsOverview`, `WalletOverviewPanel` or something similar?

src/modals/SettingsModal/index.tsx
```diff
@@ -16,50 +16,190 @@ You should have received a copy of the GNU Lesser General Public License
...
+          </ColumnHeader>
+          <Scrollbar translateContentSizeYToHolder>
+            <ColumnContent>
+              {
+                {
+                  general: <GeneralSettingsSection />,
+                  wallets: <WalletsSettingsSection />,
+                  network: <NetworkSettingsSection />,
+                  devtools: <DevToolsSettingsSection />
+                }[currentTab.value]
```
mvaivre, line 99: NIT: Same remark as before, I would extract this.

src/components/PageComponents/PageContainers.tsx
```diff
@@ -22,6 +22,8 @@ import styled from 'styled-components'
 import { fadeIn } from '@/animations'
 import { deviceBreakPoints } from '@/style/globalStyles'
 
+import Box from '../Box'
```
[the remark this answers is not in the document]
nop33: It is actually VSCode that automatically imports it like this. It only happens when we are importing a component in a component within the component folder. It'd be quite annoying to then manually change automatic imports :/

src/pages/UnlockedWallet/AddressesPage/index.tsx
nop33: Hm, I actually prefer it in the render function for one main reason: I don't have to scroll the file up and down when I am navigating the JSX of the component. If I get to line 108 and then see a variable then I have to look for the definition of the variable somewhere above... Which is how we had it before and it annoyed me a few times.

I don't have a strong opinion about this, if it's more important for you to do it otherwise I can change it.

src/modals/SettingsModal/index.tsx
nop33: Same reply as above 😇

src/pages/UnlockedWallet/OverviewPage/AssetsOverviewPanel.tsx
nop33: I renamed `AssetsOverviewPanel` to `AmountsOverviewPanel` :) Hope that's fine!

nop33:
> The amount quickly wraps if the window is not wide enough. I may need to update the design a bit... Let's not touch the layout in this PR.
> EDIT: WIP updates have been added to Figma. Don't hesitate to share your feedback in the Figma file.

I kinda like the buttons to be horizontal at the bottom! What if we just wrap the graph to go above or below the amounts in smaller screens? Also, I like how you reduced the width of the chart section, I think it's a good idea.

I made a quick example on Figma, feel free to delete it, I don't want to mess up with your mockups!

image

src/components/PageComponents/PageContainers.tsx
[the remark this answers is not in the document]
mvaivre: Yeah. Not a big deal.

src/pages/UnlockedWallet/AddressesPage/index.tsx
mvaivre: Ok. Well, finding the constant is one cmd + click away. It's the same with callback (like `onClick` methods etc.), everything could be in the render function 🙃 
Kidding, I'm fine with that! Closing.

mvaivre:
Thanks for your suggestions! Let's not forget that we have a window min-width in the production build, so the app should never get that narrow.

I'm not 100% convinced yet. Let's keep the behaviour you propose (reducing the chart section first) before I hopefully find something that works well with every width!

src/components/PageComponents/PageContainers.tsx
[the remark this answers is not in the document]
nop33: So, I figured it out, thanks for pushing me to research this in more detail!

There's a setting on VSCode that fixes that. I suggest we commit the `.vscode/settings.json` file so that we both share the same settings. So far, it only has 1 line, I suggest we add the second:

```json
{
  "typescript": {
    "tsdk": "node_modules/typescript/lib",
    "preferences.importModuleSpecifier": "non-relative"
  }
}
```


image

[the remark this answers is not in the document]
nop33: I will update the import paths of every file in another PR so that this one doesn't grow and I don't get conflicts with the following ones if you don't mind!

nop33:
> I'm not 100% convinced yet. Let's keep the behaviour you propose (reducing the chart section first) before I hopefully find something that works well with every width! But for now my preferred option is the one below, with maybe slightly taller boxes for the amount + graph cards.. :)

OK then! I can work on this in a separate PR if you don't mind! Let me know if there's something else to do in this one :)

mvaivre (approved):

mvaivre:
FYI, update :

image

PR #483, opened 2023-01-24
Cleanup global context

nop33:


mvaivre:
Mhh, the migration introducing the new format for stored wallet (string instead of object) breaks backward compatibility. Once the migration is applied, older app versions won't be able to open any already existing wallet anymore. 
This is an important decision to take. What would be the pros and cons?

mvaivre:
I agree that the cleaning is necessary and that few users should be impacted. I was thinking about potential devs forking our wallet (like Sesame) who way be confused by the  "Wrong password" error. That's an edge case.

So let's go ahead and make sure to mention the breaking change in the release notes & docs. This gives me an idea : what about fetching the release notes from Github and display them in the "Upgrade available" popup in the app ? This would give us the opportunity to properly communicate important info. WDYT?

nop33:
> I was thinking about potential devs forking our wallet (like Sesame) who way be confused by the "Wrong password" error. That's an edge case.

Yes, indeed. But at least they could still restore their mnemonic and it would work.

> This gives me an idea : what about fetching the release notes from Github and display them in the "Upgrade available" popup in the app ? This would give us the opportunity to properly communicate important info. WDYT?

This is a great idea! We can parse the markdown easily.

nop33:
@mvaivre, what is the decision regarding this PR? I am still in favor of it.

mvaivre (approved):
Sorry, I said to go ahead but didn't approve :) 
I'll create an issue about showing the release notes in the upgrade modal.

PR #485, opened 2023-01-25
Fix snapcraft action, add checksums, update network presets

### LINKED ISSUE #141: Automatic update system

### LINKED ISSUE #412: Code signing step in release toolchain
Implement code signing for at least Windows and MacOS to allow for secure auto-update & better trust.

https://www.electronjs.org/docs/latest/tutorial/code-signing

To be discussed with @killerwhile to add that in our distribution tool chain.

### LINKED ISSUE #494: Add checksums to new release workflow
The old release workflow included the generation of checksums. This is needed for at least the first install of the wallet. Any later version should be updated through the automatic update system of the app itself, so no need to download files from GitHub anymore. But for the initial install, a checksum is needed.

https://github.com/alephium/desktop-wallet/blob/master/.github/workflows/release.yml#L98-L108

nop33:
1. Fixes Linux Snapcraft action (thanks @killerwhile ❤️)
2. Adds checksum files for `.exe`, `.deb`, `.dmg`, and `.AppImage` (like our old release workflow had, see #494)
3. Updates mainnet & testnet network presets to point to:
   - `https://wallet-v16/.[mainnet|testnet].alephium.org`
   - `https://backend-v112/.[mainnet|testnet].alephium.org`

Closes #141 

Closes #412 

Closes #494

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,7 @@
   },
   "build": {
     "appId": "org.alephium.alephium-wallet",
+    "productName": "Alephium",
     "files": [
       "build/**/*",
       "public/electron.js",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -116,10 +116,10 @@
     ],
     "linux": {
       "category": "Finance",
-      "executableName": "alephium",
```
nop33, line 119: @killerwhile could it be that the issue #486 is because of this removal? :/

nop33:
This is the last PR before we can proceed with the 1.5.0 release!

killerwhile (approved):
backend-v1112.mainnet is in progress.

Checksum LG
```
shasum -a 256 -c Alephium-1.5.0-rc.1-universal.dmg.checksum
Alephium-1.5.0-rc.1-universal.dmg: OK
```

And all the rest LGTM. Great team work, thanks!

PR #489, opened 2023-01-26
Fix dev mode

### LINKED ISSUE #488: Dev mode always requires refresh to work properly
see https://github.com/alephium/desktop-wallet/issues/486#issuecomment-1404271345

nop33:
Closes #488 

@mvaivre I wonder what your thoughts are on this! I figured that `concurrently`, `cross-env` and `vite` didn't play well together (cannot find any relevant issues on the web).

1. ❌ The following replacement fixes the problem, but then we don't have the `VITE_VERSION` env var available in our code.
   ```diff
   - "start": "cross-env VITE_VERSION=$npm_package_version vite",
   + "start": "vite"
   ```
2. 🤔 Logically, in order to use the `BROWSER=none` env var in the `electron-dev` script, we should have used `cross-env` before it, as we do in other scripts that need env vars:
   ```diff
   - "electron-dev": "concurrently \"BROWSER=none npm run start\" \"electron .\"",
   + "electron-dev": "concurrently \"cross-env BROWSER=none npm run start\" \"electron .\"",
   ```
3. ✅ The following replacement fixes the problem and we can have access to the `VITE_VERSION` env var in both browser and electron environments:
   ```diff
   - "electron-dev": "concurrently \"BROWSER=none npm run start\" \"electron .\"",
   + "electron-dev": "concurrently \"cross-env BROWSER=none VITE_VERSION=$npm_package_version vite\" \"electron .\"",
   ```
4. 🧼 It is very common in frontend projects to have a `start` script to start the dev server. I think it makes more sense to rename the script to start the electron dev server to `start:electron` instead of `electron-dev`.
5. I cannot find [any references](https://vitejs.dev/guide/env-and-mode.html#env-variables-and-modes) in `vite` for the `BROWSER=none` env var. Is it needed?


mvaivre:
Thanks for this! It fixes the issue 💯 

- BROWSER=none seems to be a relic of our CRA past and can be removed.
- I agree with you about our start script name. Let's go with `start:electron`

mvaivre (approved):

PR #492, opened 2023-01-29
Select's redesign

mvaivre:
Reduce distance to travel with cursor. Automatically detects if Popup is overflowing the window, move it accordingly.

### DIFF
--- a/src/components/Popup.tsx
+++ b/src/components/Popup.tsx
@@ -15,9 +15,10 @@ GNU Lesser General Public License for more details.
 You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
+
 import { motion } from 'framer-motion'
-import { ReactNode, useRef } from 'react'
-import styled, { CSSProperties } from 'styled-components'
+import { ReactNode, useEffect, useRef, useState } from 'react'
+import styled from 'styled-components'
 
 import { fadeInOutScaleFast } from '@/animations'
 import ModalContainer from '@/modals/ModalContainer'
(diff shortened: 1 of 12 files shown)


mvaivre:
> I noticed that I cannot open the popup anymore using the keyboard.

While fixing this, I've discovered that in the current version, it's impossible to select an option using the keyboard only. I'll fix it.

mvaivre:
> I noticed that I cannot open the popup anymore using the keyboard.
While fixing this, I was struggling with the autofocus not working (ie. first option not selected on popup opening). I just realised that it's already broken on the `v2.0` branch. May be related to a layout change, I'll come back to it later.

mvaivre:
Update: in order to fully fix accessibility and keyboard navigation, I had to stop using native `select` and `option` elements which are not really made for this kind of use.

You can now navigate through options with the arrow keys or tab key. Autofocus works as expected.

nop33:
Since the token branch is still very WIP, maybe it'd make sense to base this branch on v2.0 instead.

I'll keep working on the tokens branch probably later tomorrow or next week and I'll address the tests.

src/components/Popup.tsx
```diff
@@ -15,52 +15,116 @@ GNU Lesser General Public License for more details.
...
-    <Content onClick={(e) => e.stopPropagation()} role="dialog" {...fadeInOutBottomFast}>
+const minMarginToEdge = 25
+const headerHeight = 50
+
+const Popup = ({ children, onBackgroundClick, title, hookCoordinates }: PopupProps) => {
+  const { height: windowHeight, width: windowWidth } = useWindowSize() // Recompute position on window resize
+
+  const contentRef = useRef<HTMLDivElement>(null)
+
+  const [hookOffset, setHookOffset] = useState<Coordinates | undefined>(undefined)
```
nop33, line 43: nit: I have the feeling that this is the same as:

```tsx
const [hookOffset, setHookOffset] = useState<Coordinates>()
```

nop33 (approved):
Nice stuff man! My functional tests pass ✅ 

One nit thing, I notice that the toggle section (ex: passphrase) now opens a bit "too fast" imo 😅

mvaivre:
> One nit thing, I notice that the toggle section (ex: passphrase) now opens a bit "too fast" imo 😅

Yes I agree :) We'll refine those cosmetic details (more anim variations etc.) at the very end!

PR #502, opened 2023-02-01
Pre release fixes

nop33:
The list of commits describes well the fixes. One thing to note is that I had to rework the gas fields in every send modal because it was a pain in the 🍑 to clear both gas amount & price when closing the toggled section the way it was implemented before.

I will release an rc for internal testing on top of this PR, so that others can help testing while you are reviewing it.


mvaivre:
Oof, well done! 

Functional tests look good, only one little thing related to the toggled sections : errors are still visible after closing and reopening.

Moving on to code review.

__updating...__

src/hooks/useGasSettings.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+You should have received a copy of the GNU Lesser General Public License
+along with the library. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+import { convertAlphToSet, formatAmountForDisplay, MINIMAL_GAS_AMOUNT, MINIMAL_GAS_PRICE } from '@alephium/sdk'
+import { useState } from 'react'
+import { useTranslation } from 'react-i18next'
+
+const useGasSettings = (initialGasAmount?: string, initialGasPrice?: string) => {
```
mvaivre, line 23: Good idea to use a hook for this 👍

mvaivre (approved):
Code looks good!

nop33:
Thank you for your comments @mvaivre ! I fixed both of the small things you mentioned (and some more I discovered through the help of @tdroxler 🙏 !)

In this video I am demonstrating the special handling of the Linux `deb` and `AppImage` files. I am installing the `v1.4.1` of both `deb` and `AppImage`, and I test whether the update system will work properly in both cases, informing the user that there is the `v1.4.2` available to upgrade.

https://user-images.githubusercontent.com/1579899/216323797-7ccb31d5-be2e-4c6c-a11a-5c97802d0206.mp4

Moreover, I fixed @tdroxler 's comments regarding the lock time.

@tdroxler , I have made 2 new (hopefully final) pre-releases for testing:
- v1.4.1 (you can download this one to test the lock time or whatever you want, it has dev tools enabled)
- v1.4.2 (if you want to test the update system, you'd need to mark this as the "latest" release on GitHub before you launch your v1.4.1 app. Please, be mindful of keeping v1.4.2 as the "latest" release for as short time as possible to avoid confusion with users)

mvaivre:
One issue with Network switching:
- If you select "Custom" and save, even though it's using the testnet URLs, you see no more amounts and no refresh button in the header.

nop33:
Wow, what a nice catch! 🦋 Thankfully, if you switch to mainnet and back to testnet, it works...

I can't wait for v2.0 that replaces all this ugly logic through Redux...

nop33:
@mvaivre I fixed the issue. It will massively get simplified with v2.0.

mvaivre (approved):
LGTM! 🚀

PR #511, opened 2023-02-07
Redux migration: pending transactions & saving new addresses

nop33:
Some things might be broken, the addresses migration from context to Redux is still not finished, but I am very close to it.


src/components/AddressBadge.tsx
```diff
@@ -51,17 +52,17 @@ const AddressBadge = ({
 
   if (!address) return null
 
-  return showHashWhenNoLabel && !address.settings.label ? (
+  return showHashWhenNoLabel && !address.label ? (
     <Hash className={className}>
-      {!isPassphraseUsed && address.settings.isMain && !hideStar && <Star>★</Star>}
+      {!isPassphraseUsed && address.isDefault && !hideStar && <Star>★</Star>}
```
mvaivre, line 57: 👍

mvaivre:
Functional testing (quickly, as you mentioned that things are probably broken):
- Pending transactions aren't visible.
- After a transaction was settled, clicking "refresh" doesn't show it.

mvaivre (commented):
Waiting for you to complete the changes before going deep into the review.
But first review : I haven't seen anything crazy! Great job!

nop33:
> Waiting for you to complete the changes before going deep into the review.

That will create a massive PR! I was hopping we can focus on code-review and leave the functional-review once the migration has been completed.

Alternatively, I can create a massive PR that ensures proper functionality of the whole app.

src/storage-utils/addresses.ts
```diff
@@ -0,0 +1,40 @@
...
+        isDefault: address.isDefault,
+        label: address.label,
+        color: address.color
+      }
+    })
+  )
+
+  store.dispatch(newAddressesGenerated(addresses))
+  store.dispatch(syncAddressesData(addresses.map((address) => address.hash)))
+}
```
nop33, line 40: @mvaivre I am curious to hear your thoughts on this new folder called `storage-utils`. My goal with this is to have a place where utility functions that affect both the local storage as well as the Redux store live. It didn't feel good to put them inside `@/persistant-storage`, nor in `@/store` because I didn't want to create deps between them. So I created `@/storage-utils`. I am not 100% convinced with the name. Other names I thought was `@/storage` but that's confusing, next to `@/store` and `@/persistant-storage`.

A second idea would be to restructure the `@/store` directory and instead of just adding `xyzSlice.ts` files, to have a folder structure like so:

image

src/storage-utils/addresses.ts
```diff
@@ -0,0 +1,40 @@
...
+      index: address.index,
+      settings: {
+        isDefault: address.isDefault,
+        label: address.label,
+        color: address.color
+      }
+    })
+  )
+
+  store.dispatch(newAddressesGenerated(addresses))
```
nop33, line 38: I should probably rename this to `newAddressesSaved` since that's what happens.

src/storage-utils/addresses.ts
mvaivre: Thanks for the heads-up. I saw this one yesterday and was not sure about how I felt about it. Today it's a bit clearer :)

- I agree with you on the first point. Those functions should be decoupled from both `store` and `persistent-storage`. They should be one level above in the tree IMO.
- The second idea looks good on paper, but only if those utils functions were used exclusively in the sibling slice found next to them in the tree. It's a bit weird IMO to bury utils functions like `saveNewWallet` in the hierarchy, when they are used globally. 

My proposition would be something like :

```
storage
├── app-state
│   ├── store.ts
│   ├── actions.ts
│   └── slices
│       ├── addressesSlice.ts
│       ├── networkSlice.ts
│       └── ...
├── persistant-store
│   ├── addressesPersitance.ts
│   ├── contactsPersistance.ts
│   └── ...
└── storage-utils
    ├── addressesStorageUtils.ts
    ├── walletStorageUtils.ts
    └── ...
```

Note that I used longer but more explicit file names to help with IDE search :)

mvaivre:
> Alternatively, I can create a massive PR that ensures proper functionality of the whole app.

Honestly, I think I'd rather go for this one 🙃
Ideally, my review should focus more on those new additions or changes you made, but they are sometimes hard to spot because they're mixed up with a lot of already reviewed code & logic ported from the mobile app. 

One thing I would love to help me with the review would be if you could point me to the points you're the most unsure about, so that I can spend my full attention span on those ones (like you did [here](https://github.com/alephium/desktop-wallet/pull/511#discussion_r1099873989) fo instance). Just some 👆 will do! 

I'll of course review the whole thing but I'll focus a bit more on the parts you highlight.

Also, having the full functional PR will also give us more time & room for proper manual testing IMO!

🚀

src/storage-utils/addresses.ts
nop33: I think your proposal is awesome.

nop33:
I agree with your points! I'll point you to the things to pay attention to. But for now my 🧠 doesn't work anymore, I'll do that tmr 😇

src/storage-utils/addresses.ts
nop33: I'll go for this!

image

src/utils/addresses.ts
```diff
@@ -20,15 +20,6 @@ import { Transaction } from '@alephium/sdk/api/explorer'
...
 import { AddressTransaction, PendingTransaction } from '@/types/transactions'
-import { getRandomLabelColor } from '@/utils/colors'
-
-export const sortAddressList = (addresses: AddressRedux[]): AddressRedux[] =>
-  addresses.sort((a, b) => {
-    // Always keep default address to the top of the list
-    if (a.isDefault) return -1
-    if (b.isDefault) return 1
-    return (b.lastUsed ?? 0) - (a.lastUsed ?? 0)
-  })
```
nop33, line 31: One more 🆒 thing about using Redux, is that our state is always sorted properly, so we can ensure that when we fetch addresses to display to the frontend, they are already sorted out! No need for this utility function anymore 💪

src/utils/transactions.ts
```diff
@@ -24,52 +24,18 @@ import {
...
 import { NetworkName } from '@/types/network'
-import { PendingTx, TransactionStatus } from '@/types/transactions'
-
-export type TransactionVariant = Transaction | PendingTx
-type IsTransactionVariant<T extends TransactionVariant> = T extends Transaction
-  ? Transaction
-  : T extends PendingTx
-  ? PendingTx
-  : never
-export type BelongingToAddress<T extends TransactionVariant> = { data: IsTransactionVariant<T>; address: Address }
```
nop33, line 38: I feel so clean we don't need this 💩 anymore!

src/utils/transactions.ts
```diff
@@ -24,52 +24,18 @@ import {
...
-export function sortTransactions(a: HasTimestamp, b: HasTimestamp): number {
-  const delta = b.timestamp - a.timestamp
-
-  // Sent and received in the same block, but will not be in the right order when displaying
-  if (delta === 0) {
-    return -1
-  }
-
-  return delta
-}
```
nop33, line 70: Yep, Redux keep'em sorted, this is not needed! 💪

nop33:
@mvaivre I went through the diff again and wrote comments on the places I want to draw your attention to. I'll do the same with the next PR, and then open a new PR with the fixes and ask you to help me with the functional testing!

src/utils/addresses.ts
mvaivre: So nice.

src/utils/addresses.ts
```diff
@@ -46,7 +37,7 @@ export const selectAddressTransactions = (
...
 }
 
-export const initialAddressSettings = {
-  isDefault: true,
-  color: getRandomLabelColor()
-}
+export const getAvailableBalance = (address: AddressRedux): bigint =>
+  BigInt(address.balance) - BigInt(address.lockedBalance)
+
+export const getName = (address: AddressRedux): string => address.label ?? `${address.hash.substring(0, 10)}...`
```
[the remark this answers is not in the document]
mvaivre: You know me well, that would have been a NIT 🙃 But I'm perfectly okay to keep them short 🤝 
And I would have done the same (ie. not storing this in our store).

src/utils/transactions.ts
mvaivre: Oh yes.

mvaivre (approved):
Looks all good to me! On to the next one 🚀

PR #515, opened 2023-02-08
Redux migration: Goodbye addresses context

nop33:
@mvaivre I did not take the time to properly test everything yet. I only tested that a transaction can be sent from one address to another. But hey, we're almost there! Feel free to have a look if you want. I'll continue tomorrow


src/App.tsx
```diff
@@ -111,6 +115,20 @@ const App = () => {
...
+      } else {
+        clearInterval(interval)
+      }
+    }, 2000)
+
+    return () => {
+      clearInterval(interval)
+    }
+  }, [pendingTxHashes, dispatch])
+
```
nop33, line 131: FYI, I've already updated this in the next branch to use `useInterval`

src/api/addresses.ts
```diff
@@ -42,20 +44,22 @@ export const fetchAddressesData = async (addressHashes: AddressHash[]) => {
       hash: addressHash,
       details,
       transactions,
+      // Type casting needed due to https://github.com/alephium/explorer-backend/issues/427
+      unconfirmedTransactions: unconfirmedTransactions as UnconfirmedTransaction[],
```
nop33, line 48: Soon we won't need this additional type casting.

src/components/IOList.tsx
```diff
@@ -53,14 +50,15 @@ const IOList = ({
...
-  const { getAddress } = useAddressesContext()
   const handleShowAddress = useAddressLinkHandler()
+
   const io = (isOut ? outputs : inputs) as Array<Output | Input> | undefined
 
   if (io && io.length > 0) {
     const isAllCurrentAddress = io.every((o) => o.address === currentAddress)
     const notCurrentAddresses = _(io.filter((o) => o.address !== currentAddress))
       .map((v) => v.address)
+      .filter((v): v is string => v !== undefined)
```
nop33, line 61: some typescript magic to remove `undefined` values so that we don't need to complexify the `map` below.

src/components/WalletSwitcher.tsx
```diff
@@ -71,15 +71,16 @@ const WalletSwitcher = () => {
     if (!switchToWalletName) return
 
     setIsPasswordModalOpen(false)
-    unlockWallet(
-      switchToWalletName,
+    unlockWallet({
+      event: 'switch',
```
nop33, line 75: It makes our state easier to reason about if we know exactly what triggered the wallet unlock (whether it was the user switching wallets after already having one unlocked, or if it was a normal login when no other wallet was unlocked. That's why I introduced the event type (see other comment).

src/contexts/global.tsx
```diff
@@ -16,40 +16,33 @@ You should have received a copy of the GNU Lesser General Public License
...
 import { migrateUserData } from '@/utils/migration'
 
-const deriveAddressesFromIndexesWorker = new Worker(
-  new URL('../workers/deriveAddressesFromIndexes.ts', import.meta.url),
-  { type: 'module' }
-)
-
-export type Client = Exclude<AsyncReturnType<typeof createClient>, undefined>
+interface WalletUnlockProps {
+  event: 'login' | 'switch'
```
nop33, line 37: Here's where I introduce the event type

src/contexts/global.tsx
```diff
@@ -82,75 +74,37 @@ export const GlobalContextProvider: FC<{ overrideContextValue?: PartialDeep<Glob
...
-        deriveAddressesFromIndexesWorker.postMessage({
-          mnemonic: wallet.mnemonic,
-          indexesToDerive: addressesMetadata.map((metadata) => metadata.index)
-        })
+      const payload = {
+        name: walletName,
+        mnemonic: wallet.mnemonic,
+        isPassphraseUsed
       }
+      dispatch(event === 'login' ? walletUnlocked(payload) : walletSwitched(payload))
```
nop33, line 103: And here is where I decide which action to trigger.

src/hooks/useAddressGeneration.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  return {
+    generateAddress,
+    generateAndSaveOneAddressPerGroup,
+    restoreAddressesFromMetadata,
+    discoverAndSaveUsedAddresses
+  }
+}
+
+export default useAddressGeneration
```
nop33, line 194: I am curious about your opinion on this whole file. My idea was to group together all "address generation" functions. The main reason is that all these functions are "expensive" cryptographic operations and utilize web workers to make the experience smoother. I went for a hook for the following reasons:
- Keeps the number of arguments to the functions to the minimum necessary
- Can make use of `useCallback` because some of them are used as deps in `useEffects`

Note that some functions still need optional wallet params (mnemonic, walletName) because they can be used in places where the store hasn't yet been informed about the new wallet (wallet creation/import page).

src/modals/NewAddressModal.tsx
```diff
@@ -50,47 +44,35 @@ interface NewAddressModalProps {
...
 
   useEffect(() => {
-    singleAddress && generateNewAddress()
-  }, [generateNewAddress, singleAddress])
+    if (singleAddress) {
+      const address = generateAddress()
+      setNewAddressData(address)
+      setNewAddressGroup(addressToGroup(address.hash, TOTAL_NUMBER_OF_GROUPS))
+    }
+  }, [generateAddress, singleAddress])
```
nop33, line 62: This is where the address generation function is used as a `useEffect` dep.

src/modals/SendModals/SendModalDeployContract.tsx
```diff
@@ -195,16 +196,18 @@ const DeployContractBuildTxModalContent = ({ data, onSubmit, onCancel }: DeployC
...
+  const data = await signAndSendTransaction(
     txData.fromAddress,
     context.unsignedTxId,
-    context.unsignedTransaction.unsignedTx,
-    context.currentNetwork
+    context.unsignedTransaction.unsignedTx
   )
 
+  // TODO: Should we display a pending transaction when deploying a contract? What would the "toAddress" be in that case?
+  // Since there's no toAddress, we need to rethink our UI.
```
nop33, line 209: In the next branch I am implenenting this by just showing a badge with the text `Smart contract` for the "to" address.

src/modals/SendModals/SendModalScript.tsx
```diff
@@ -168,15 +169,13 @@ const buildTransaction = async (client: Client, txData: ScriptTxData, ctx: TxCon
...
-  const data = await client.signAndSendContractOrScript(
-    txData.fromAddress,
-    ctx.unsignedTxId,
-    ctx.unsignedTransaction.unsignedTx,
-    ctx.currentNetwork
-  )
+  const data = await signAndSendTransaction(txData.fromAddress, ctx.unsignedTxId, ctx.unsignedTransaction.unsignedTx)
+
+  // TODO: Should we display a pending transaction when calling a script? What would the "toAddress" be in that case?
+  // Since there's no toAddress, we need to rethink our UI.
```
nop33, line 178: Same here

src/modals/SendModals/SendModalTransfer.tsx
```diff
@@ -254,7 +263,17 @@ const handleSend = async (client: Client, transactionData: TransferTxData, conte
...
+      store.dispatch(
+        transactionSent({
+          hash: data.txId,
+          fromAddress: fromAddress.hash,
+          toAddress,
+          timestamp: new Date().getTime(),
+          type: 'transfer',
+          lockTime: lockTime?.getTime(),
+          status: 'pending'
+        })
```
nop33, line 275: You might notice the amount is missing, I've added it back in the next branch and fixed some issues with types.

src/pages/UnlockedWallet/AddressDetailsPage.tsx
```diff
@@ -146,23 +146,20 @@ const AddressDetailsPage = () => {
       </DataList>
       <PageH2>{t`Transaction history`}</PageH2>
       <Table minWidth="500px">
-        {pendingTxs
-          .slice(0)
-          .reverse()
-          .map((transaction) => (
```
nop33, line 152: Thanks to redux, we don't have to do this train anymore 🛤️ 🚫

src/pages/UnlockedWallet/AddressesPage/AddressCard.tsx
```diff
@@ -44,23 +46,23 @@ const AddressCard = ({ hash, className }: AddressCardProps) => {
...
   // TODO: Fetch tokens from explorer API and store in Redux
-  const tokens: string[] = address.details.balance !== '0' ? ['ALPH'] : []
+  const tokens: string[] = address.balance !== '0' ? ['ALPH'] : []
 
   const navigateToAddressDetailsPage = (hash: AddressHash) => () => navigate(`/wallet/addresses/${hash}`)
 
   const setAsDefaultAddress = (event: MouseEvent<HTMLDivElement>) => {
-    mainAddress && updateAddressSettings(mainAddress, { ...mainAddress?.settings, isMain: false })
-    updateAddressSettings(address, { ...address.settings, isMain: true })
+    changeDefaultAddress(address, { walletName, mnemonic })
```
nop33, line 65: Once we figure out this "data key", our code of updating encrypted settings will become simpler.

src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
```diff
@@ -77,7 +77,6 @@ const AddressesTabContent = () => {
         </HideEmptyAddressesToggle>
       }
     >
-      {/* TODO: Use sortAddressList(visibleAddresses) when dependency from contexts/addresses is removed */}
```
nop33, line 80: Nope, not needed!

src/store/confirmedTransactionsSlice.ts
```diff
@@ -50,6 +52,8 @@ const confirmedTransactionsSlice = createSlice({
       .addCase(syncAddressesData.fulfilled, addTransactions)
       .addCase(syncAddressTransactionsNextPage.fulfilled, addTransactions)
       .addCase(syncAllAddressesTransactionsNextPage.fulfilled, addTransactions)
+      .addCase(walletLocked, () => initialState)
+      .addCase(walletSwitched, () => initialState)
```
nop33, line 56: This is the reason why I wanted to differentiate between `walletUnlocked` and `walletSwitched`. If I didn't, then we'd need to have here:
```ts
      .addCase(walletLocked, () => initialState)
      .addCase(walletUnlocked, () => initialState)
```

which is quite weird 🤔 why the hell do I need to clean the state when locking AND when unlocking?! The reason is that unlocking can happen when we are already unlocked and we switch a wallet. Thus, I made a new action for clarity.

nop33:
@mvaivre I went through the diff and left comments to attract your attention to the important bits!

src/components/WalletSwitcher.tsx
mvaivre: Yes, it makes total sense.

src/pages/UnlockedWallet/AddressesPage/AddressCard.tsx
mvaivre: 💯

src/App.tsx
mvaivre: Perfect.

src/hooks/useAddressGeneration.tsx
mvaivre: The more I think about this, the more I believe that some cleaning of what we call "utils" could help us a lot.
It's even more true now that we have those `storage-utils`. What makes some functions go into the latter folder vs. the generic utils folder?
Why is `loadContacts` in the generic `utils/contacts.ts` file and not in the `storage-utils`, when it's pretty much focused on loading things from our storage?

Should we maybe abstract things a little bit more? Ie. organise things by [feature / topic rather than technicalities & file types](https://reactjs.org/docs/faq-structure.html) (hooks vs. storage vs...)? 

I'm throwing rough thoughts on the table to start a discussion, I don't have a perfect solution in mind just yet. But I really feel like we could squash and merge some files and make our project structure slimmer and less confusing :)

mvaivre: A rough first tree...

```
features
├── wallet
│   ├── walletCreation.ts
│   ├── walletSwitch.ts
│   └── ...
├── addresses
│   ├── addressGeneration.ts
│   ├── addressSettingsUpdate.ts
│   └── ...
├── transactions
│   ├── transactionSend.ts
│   ├── transactionDisplay.ts // format outputs, get amounts etc.
│   └── ...
├── theme
│   ├── themeSwitch.ts
│   └── ...
└── amounts
    ├── amountsFormating.ts
    └── ...
...
```

src/contexts/global.tsx
```diff
@@ -16,40 +16,33 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
```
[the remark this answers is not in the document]
nop33: Yes, it's not blocking me from continuing with the main tasks of the v2.0, like introducing tokens, layout changes, etc. I'd leave this for another day :)

src/hooks/useAddressGeneration.tsx
nop33: > Is it really necessary to separate our hooks from the other utils functions? We know from their names (use prefix) what they are.

I don't see any benefits in having all hooks grouped together. So no, not necessary.

> The more I think about this, the more I believe that some cleaning of what we call "utils" could help us a lot.
It's even more true now that we have those storage-utils. What makes some functions go into the latter folder vs. the generic utils folder?

I agree on reconsidering what we call "utils". As for the question, the functions that focus on app/persistent storage go to storage-utils. I think that's an easy distinction.

> Why is loadContacts in the generic utils/contacts.ts file and not in the storage-utils, when it's pretty much focused on loading things from our storage?

Because storage-utils was introduced after loadContracts was first introduced and I did not remember that I had created it in the past :)

> But I really feel like we could squash and merge some files and make our project structure slimmer and less confusing :)

Yes!

> A rough first tree...

Would that mean we will also put our components, api, and other files inside the features folder? Or is this only about utils? :/

nop33: I have the feeling that if we go for a features folder structure then everything needs to go into it. Wouldn't it be weird to have function related to wallet creation inside `@/features/wallet/walletGeneration.ts` but the JSX of the screen to be at `@/pages/NewWallet/CreateWalletPage.tsx`?

I wanna put more thought into this, but I also want to keep implementing features of the v2.0 release so that we're not late!

nop33: One thing that is quite clear to me is that I would like things that are general and are used in many places to be grouped by type (think basic components like buttons, I think they should stay at `@/components/buttons`) but I would definitely like a closer coupling of things that are not general and are very related to each other (for example, I don't like that we have `@/hooks/useGasSettings.ts`, `@/modals/SendModals/SendModalTransfer.tsx`, and `@/modals/SendModals/GasSettingsExpandableSection.tsx`)

In your proposal above I see that you grouped features together, but keep in mind that many might be related to more than 1 category, leading to confusion when looking for something.

I think it'd be easier if we had a call about this, every time I am writing a proposal I have to scratch it because I immediately think of drawbacks.

mvaivre: To clarify : my proposal would mean that we (try to) split presentational functions (pages, components, modals) on one side and logic / utils functions on another side. `pages`, `components` and `modals` could be placed into a `ui` parent folder. Those presentational components are making use of the logic placed in the `features` folder. But then, our tree goes deeper, and I don't know if that's something we want. 

Anyway, I wanted to raise the topic now, because I wanted to know if that was something that bothered you in the past. But we can definitely postpone the (potential) changes to later, and schedule a call when we wanna talk more about this.

mvaivre (approved):
Alright, let's move ahead! Great work!

We'll come back to the discussion about the tree structure later.

src/hooks/useAddressGeneration.tsx
nop33: I really like the idea of `features` a lot! I am just not sure whether the presentational stuff should be separate from them. I'll keep working on v2.0 and if I have any good inspiration and ideas I will let you know!

Issue #436, opened 2022-11-22
Numbers display: use tabular version of the font instead of mono font


mvaivre:
See https://rsms.me/inter/#features/tnum

nop33:
Hm, replacing `font-family: monospace` with `font-feature-settings: 'tnum'` does not produce a consistent string length. We need to investigate more how we create this middle ellipsis...

| monospace | tnum |
|--------|-------|
| <img width="399" alt="image" src="https://user-images.githubusercontent.com/1579899/212075659-fba05aaf-e62b-4615-9a61-68d7f1b6ecfd.png"> | <img width="388" alt="image" src="https://user-images.githubusercontent.com/1579899/212075827-d4197f02-b72e-4853-8c75-528dba4e5f26.png"> |

nop33:
the `tnum` is supposed to work for numbers, based on the link you shared:

> Fixed-width numbers are useful for tabular data, where comparing columns across rows is desired.

Maybe that's why it doesn't work with address hashes.

nop33:
@mvaivre Since this only works with numbers, should we keep this issue open?

mvaivre:
I'll close this one in favour of a new one: find a better mono font! 😄

PR #530, opened 2023-02-21
Implement new Send modal UI & support token display in confirmed tx rows and tx details modal

### LINKED ISSUE #72: Tokens support
We need to support displaying tokens (name, amount) in our main wallet, and allow for sending them around.

Features

- [ ] https://github.com/alephium/alephium-js/issues/10

**TBD**

nop33:
I've created a new PR since there are quite many new changes added, let me know if you want me to merge it with the other one.

Closes #72

https://user-images.githubusercontent.com/1579899/220391508-00e7aff9-4576-4503-8a4e-45dedf0980b9.mp4


.vscode/settings.json
```diff
@@ -1,4 +1,8 @@
 {
   "typescript.tsdk": "node_modules/typescript/lib",
-  "typescript.preferences.importModuleSpecifier": "non-relative"
+  "typescript.preferences.importModuleSpecifier": "non-relative",
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
+  },
+  "eslint.validate": ["javascript", "typescript"]
```
nop33, line 7: This is magic. No more linting errors, they are just solved on save.

src/App.tsx
```diff
@@ -53,12 +53,12 @@ const App = () => {
...
     (tx) => tx.address.hash
   )
-  const [settings, network, addressesStatus, isPassphraseUsed, tokens] = useAppSelector((s) => [
+  const [settings, network, addressesStatus, isPassphraseUsed, assetsInfo] = useAppSelector((s) => [
     s.settings,
     s.network,
     s.addresses.status,
     s.activeWallet.isPassphraseUsed,
-    s.tokens
+    s.assetsInfo
```
nop33, line 61: I decided to rename the slice because `asset` is a more accurate name than `token`, since ALPH is not a token, but a coin. Adding the ALPH info (decimals, name, symbol, etc) in the assets slice simplifies our code a lot (this way we don't have to add special cases to treat ALPH separately from tokens. They are both "assets" :)

src/index.css
```diff
@@ -44,6 +44,7 @@ button {
   outline: none;
   border: none;
   background: transparent;
+  font-family: 'Inter', sans-serif;
```
nop33, line 47: For some reason, HTML buttons default to `Arial`...

src/modals/AddressDetailsModal.tsx
```diff
@@ -44,7 +44,7 @@ const AddressDetailsModal = ({ addressHash, onClose }: AddressDetailsModalProps)
     <SideModal
       onClose={onClose}
       label={t('Address details')}
-      width={670}
+      width={800}
```
nop33, line 47: I increased the width to give more space for the tokens

src/App.tsx
mvaivre: Yes, I thought about this as well as I was designing. That's perfect 💯

mvaivre:
Great job! Super sleek. Some things I encountered along the way:

- Transactions are shown twice in the list, even after a full app reload:

image

- "Discover addresses" doesn't work. I get the following error `Uncaught (in promise) Error: Could not generate addresses, active wallet not found`

- This one is on me (it was wrong in the mockup) : the amount input in the Send modal shouldn't be in lighter color when not focused

nop33:
> Transactions are shown twice in the list, even after a full app reload:

Fixed, good catch ✅ 

> NIT: We could make sure that an address without label is not displayed with its hash shown twice:

Good recommendation, done ✅ 

> Let's remove the border around the contact name in the input, it's gonna be lighter, less tight.

✅ 

> the amount input in the Send modal shouldn't be in lighter color when not focused

✅ 

> Hashes: let's try to use the inter font in tabular mode instead of the current mono font, to see how it looks

I remember encountering some problems with this here: https://github.com/alephium/desktop-wallet/issues/436

> "Discover addresses" doesn't work. I get the following error Uncaught (in promise) Error: Could not generate addresses, active wallet not found

Thanks for testing this, I hadn't tested it yet. Fixed ✅

nop33:
@mvaivre lemme know if I can merge this :) I look forward to do that so I can merge master into v2.0 to get the latest updates we pushed to the 1.5.0 release.

mvaivre:
All good! Thanks for the fixes! Merging.

DeafEyeJedi:
Thanks @Ilias for your never ending energy on this one... looking forward to the updates!

PR #537, opened 2023-02-28
Fix switching to system theme settings

### LINKED ISSUE #490: Theme select doesn't switch to "System"
On newest dev branches it's impossible to select "System" as an option for the theme.

nop33:
Closes #490

The bug was introduced when I migrated the general settings to Redux. When listening to OS theme changes, I was dispatching an action that would change the theme settings to either `light` or `dark` and store those settings in localStorage. Instead, I should have separated the theme settings into 2 settings: stored theme settings (`system | light | dark)` and "currently visible theme" (`light | dark`). Now, there are 2 actions:
- `osThemeChangeDetected` (triggered by the OS, can be `light` or `dark`), which sets the theme and doesn't store it in localStorage
- `themeSwitched` (triggered by the user, can be `system`, `light` or `dark`), which sets the theme and stores it in localStorage

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -54,10 +54,10 @@ const App = () => {
   const pendingTxHashes = useAppSelector((s) => selectAddressesPendingTransactions(s, addressHashes)).map(
     (tx) => tx.address.hash
   )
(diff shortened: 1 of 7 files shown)


mvaivre (approved):
Code + tests LGTM! 🥇

PR #538, opened 2023-02-28
Fix wallet listing after creation

### LINKED ISSUE #495: Wallet is not visible in wallet select just after creation
**Describe the bug**
Wallet is not visible in wallet select just after creation.

**To Reproduce**
1. Create a wallet
2. Lock the wallet
3. Select a wallet to unlock
4. New wallet isn't visible in the list

**Expected behavior**
Wallet should be visible in the list.

nop33:
Closes #495


mvaivre (approved):
It works well 👏 LGTM.

PR #557, opened 2023-03-07
[v2.0] Allow user to change active wallet name

nop33:
Now that:
1. we properly store local storage data using IDs as keys,
2. we have consistent classes for storing persistent data, and
3. we use Redux

it became so simple to make changes like this ⭐ 

https://user-images.githubusercontent.com/1579899/223408987-7aa594bc-ba49-49b4-b9e3-dda8d9bdae41.mp4

### DIFF
--- a/src/utils/form-validation.ts
+++ b/src/utils/form-validation.ts
@@ -16,7 +16,7 @@ You should have received a copy of the GNU Lesser General Public License
 along with the library. If not, see <http://www.gnu.org/licenses/>.
 */
 
-import { isAddressValid as _isAddressValid } from '@alephium/sdk'
+import { isAddressValid as isAddressHashValid } from '@alephium/sdk'
 
 import i18n from '@/i18n'
 import { store } from '@/storage/app-state/store'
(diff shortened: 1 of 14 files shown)


MaudSimon:
Could it also allow people to change the password of an active wallet? (If we want to, which is another question, but it is something people are asking about).

nop33:
> Could it also allow people to change the password of an active wallet? (If we want to, which is another question, but it is something people are asking about).

Thanks for bringing it to our attention, I'll create an issue for it.

mvaivre:
👏 yeah!

One first comment after seeing your video: I would expect to see a confirmation snackbar when the name has been updated. (`Wallet name updated to ...`)

src/utils/form-validation.ts
```diff
@@ -45,3 +45,12 @@ export const validateIsContactNameValid = ({ name, id }: Omit<Contact, 'address'
     ? i18n.t('A contact with this name already exists') + `: ${existingContact.address}`
     : true
 }
+
+export const validateIsWalletNameValid = ({ name, previousName }: { name: string; previousName?: string }) =>
```
mvaivre, line 49: Nit: I didn't mention it last time, but why prefixing the name with `validate`? It's a bit redundant to "validate if valid"?

Why not just `isWalletNameValid`? :)

nop33: haha very good point! I did it because I wanted to use the `isAddressValid` function inside a validation function and I could not have 2 functions with the same name 😅 

```ts
export const validateIsAddressValid = (value: string) => isAddressValid(value) || i18n.t('This address is not valid')
```

How do you propose I address (pun intended) this?

nop33: I can do this!

```ts
import { isAddressValid as _isAddressValid } from '@alephium/sdk'
```

nop33:
@mvaivre all comments addressed!

src/utils/form-validation.ts
mvaivre: Or `isAddressHashValid`? :)

PR #559, opened 2023-03-07
[v2.0] Fix bad loading states

### LINKED ISSUE #544: Stuck in loading state when syncAddressesData promise gets rejected
**Describe the bug**
Promise rejections of the `syncAddressesData` thunk are not reported to the user and the app is stuck at loading state.

**To Reproduce**
Use old explorer backend API settings that is missing some of the endpoints or just manually reject the promise.

**Expected behavior**
Get a Snackbar message with the error and stop loading

**Additional context**

We have a few options for approaching this:
1. Wrap every `dispatch(syncAddressesData())` call with a `try/catch` and call `setSnackbarMessage`. I don't like this much because:
   1. it's not very DRY (unless we create a hook for it that can access `setSnackbarMessage` from the `global` context)
   2. it can only be used within hooks are components
1. Move the snackbar message mechanism out of the global context and into Redux

nop33:
~~Tiny PR with a bug I noticed on v2.0~~

Turns out I could bundle some more things related to broken loading states. This PR fixes #544 by implementing the second proposal (moving `snackbarMessage` into Redux. It also implements a mechanism of queuing messages.


src/storage/app-state/slices/snackbarSlice.ts
```diff
@@ -0,0 +1,130 @@
...
+      if (state.messages.length > 0) state.messages.shift()
+    },
+    passwordValidationFailed: (state) =>
+      displayMessageImmediately(state, { text: i18n.t('Invalid password'), type: 'alert' }),
+    copiedToClipboard: (state) => displayMessageImmediately(state, { text: i18n.t('Copied to clipboard!') }),
+    transactionBuildFailed: displayError,
+    transactionSendFailed: displayError,
+    contactStorageFailed: displayError,
+    walletCreationFailed: displayError,
+    transactionsSendSucceeded: (state, action: PayloadAction<number>) => {
```
mvaivre, line 60: I would name the payload param (`{nbOfTransactionsSent: number}`)

src/storage/app-state/slices/snackbarSlice.ts
```diff
@@ -0,0 +1,130 @@
...
+  name: sliceName,
+  initialState,
+  reducers: {
+    snackbarDisplayTimeExpired: (state) => {
+      if (state.messages.length > 0) state.messages.shift()
+    },
+    passwordValidationFailed: (state) =>
+      displayMessageImmediately(state, { text: i18n.t('Invalid password'), type: 'alert' }),
+    copiedToClipboard: (state) => displayMessageImmediately(state, { text: i18n.t('Copied to clipboard!') }),
+    transactionBuildFailed: displayError,
```
mvaivre, line 56: Discussion : I'm really not sure about listing all the uses for snackbar messages in one place, especially since it's still needed to provide the message text.

As a developper who hasn't written the initial code, it's not intuitive to me to find the needed action here: `import { transactionBuildFailed } from '@/storage/app-state/slices/snackbarSlice'`.

I'd rather go for a more generic action like `snackbarMessageShown`.

As always, I'm open to be convinced otherwise! :)

src/storage/app-state/slices/snackbarSlice.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+  },
+  extraReducers(builder) {
+    builder
+      .addCase(syncAddressesData.rejected, (state, action) => {
+        const message = action.payload
+
+        if (message) queueMessage(state, message)
+      })
+      .addCase(apiClientInitSucceeded, (state, action) => {
```
mvaivre, line 76: Linked to previous comment: having side effects like this one scattered around the slices makes it hard to understand what's going on IMO.
What about dispatching a `snackbarMessageShown` action where the `apiClientInitSucceeded` fn is executed (in the network slice)?

nop33: I understand why you would see this as a side effect, but this is truly where Redux shines! One action may trigger several reducers. There's no one-to-one relationship between action and reducer.

> A given action could be handled by all, some, or none of them. This keeps components decoupled from the actual data changes, as one action may affect different parts of the state tree, and there is no need for the component to be aware of this.
https://redux.js.org/faq/actions#is-there-always-a-one-to-one-mapping-between-reducers-and-actions


> This is only true in trivial examples, but it is extremely limiting in real applications. Beginners exposed to this pattern couple reducers and action creators, and fail to realize they're meant to be many-to-many and decoupled.
>
> Many reducers may handle one action. One reducer may handle many actions. Putting them together negates many benefits of how Flux and Redux application scale. This leads to code bloat and unnecessary coupling. You lose the flexibility of reacting to the same action from different places, and your action creators start to act like “setters”, coupled to a specific state shape, thus coupling the components to it as well.
https://github.com/Pitzcarraldo/reduxible/issues/8

src/storage/app-state/actions.ts
```diff
@@ -18,5 +18,15 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
 export const languageChangeStarted = createAction('app/languageChangeStarted')
 export const languageChangeFinished = createAction('app/languageChangeFinished')
+export const copiedToClipboard = createAction('app/copiedToClipboard')
+export const passwordValidationFailed = createAction('app/passwordValidationFailed')
+
+export const transactionBuildFailed = createAction<ErrorMessage>('tx/transactionBuildFailed')
+export const transactionSendFailed = createAction<ErrorMessage>('tx/transactionSendFailed')
+export const transactionsSendSucceeded = createAction<{ nbOfTransactionsSent: number }>('tx/transactionsSendSucceeded')
+
+export const walletCreationFailed = createAction<ErrorMessage>('wallet/walletCreationFailed')
```
mvaivre, line 32: Nit: The functions above those are "actions" in the sense that they're the direct result of a user's click. But the ones I highlighted are something else IMO (maybe even including `passwordValidationFailed`). They are the result of a failure / success of a process, not linked directly to a use click (there might be a network issue, a wrong param entered). 

My point is that I don't know when I should create an action here, or in another slice. Why is `walletCreationFailed` here when there's a `walletDeleted` action in `appSlice`? Or why is `modalClosed` in `appSlice` as opposed to here?

Just my 2 cents with my brain wired a bit differently 😄

src/storage/app-state/slices/snackbarSlice.ts
mvaivre: Of course! I simplify think that we should be careful about creating too many slices in the long run. It's sometimes difficult to know where to declare which action (cf. previous comment).

mvaivre:
I'm interested to know your opinion on where to declare actions. After this, I believe we can merge! 🤝

src/storage/app-state/actions.ts
nop33: I absolute see your points and I think they are valid. I share your concerns of "findability". I'll try to think of a reply tomorrow.

nop33: Trying to think of a proper answer, but since I can't find one that I am happy with, I'll just explain to you my thought process, hoping that we'll manage to find one together.

> why is `modalClosed` in `appSlice` as opposed to here?

When added the `visibleModals` in the `app` slice, it felt natural to me to define the reducer and action directly in `appSlice`:

```ts
const initialState: AppState = {
  visibleModals: []
}

const appSlice = createSlice({
  reducers: {
    modalClosed: (state) => {
      state.visibleModals.pop()
    },
})

export const {
  modalClosed
} = appSlice.actions
```

and then import it in the `ModalContainer` component:

```ts
import { modalClosed, modalOpened } from '@/storage/app-state/slices/appSlice'
```

So far, all felt "good", nothing out of place for me.

> Why is `walletCreationFailed` here when there's a `walletDeleted` action in appSlice?

That's a very good point. `walletDeleted` was added in `appSlice` because that's where the only reducer (at the time of writing) was, so I needed to write less code. It didn't fit in the `activeWallet` slice, because deleting a wallet is not strictly related to having an unlocked wallet, you can also delete it from the home screen. But now, I totally agree that it would better fit wherever `walletCreationFailed` would also be. Since we don't have a slice for explicitly managing wallets, (and since we both agree that we should try to keep the number of slices to the minimal, to avoid making our decisions more difficult on where to put stuff), I proposed that we use the "generic" `src/storage/app-state/actions.ts` file. That might be a good place for the `walletDelete`.

> Nit: The functions above those are "actions" in the sense that they're the direct result of a user's click. But the ones I highlighted are something else IMO (maybe even including passwordValidationFailed). They are the result of a failure / success of a process, not linked directly to a use click (there might be a network issue, a bug, a wrong param entered).

Even though it's true that it's failure / success of processes, an argument could be made that it was the user "clicking" the transaction Send button, or that it was the user "entering" the wrong password and clicking "unlock", etc. Even in the case of language switch and copying to clipboard, there are async processes that might fail (ex: `navigator.clipboard.writeText` is async, `i18next.changeLanguage` is async). But anyways, I feel like we are diverging from the original question:

### What rule do I follow when I want to define an action that will trigger at least one (or more) reducer(s)

Ideas:
1. If the action is strictly related to the slice, put it in the slice. Example: `newAddressesSaved`, `defaultAddressChanged`, or `addressSettingsSaved` shall go in `addressesSlice`.
2. ?

A problem I noticed when defining the actions in the same place as the slice: Circular imports. It might be that slice `A` has a reducer that gets triggered by an action of slice `B` and `B` has a reducer that gets triggered by an action of slice `A`. This creates a circular import and the error in the console is really not helpful to figure that out. A solution method I used was to define the action outside of the slice files (in `src/storage/app-state/actions.ts`). Maybe we should keep that in mind when defining out rules.

nop33: 💡 Idea

Change our file structure to:

```
src/
└── storage/
    └── app-state/
        ├── addresses/
        │   ├── addressSlice.ts
        │   ├── contactsSlice.ts
        │   └── actions.ts
        ├── transactions/
        │   ├── confirmedTransactionsSlice.ts
        │   ├── pendingTransactionsSlice.ts
        │   └── actions.ts
        ├── wallets/
        │   ├── activeWalletSlice.ts
        │   └── actions.ts
        ├── global/
        │   ├── globalSlice.ts # Rename of appSlice
        │   └── actions.ts
        ├── settings/
        │   ├── networkSlice.ts
        │   ├── settingsSlice.ts
        │   └── actions.ts
        └── auth/
            └── actions.ts
```

(not exhaustive list)

1. No more circular imports by having all actions defined in dedicated files (we don't make use of `slice.actions` anymore.
2. Our state starts looking more like our UI features (ex: contacts and addresses are both addresses)
3. It gives us a good hieracy to also start extracting complicated reducer functions out of the slice files (if you look at the `addressesSlice` with all the async thunks and reducers, it starts becoming quite a long and complicated file.

nop33: Assuming the above structure, here's where I would distrubute the actions that started this thread:
- `languageChangeStarted`  -> `settings/actions.ts`
- `languageChangeFinished` -> `settings/actions.ts` 
- `copiedToClipboard` -> `global/actions.ts`
- `passwordValidationFailed` -> `auth/actions.ts`
- `transactionBuildFailed` -> `transactions/actions.ts`
- `transactionSendFailed` -> `transactions/actions.ts`
- `transactionsSendSucceeded` -> `transactions/actions.ts`
- `walletCreationFailed` -> `wallets/actions.ts`
- `walletDeleted` -> `wallets/actions.ts`

nop33: To go a step further, I'd even be more include to fix the bad naming of "actions" and rename the folders to `events` instead. Think of the mindset shift that this affords!

mvaivre: I... love it! This addresses all the questions I asked myself 💯 

- I agree that `events` would be a better fit, but I suggest we do the `actions => events` mapping in our head - I fear that it may be confusing for others theoretical contributors (why are we using fns like `createAction` with events etc.). Let's follow the Redux jargon for now :)

- Does your proposal means that we'll eventually extract all the actions exported from the slices into their own file? This would be awesome IMO, it would make it super easy to understand all available actions (events)! This would clear up space for the rather complex case reducers logic.

- Selectors could be extracted to their own files as well? :)

- For better discoverability, let's use more precise naming for action files? Like `walletActions.ts`, `addressActions` etc. ? 


Love all this! This will make it way easier for other brains to understand what events are propagated in our wallet, and what selectors can be used to fetch state values 🚀

nop33: > Does your proposal means that we'll eventually extract all the actions exported from the slices into their own file? This would be awesome IMO, it would make it super easy to understand all available actions (events)! As you said, this would clear up space for the rather complex case reducers logic.

Yes.

> Selectors could be extracted to their own files as well? :)

Totally.

> For better discoverability, let's use more precise naming for action files? Like walletActions.ts, addressActions etc. ?

💯 

--- 

This will create a big diff do u mind if we merge this one and I do it in the next PR?

mvaivre: > This will create a big diff do u mind if we merge this one and I do it in the next PR?

Sure, let's do this! I let you merge it 🚀

mvaivre (approved):
Great work! 🌞

PR #572, opened 2023-03-08
[v2.0] Private key export

### LINKED ISSUE #484: Allow private key export in devtools settings
> Add a button somewhere to support it, as it’s convenient for development.
> Metamask supports private key exporting, same for Argent-X.

> Q: Since every address of the wallet has its own private key, do we want to let the user export ANY private key or just the one of the initial address (address index 0)?
> A: Yes, we will support private key exporting for every addresses. For example, if I have 4 addresses, I could export any private key of the 4 addresses and then deploy smart contracts with any of them

nop33:
Closes #484

https://user-images.githubusercontent.com/1579899/223760393-8f0bec7d-f608-4252-9fbc-1b03a890d8c5.mp4

### DIFF
--- a/locales/de-DE/translation.json
+++ b/locales/de-DE/translation.json
@@ -306,7 +306,7 @@
   "Addresses": "Adressen",
   "Developer tools": "Developer tools",
   "Enable developer tools": "Enable developer tools",
-  "Deploy and call smart contracts": "Deploy and call smart contracts",
+  "Features for developers only": "Features for developers only",
   "Transfers": "Transfers",
   "Lock time must be in the future.": "Lock time must be in the future.",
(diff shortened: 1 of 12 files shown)


polarker:
The demo looks great to me. I will leave the strict review to Mika :)

mvaivre:
Good job 🤙 

Let's deactivate the shortcut to copy an address hash to avoid confusion & mistakes, WDYT? :

image

nop33:
> Let's deactivate the shortcut to copy an address hash to avoid confusion & mistakes, WDYT? :

Done!

mvaivre (approved):
LGTM! 🥇

mvaivre (changes requested):
image

A last finding: let's just make sure to hide this if no wallet is unlocked.

nop33:
@mvaivre I removed the whole DevTools section because it doesn't make sense to have it when the wallet is locked (you clicking the smart contract buttons wouldn't open any send modal).

mvaivre (approved):
🎸

PR #582, opened 2023-03-14
[v2.0] Reorganize state

nop33:
Making a new PR because I made 1'518 changes.

This PR is a result of [this discussion](https://github.com/alephium/desktop-wallet/pull/559#discussion_r1130772609).

Decisions I took:
1. I renamed the `appSlice` to `globalSlice`
1. I removed the `sliceName` variable from each slice, I find it useless now that the slice components has been split around in multiple files
2. I extracted adapters and selectors in their own files. The reason is that once I extracted the address actions into their own file, I wanted to include in the same file the async thunks (which are also actions). Some of those thunks were using address selectors. Those address selectors were defined in the slice file, through the adapter object. This created a circular import 😵 💫 . Separating the adapter from the slice file solved this problem.
3. I decided to merge the contact actions/selectors together with the addresses actions/selectors in files called `addresses[Actions|Selectors].ts`. I think those methods belong well together, even though they refer to a different slice. It shouldn't make a difference, however, to the developer who wants to import those functions. Let's see if this approach is the correct one or if it creates confusion.
4. I moved the `priceApiSlice.ts` into the `assets` folder, since imo it's tightly coupled to the "assets" feature.
5. I spent quite some time thinking whether I should separate the actions of `networkSlice` and `settingsSlice` or put them together in `settingsActions`. I went for the former. It didn't feel right to have `apiClientInitSucceeded` in the `settingsActions.ts`. It feels more natural to see it in `networkActions.ts`.

The final result looks like this. I'm quite happy with it 💛 

imageimage


mvaivre:
This looks indeed much better than before, and I'm also happy with it 💙 
After thinking a bit more about this, I don't know if I'm 100% sure about grouping everything in one parent folder called "storage", as managing events and side effect is not only limited to "storing things" - it's also "reacting to things". I would personnally keep the separation between `state` & `storage` (so 2 different parent folders). 
But as I'm not a reviewer, I'll keep my mouth shut 🙃 😆

nop33:
shit, I forgot to assign you, my bad!

nop33:
> I would personnally keep the separation between state & storage (so 2 different parent folders).

Then we have the same problem again: Where to place functions that update both storage and state.

nop33:
I think state and storage are very close related: They are both the place where we "save" information, either long term (storage) or short term (state).

mvaivre:
Of course, let's do this in another PR!

mvaivre (approved):

PR #585, opened 2023-03-14
[v2.0] Improve display of unknown tokens throughout the app

### LINKED ISSUE #556: AssetList: allow to show/hide unknown tokens
We should :

1) Display unknown tokens properly (use a hash instead of a name, use no ticker etc.)
2) Allow to show or hide unknown token by default. It should probably be shown by default(?)

nop33:
Closes #556

### Asset list

- Instead of a logo, display a `?`
- Instead of name, display `Unknown token`
- Instead of symbol, display copy-able token hash
- Instead of symbol next to amount, display nothing, and show the text `Raw amount` below it

image

### Transaction list
- Instead of logo and symbol, simply display the number of unknown tokens (should we add a tooltip that says `5 unknown tokens`?)
 
image

### Transaction details modal

- do not display amounts of unknown tokens in the header and instead list then in a separate section in the bottom
- instead of token symbol next to amount, show copy-able token hash

image

### Address cards

- Instead of symbol, follow similar approach with transaction list

image

### Asset select

image

### DIFF
--- a/locales/en-US/translation.json
+++ b/locales/en-US/translation.json
@@ -354,5 +354,9 @@
   "Change wallet name": "Change wallet name",
   "Could not save new wallet name.": "Could not save new wallet name.",
   "Wallet name updated to: {{ newWalletName }}": "Wallet name updated to: {{ newWalletName }}",
-  "Shortcuts": "Shortcuts"
+  "Shortcuts": "Shortcuts",
+  "Raw amount": "Raw amount",
+  "Unknown token": "Unknown token",
+  "Unknown tokens": "Unknown tokens",
+  "Copy token hash": "Copy token hash"
 }
--- a/src/components/AddressBadge.tsx
+++ b/src/components/AddressBadge.tsx
@@ -20,10 +20,10 @@ import { ComponentPropsWithoutRef } from 'react'
 import { useTranslation } from 'react-i18next'
 import styled, { css } from 'styled-components'
 
-import AddressEllipsed from '@/components/AddressEllipsed'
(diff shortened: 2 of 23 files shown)


mvaivre:
Awesome 🚀 

> Instead of logo and symbol, simply display the number of unknown tokens (should we add a tooltip that says 5 unknown tokens?)

After trying this on the explorer, I'm leaning towards showing only the symbols (remove the asset tickers) with the hash shown in a tooltip. WDYT?

nop33:
> After trying this on the explorer, I'm leaning towards showing only the symbols (remove the asset tickers) with the hash shown in a tooltip. WDYT?

Happy to implement that! But do you like the approach of the `(+5)` for unknown tokens?

mvaivre:
> Happy to implement that! But do you like the approach of the (+5) for unknown tokens?

That's a good question. I think that I'd show the 5 "?" logos. It makes it easier to judge the quantity of assets exchanged while you scroll quickly + I don't think that it's be frequent that 1 TX moves that many assets at the same time, but I may be wrong.

nop33:
Hm, ok, I'll go with that.

mvaivre:
Thanks for all the tries :) My favorites are... the ghost (a touch of fun 👌), the "?" and the circle. 
Your favorite makes me feel like an icon isn't loading properly 😅

nop33:
Fair enough! I can update the PR tmr to add the ghost if you agree!

nop33:
@mvaivre Ready for review again! :)

mvaivre (approved):
LGTM!

Issue #584, opened 2023-03-14
Improve transaction row info display


nop33:
<img width="824" alt="image" src="https://user-images.githubusercontent.com/1579899/225038152-311d4b57-3cc0-47ab-918e-f20cf00662df.png">

nop33:
Feedback from @mvaivre:

> We need to work a bit on the transfers list UI to give it some more space! I like how it looks on the explorer now, we could take inspiration from there. Here's what I have in mind :
>
> - Reduce length of ellipsed hashes.
> - If an address is in the contact list, use the contact name.
> - Format amounts the same way as in the explorer (emphasis put on amounts rather than tickers by using different colours and font-weights. Don't change decimals opacity).

nop33:
All the above are done in the latest PR: https://github.com/alephium/desktop-wallet/pull/593

nop33:
Closing since it's merged in v2.0!

PR #588, opened 2023-03-20
Transfers page

nop33:
Damn, that turned out to be bigger than I expected. Here's the gist:
1. I create a new `MultiSelect` component from scratch because I lost many hours trying to adapt the `Select` to work for both single and multi select. I believe our `Select` component could use some love in the future.
2. I implemented the `TransfersPage`, following the mockups
3. I could use some help with using Framer, I couldn't yet figure out how to hide the "Receive/Send" buttons on scroll


src/components/Inputs/AddressSelect.tsx
```diff
@@ -137,14 +136,7 @@ function AddressSelect({
...
               const address = addresses.find((address) => address.hash === option.value)
-              if (!address) return
-
-              return (
-                <SelectOptionItemContent
-                  ContentLeft={<AddressBadgeStyled addressHash={address.hash} showHashWhenNoLabel disableA11y />}
-                  ContentRight={<AmountStyled value={BigInt(address.balance)} fadeDecimals />}
-                />
-              )
+              if (address) return <SelectOptionAddress address={address} />
```
nop33, line 139: Since we have more than 1 place where the address select option is useful (address select modal in Send modal and address select modal in Transfers page filtering), I extracted this to a new component. I did the same with the asset select option.

src/components/Inputs/MultiSelect.tsx
```diff
@@ -0,0 +1,169 @@
...
+        ))}
+      </Options>
+    </Popup>
+  )
+}
+
+export default MultiSelect
+
+const SelectedValue = styled.div`
+  ${inputDefaultStyle(true, true, true)};
```
nop33, line 146: I would like at some point to review our approach to inputs in general: text areas, input fields, select fields, etc. I feel that it could be done simpler.

src/components/TransactionList.tsx
```diff
@@ -28,70 +28,108 @@ import { useAppDispatch, useAppSelector } from '@/hooks/redux'
...
-  const showAllTransactionsLoadedMsg = singleAddress ? address?.allTransactionPagesLoaded : allTransactionsLoaded
+  const showSkeletonLoading = isLoading && !filteredConfirmedTxs.length && !pendingTxs.length
+
+  // TODO: How do we handle paging when addresses filtering changes? We need to keep track of the loaded page for every
+  // combination of selected addresses and in general all filtering criteria. That sounds very complex. Is there a
+  // simpler way? What if no matter what the filtering criteria are, we always:
+  // 1. fetch the next page of ALL addresses
+  // 2. check to see if we received any results
+  // 3. filter the results
+  // 4. if the filter results are 0 but the results are not 0, fetch again
```
nop33, line 97: Drawing your attention on this one to hear your thoughts.

src/components/TransactionalInfo.tsx
```diff
@@ -60,7 +59,7 @@ const TransactionalInfo = ({
   const { addressHash: addressHashParam = '' } = useParams<{ addressHash: AddressHash }>()
   const addressHash = addressHashProp ?? addressHashParam
   const address = useAppSelector((state) => selectAddressByHash(state, addressHash))
-  const { assets, direction, outputs, lockTime, infoType } = useTransactionInfo(tx, addressHash, showInternalInflows)
+  const { assets, direction, outputs, lockTime, infoType } = getTransactionInfo(tx, showInternalInflows)
```
nop33, line 62: I changed this hook to a simple function so that it can be used in the `applyFilters` method just above (in `TransactionList.tsx`.

src/modals/AddressDetailsModal.tsx
```diff
@@ -100,44 +96,20 @@ const AddressDetailsModal = ({ addressHash, onClose }: AddressDetailsModalProps)
...
-              onClick={() => setIsSendModalOpen(true)}
-              Icon={ArrowUp}
-              iconColor={theme.global.accent}
-            >
-              <ButtonText>{t('Send')}</ButtonText>
-            </ShortcutButton>
-            <ShortcutButton transparent borderless onClick={() => setIsAddressOptionsModalOpen(true)} Icon={Settings}>
-              <ButtonText>{t('Settings')}</ButtonText>
-            </ShortcutButton>
+            <ShortcutButtons receive send addressSettings addressHash={address.hash} />
```
nop33, line 99: The same buttons were repeated in 3 different places, so I extracted them all to one component.

src/pages/UnlockedWallet/TransfersPage/index.tsx
```diff
@@ -16,19 +16,206 @@ You should have received a copy of the GNU Lesser General Public License
...
+          <Button role="secondary" short onClick={resetFilters}>
+            {t('Reset filters')}
+          </Button>
+        </Buttons>
+      </Filters>
+      <UnlockedWalletPanel top>
+        <TransactionList
+          addressHashes={map(selectedAddresses, 'hash')}
+          directions={map(selectedDirections, 'value')}
+          assetIds={map(selectedAssets, 'id')}
```
nop33, line 154: Instead of always redefining the simple function that creates an array of a specific property of an objects array, I discovered that `lodash` provides the `map` function!

src/storage/addresses/addressesActions.ts
```diff
@@ -93,42 +93,42 @@ export const syncAddressTransactionsNextPage = createAsyncThunk(
...
     dispatch(loadingStarted())
 
     const state = getState() as RootState
-    const addresses = selectAllAddresses(state)
-    let nextPage = state.addresses.transactionsPageLoaded
+    const addresses = selectAddresses(state, addressHashes)
+
+    let nextPageToLoad = nextPage
     let newTransactionsFound = false
-    let allTransactionsLoaded = state.addresses.allTransactionsLoaded
```
nop33, line 105: I removed `allTransactionsLoaded` from the addresses state because it doesn't bring any value anymore. The pagination state is handled completely on the Transfers page now.

src/storage/global/snackbarSlice.ts
```diff
@@ -124,6 +125,12 @@ const snackbarSlice = createSlice({
         })
       )
       .addCase(walletNameStorageFailed, displayError)
+      .addCase(csvGenerationStarted, (state) =>
+        displayMessageImmediately(state, {
+          text: i18n.t('Your CSV is being compiled in the background'),
+          duration: -1
+        })
+      )
```
nop33, line 133: Just some pre-work on the CSV feature, nothing implemented yet :)

src/components/Inputs/MultiSelect.tsx
mvaivre: Agreed 💯

nop33:
> Having a lot of assets in the filter will push the rest of the toolbar out of the window :) 

Thanks, I didn't add a `min-width: 0` on a flexbox item. I fixed it.

> If possible, we should cut the label and add a (+X) suffix... maybe we can reuse the logic you built for the Addresses page?

I'll have a look, but I think I'd like to rework that logic and come up with a different approach if we were to use it here as well. That one was just a prototype (it doesn't work when making the window larger, only when making it smaller.

> If you click on "Load more" then open a filter that a you have modified before, nothing in the multi-select is selected anymore.

Yeah, the "load more" logic needs some decisions: https://github.com/alephium/desktop-wallet/pull/588/files#r1142989469. But apart from that you're right, clicking "Load more" resets the filters. I'm having a look into that...

> We need to work a bit on the transfers list UI to give it some more space! I like how it looks on the explorer now, we could take inspiration from there. Here's what I have in mind :

I had created a separate issue for that in the past, I'll put your feedback there and work on it in a following PR!

https://github.com/alephium/desktop-wallet/issues/584

> Maybe add a header to the transfer list (same as in the explorer) to better separate the table from the filter toolbar (labels would be "Direction", "Assets", "From", "To", "Amounts"). WDYT?

Hm, I think the table is quite self-explanatory... I don't think it's needed to add table headers. The row reads quite easily


> Let's try making the "Receive / Send" actions float in the middle of the content section rather than on the right?

👍 

> NIT: There's no loading animation:

It's added in the last PR :)

nop33:
@mvaivre, requesting a new round of review 🙏

src/components/TransactionList.tsx
mvaivre: This sounds good to me - in fact, the only solution I see - until we have a more complex endpoint which would allow more query params. :)

mvaivre:
Some more little things:

**Asset selection modal:**
- Unselect all doesn't work
- Let's hide the amounts there 

Other:
- The directions options in the direction modal don't seem to be translated.

src/storage/addresses/addressesActions.ts
mvaivre: Nice! This make me think: let's rename the "Transactions" widget in the Dashboard page into "Latest transactions"?

mvaivre:
I gave a go at hiding the "shortcut buttons" on scroll [here](https://github.com/alephium/desktop-wallet/commit/c30d10103172120af0278ad31e0d8bbfda00af3d). Unfortunately, the use of `useScrollContext` provokes unwanted re-renders, even though I make sure to update the scroll direction only when necessary. 

I can't find how to prevent this for now, it's probably something obvious though : I need a break.

src/pages/UnlockedWallet/TransfersPage/index.tsx
```diff
@@ -16,19 +16,228 @@ You should have received a copy of the GNU Lesser General Public License
...
+    <UnlockedWalletPage
+      title={t('Transfers')}
+      subtitle={t('Browse and download your transaction history. Execute new transfers easily.')}
+      isInfoMessageVisible={!infoMessageClosed}
+      closeInfoMessage={closeInfoMessage}
+      infoMessageLink={links.faq}
+      infoMessage={t('You have questions about transfers ? Click here!')}
+      className={className}
+    >
+      <Filters>
```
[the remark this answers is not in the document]
nop33: I'll look into it!

src/components/TransactionList.tsx
nop33: I'll tackle this in a follow-up PR if that's ok

nop33:
This translation puzzle took me half of the day to solve it, but I've found a very elegant solution! I would like to apply it to a new PR, if that's ok, cause it touches many lines! I think you'll like it A LOT.

src/pages/UnlockedWallet/TransfersPage/index.tsx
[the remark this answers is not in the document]
nop33: ✅ done!

nop33:
> These values don't trigger React renders.

I read a bit on how they work! Since these values are objects and not primitives, changing the property of the object (with `value.set(x)`, the reference to the object hasn't changed, so no re-render!

nop33:
@mvaivre thank you a lot for your guidance! I think I came up with a very elegant solution! I basically implemented what I mentioned above: store a Motion Value instead of the whole ScrollState in our scroll context. In addition, I added the direction of scroll as well inside the context. I did some debugging and our AppHeader component re-renders got reduced from 840 for a single page scroll down to 20! That's massive! Here's [the commit](https://github.com/alephium/desktop-wallet/pull/593/commits/e023a2d8fd3b2409a42eaee5b9577ad6ed538a3f) (it's part of the next PR, sorry!)

mvaivre (approved):
Alright, thanks for the fixes and improvements! Let's merge this one and focus on the follow up PR.

PR #593, opened 2023-03-22
[v2.0] Addresses page

nop33:


src/components/Ellipsed.tsx
nop33, line 1: The changes in this file make use of a new simple hook I created (`useWindowResize`) and simplify the code massively. No changes in functionality.

src/pages/UnlockedWallet/AddressesPage/AddressGridRow.tsx
```diff
@@ -0,0 +1,203 @@
...
+      newArray.push(`+${assetsWithBalance.length - assetSymbolTexts.length}`)
+      setAssetSymbolTexts(newArray)
+    }
+  }, [assetSymbolTexts, assetsWithBalance.length])
+
+  useEffect(() => {
+    ensureAssetsDisplayedInOnlyOneRow()
+  }, [ensureAssetsDisplayedInOnlyOneRow])
+
+  useWindowResize(ensureAssetsDisplayedInOnlyOneRow)
```
nop33, line 91: This piece of code is responsible for this. It works when resizing the window as well, but only when scaling down 😅 

image

src/pages/UnlockedWallet/AddressesPage/AddressesTabContent.tsx
```diff
@@ -95,3 +119,48 @@ const ToggleText = styled.div`
...
+const HeaderMiddle = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  gap: 20px;
+  flex: 1;
+`
+
+const TableGrid = styled(Box)`
+  contain: paint; // This is amazing. It replaces "overflow: hidden". Using "overflow" on this prevents us from having a sticky table header.
```
nop33, line 132: This took me many hours to figure out (look at the git history and you'll understand!)

mvaivre: Oh wow! The number of times I was struggling with avoid the use of `overflow: hidden` to allow for "stickiness" 🙃.. <3

src/components/SkeletonLoader.tsx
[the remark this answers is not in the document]
nop33: The skeleton loaders were inspired by Notion's latest update 😃

public/electron.js
```diff
@@ -151,7 +151,7 @@ function createWindow() {
   mainWindow = new BrowserWindow({
     width: 1200,
     height: 800,
-    minWidth: 1000,
+    minWidth: 1200,
```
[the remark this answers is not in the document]
nop33: Making the wallet responsive with smaller than 1200 screens could be a nice task for a following release, WDYT?

Analytics from our users might help here...

src/pages/UnlockedWallet/AddressesPage/AddressGridRow.tsx
nop33: it now works on all window resize! I wonder if you could give me a hand with framer motion so that we don't have that ugly glitch artifact.

public/electron.js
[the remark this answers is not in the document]
mvaivre: Ledger Live seems to be using a min width of 1020px so I decided to follow their lead, even though I found this to be a "lazy" strategy 🙃.
When our wallet will be used alongside dApps, it's gonna be useful to be able to reduce its size, so yes, working on a nice responsive layout should be considered for a future release. I'll create a task.

src/pages/UnlockedWallet/AddressesPage/AddressGridRow.tsx
nop33: Another idea I had in mind is to create an invisible clone of the `AssetSymbols` component instance and position it `absolute` so that it's over the visible one. First do the "add/remove" operation on the invisible component and after a short timeout, based on the content of the invisible container perform the same operation to the visible one. Hard to explain 😅

src/components/SkeletonLoader.tsx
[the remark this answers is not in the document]
mvaivre: I really like it ✨ !

src/pages/UnlockedWallet/AddressesPage/AddressGridRow.tsx
mvaivre: Soooo... After playing a bit with this, I decided to drastically simplify it 🙃 

New design is also more consistent with other of our UIs (tx list, explorer). I really hope you don't mind!
We will probably be able to reuse your logic somewhere else.

src/pages/UnlockedWallet/AddressesPage/AddressGridRow.tsx
```diff
@@ -59,17 +60,32 @@ const AddressGridRow = ({ addressHash, className }: AddressGridRowProps) => {
...
-  const [nbOfHiddenSymbols, setNbOfHiddenSymbols] = useState(0)
 
   const assetsCellRef = useRef<HTMLDivElement>(null)
 
   if (!address) return null
 
   const fiatBalance = calculateAmountWorth(BigInt(address.balance), price ?? 0)
 
+  const renderAssetLogos = () => {
+    const [displayedAssets, hiddenAssets] = chunk(assetsWithBalance, maxDisplayedAssets)
```
nop33, line 71: This is problematic because `assetsWithBalance` might contain in various places unknown tokens where the symbol is `undefined`. I'll update the code to fix it!

nop33, line 71: This way we are only taking the first 2 chunks and we're losing the rest! I'll update the code to fix that

mvaivre: Arh, didn't notice that this fn is clever enough to create an infinite number of chunks if needed, and I only tested with a length of less than max * 2. Sorry about that.

mvaivre: Haha I saw. I'm okay with that :)

mvaivre:
This looks better and better, delicious. 🍬 

Still a couple of issues though:

- After I click on "load more" a couple of times, the asset previously selected in the asset filter modal are deselected.
- The addresses filter seems to cumulate (linked to loading more TXs it seems):

image

⬆️ I have 4 addresses, but it shows 8 selected.

- We should have a placeholder when no TXs correspond to a set of filter:

image

mvaivre (approved):
Pre approval! Code looks really good 👏  If you can take a look at the couple of issues I found, after that, feel free to merge 🚀

nop33:
> The "copy address" shortcut on address hashes pops up above the action shortcuts :D

✅  Haha, what a nice catch! Fixed! 

> By messing around with the transfers page a lot (adding, removing filters, loading more TXs...) I end up at a point where I have 2 scroll contexts, and I can scroll the whole content (including the navigation sidebar) :

LOL! How the heck did you do that 😆 I've been playing forcefully with the transfers page, resizing the window, adding filters, removing, and scrolling, can't reproduce it though! I'll keep an eye on it, hopefully it appears! If it re-appears on your side, please do some element inspection to provide more feedback! It looks like the second scrollbar is the native scrollbar (not the one created by the `<Scrollbar>` component (which is only instantiated once on that page.

> Oof and that's a tricky one, I couldn't reproduce it so far. It happened as soon as I opened the Addresses page. Resizing the window made the glitch go away 👻

Oh, I have some ideas of what it is: It's probably because we changed our mono font so the calculation of the middle ellipsis is not working properly. Since the font is not native to the system and it needs to be loaded separately, it might cause some problems when calculating the character width. Or maybe it's a rare corner-case of the ellipsis mechanism even without the font change.

nop33:
I'll merge even though I didn't manage to reproduce your last 2 comments. I created an issue for one of them.

PR #597, opened 2023-03-28
[v2.0] 👶 Add additional lock time confirm modal

### LINKED ISSUE #548: More explicit warning in the Review screen when setting a Lock Time
Feedback from the community: "Lets say I want to send coins with lock time.Then I send not a small amount with a locktime until 2023 but instead 2023 I write 2032 and I dont check it again, the coins are gone for the next 10 years. 

**Describe the solution you'd like**
"So, if you set a lock time, I would think it would be better to implement a warning with "please check the correct date you want to lock your coins and press okay"."

**Additional context**
The review screen is easy to mindfully click through which can make it easy for inattentive users to make mistakes.

nop33:
Closes #548


mvaivre (approved):
Very nice! 👏

PR #600, opened 2023-03-29
Capture analytics

### LINKED ISSUE #477: Add anonymised interaction collection to monitor wallet use
**First use case : collect how many addresses our users have.**

Other potential use cases : 

- At what frequency is the wallet opened?
- How frequently is the main address changed?
- How many times are the advanced actions used?
- How frequently is the scan to import feature used?
- How many wallets?
- ...

nop33:
Closes #477 

In this PR I install the PostHog JS SDK and use it to track several properties of our users as well as actions they perform in the wallet.

## Backend

I decided to experiment with PostHog Cloud. It has a very generous [free tier](https://posthog.com/pricing). I created an account with my Alephium email address. It was super easy to start a project, great experience. To connect the desktop wallet, I created an `.env` file, as described by the [React docs](https://posthog.com/docs/libraries/react). @mvaivre I invited you to join the server, you can find the API keys there (it's ok to publish them anyway, but just to be save I didn't commit them).

We can always decide to [self-host](https://posthog.com/docs/self-host).

## Identifying users

As discussed in https://github.com/alephium/desktop-wallet/issues/477#issuecomment-1403309524, the app creates an analytics ID if it cannot find one and stores it locally. It uses this ID to identify the user.

### Privacy

I disabled all settings related to identifiable information that I could find:
- Config options: https://github.com/alephium/desktop-wallet/blob/posthog/src/contexts/analytics.tsx#L28-L34
- PostHog server
   image
   image
   image

## User properties tracked
- Number of wallets
- Theme preference
- Dev tools preference
- Lock time preference
- Language preference
- Password requirement preference

## Events tracked
- Toggled theme
- Locked wallet
- Deleted wallet
- New wallet created
- New wallet imported
- Receive button clicked
- Send button clicked
- Wallet settings button clicked
- Address settings button clicked
- Wallet unlocked (including data for number of addresses and contacts)
- Wallet switched (including data for number of addresses and contacts)
- Changed address settings
- Changed default address
- Swept address assets
- Saved new contact
- Edited contact
- Exported CSV
- New address created
- One address per group generated
- Deployed smart contract
- Called smart contract
- Sent transaction (including data for number of tokens and whether lock time is set)
- Enabled dev tools
- Copied address private key
- Changed wallet name
- Enabled password requirement
- Disabled password requirement
- Changed language
- Changed wallet lock time
- Changed theme
- Changed network
- Saved custom network settings
- Generated one address per group on wallet creation
- Refreshed data
- Advanced operation to generate one address per group clicked
- Advanced operation to discover addresses clicked
- Advanced operation to consolidate UTXOs clicked
- Advanced operation to share ideas clicked

## Questions/future work

- [ ] Do we need consent from users? Based on my research, since we are not capturing any private info, we don't have to.
- [ ] Check what Ledger is doing.


DeafEyeJedi:
Excellent idea actually!!!

mvaivre (commented):
Implementation looks super clean and simple, I like it. ✨ 

Tested the implementation and their dashboard => Everything's seems to be working perfectly! 👏 

- Could we make sure NOT to send analytics when in dev mode?
- Let's add the mentioned "Analytics" toggle (to be the last item in the "General" settings list)?
- Let's monitor the data collection once we start testing v2.0 internally to make sure everything's is setup properly.

src/storage/settings/settingsPersistentStorage.ts
```diff
@@ -49,7 +49,8 @@ export const defaultSettings: Settings = {
     discreetMode: false,
     passwordRequirement: false,
     language: undefined,
-    devTools: false
+    devTools: false,
+    analytics: true
```
nop33, line 53: @mvaivre I wonder what your thoughts on this are. Do we enable analytics by default?

I did a fresh install of Ledger Live on my computer. It was enabled by default and there was no consent asked.

Based on GDPR, we [MUST acquire unambiguous consent](https://posthog.com/docs/privacy/gdpr-compliance#you-must-acquire-unambiguous-consent). I am not sure whether that only applies when we capture [personal data (or PII - Personal Identifiable Information)](https://posthog.com/blog/what-is-personal-data-pii), or if it applies in all cases. [This suggests](https://posthog.com/docs/privacy/gdpr-compliance#step-4-configure-consent) that only if we collect personal data do we need to ask for consent, but it's not 100% clear. To the best of my research and knowledge, I have configured out analytics to not capture personal data.

> If you require robust GDPR compliance, we recommend using [PostHog Cloud EU](https://posthog.com/eu) – a managed version of PostHog that's hosted on servers based in Frankfurt.

✅ we do that already.

nop33:
> Could we make sure NOT to send analytics when in dev mode?

✅  done!

image

> Let's add the mentioned "Analytics" toggle (to be the last item in the "General" settings list)?

✅  done! (commit pushed in this PR)

> Let's monitor the data collection once we start testing v2.0 internally to make sure everything's is setup properly.

🆗

src/storage/settings/settingsPersistentStorage.ts
nop33: @mvaivre Wait, I didn't get your feedback here 😅

mvaivre: We don't collect any personal data & ledger activates this by default => I think we're good to go.

Now, if we want to make extra sure we're fine, we could add a consent box during the onboarding (when creating or importing a wallet for the first time), at the very last step with the confettis. The "let's go" button could be deactivated until the user accept the condition. Should we open an issue and potentially plan this for a following release?

nop33: > The "let's go" button could be deactivated until the user accept the condition.

You mean the user wouldn't be able to use the wallet without enabling analytics? I guess this is not what you mean cause that sounds bad.

> Should we open an issue and potentially plan this for a following release?

Let's do that.

PR #602, opened 2023-04-03
Fix WalletConnect

nop33:
This massive PR fixes bugs that I discovered while testing WalletConnect using the DEX and NFT projects. I believe there is more stuff to be solved, but I decided to start the reviewing process before it grows even further. I think we could start testing and making a list of the remaining things to fix.

This is the procedure I used for testing:

## Testing with NFT project

1. Get a local copy of the NFT project and use the `update-packages` branch (until https://github.com/alephium/alephium-nft/pull/8 is merged) or afterwards the `nft-standard` branch (until https://github.com/alephium/desktop-wallet/pull/602 is merged) or afterwards `master`.
2. Follow the NFT project readme instructions
3. Connect the desktop wallet to the same devnet as the NFT project and restore the [devnet mnemonic](https://github.com/alephium/alephium-stack/blob/master/devnet/devnet.conf#L3):
   ```
   node: http://localhost:22973
   explorer: http://localhost:9090
   ```

## Testing with the DEX project

1. Get a local copy of the DEX project and use the `test-walletconnect-by-old-dex` branch (until https://github.com/alephium/alephium-dex/pull/12 is merged) or afterwards `master`.
2. Run `npm i` and `npm start`
3. Connect the desktop wallet to testnet


src/contexts/walletconnect.tsx
```diff
@@ -204,24 +264,18 @@ export const WalletConnectContextProvider: FC = ({ children }) => {
...
+              response: { id: event.id, jsonrpc: '2.0', result }
             })
             break
           }
           case 'alph_requestExplorerApi': {
             const p = request.params as ApiRequestArguments
-            const result = await client.explorer.request(p)
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            const call = (client.explorer as any)[`${p.path}`][`${p.method}`] as (...arg0: any[]) => Promise<any>
+            const result = await call(...p.params)
```
nop33, line 275: I am not sure why this change was needed, maybe @Lbqds has more info as to why `client.explorer.request(p)` was not sufficient.

src/modals/SendModals/SendModal.tsx
```diff
@@ -99,48 +103,66 @@ function SendModal<PT extends { fromAddress: Address }, T extends PT>({
...
+      setIsLoading(false)
+    },
+    [buildTransaction, dispatch, isConsolidateUTXOsModalVisible, t, txContext]
+  )
+
+  useEffect(() => {
+    if (initialStep === 'info-check' && transactionData) {
+      buildTransactionExtended(transactionData)
+    }
+  }, [buildTransactionExtended, initialStep, transactionData])
```
nop33, line 153: This was needed so that we can skip the build step and go directly to the info-check step when receiving a WC request with predefined tx data (ex: when swapping).

src/modals/SendModals/SendModal.tsx
```diff
@@ -99,48 +103,66 @@ function SendModal<PT extends { fromAddress: Address }, T extends PT>({
...
   const onCloseExtended = useCallback(() => {
     setDappTxData(undefined)
     onClose()
   }, [onClose, setDappTxData])
 
+  const onUserCancel = () => {
+    setDappTxData(undefined)
+    onClose()
+    onError('the user cancelled the operation', requestEvent)
+  }
```
nop33, line 164: I need to review this again.

mvaivre:
That's awesome, really! Thanks for this. Of course there's some room for improvement, but the UX is already much better compared to many big projects around here IMO 🚀 

Functional testing, what I discovered so far:

- In the connection modal, we should allow any discovered address to be used (even if it contains not funds). The WC connection could be used for signing only, no funds required.
- After successfully connecting, I wouldn't show the "disconnect" modal, as it can be disturbing to see the big red disconnect button, looks like an error. Instead, I'd show a success snackbar message, and highlight the wallet connect button in the toolbar with a little animation (I will take care of the latter in my cleanup PR!).
- Disconnection from the wallet wasn't properly acknowledged in the NFT project, but that's is probably not an issue with this PR.

**I'll continue testing later.**

src/contexts/walletconnect.tsx
Lbqds: The dApp connected through WalletConnect converts the request parameters to the `ApiRequestArguments` type in `@alephium/web3` (code is available [here](https://github.com/alephium/alephium-web3/blob/18b7c2b0a457144367bd12cac94ed46b4ae24419/packages/web3/src/api/types.ts#L251)) when sending a request. However, the `ExplorerClient` used by desktop-wallet comes from `@alephium/sdk`, which cannot directly handle requests of type `ApiRequestArguments`. Therefore, we need to make this change, and the code for `@alephium/web3` is available [here](https://github.com/alephium/alephium-web3/blob/18b7c2b0a457144367bd12cac94ed46b4ae24419/packages/web3/src/api/types.ts#L262).

nop33:
> In the connection modal, we should allow any discovered address to be used (even if it contains not funds). The WC connection could be used for signing only, no funds required.

I think only the addresses in the group that the dApp requests are being displayed, not the ones with funds. Need to check.

> After successfully connecting, I wouldn't show the "disconnect" modal, as it can be disturbing to see the big red disconnect button, looks like an error. Instead, I'd show a success snackbar message, and highlight the wallet connect button in the toolbar with a little animation (I will take care of the latter in my cleanup PR!).

👍 

> Disconnection from the wallet wasn't properly acknowledged in the NFT project, but that's is probably not an issue with this PR.

Yes, I believe its the NFT project that needs to add a listener on the `session_delete` event.

nop33:
> After successfully connecting, I wouldn't show the "disconnect" modal, as it can be disturbing to see the big red disconnect button, looks like an error. Instead, I'd show a success snackbar message, and highlight the wallet connect button in the toolbar with a little animation (I will take care of the latter in my cleanup PR!).


Hm, how should the user be able to inspect the current connection then and disconnect if they want to?

mvaivre:
> Hm, how should the user be able to inspect the current connection then and disconnect if they want to?

By clicking on the WC icon in the toolbar :) This would open the current modal.

nop33:
Ah, so the only change you would make is to change the color of the disconnect button?

PR #613, opened 2023-04-17
[v2.0] Price history chart

nop33:
https://user-images.githubusercontent.com/1579899/232534257-b60e3867-76c8-4307-ab92-65252ff4fdbc.mp4


mvaivre:
> Greatly simplify select option navigation

Problem is: navigating with tab doesn't respect accessibility conventions.. Try and navigate in a native select, or [here](https://react-select.com/home). In my previous working implementation, I reproduced this native behaviour.
However, I think that this compromise is fine at the moment. Let's see if our users in need of such features will find it sufficient. 
Also, the current solution doesn't seem to work with multi selects?

PR #636, opened 2023-04-25
Tuesday fixes

nop33:


src/components/TabBar.tsx
```diff
@@ -101,6 +102,7 @@ export const Tab = styled.div<{ isActive: boolean }>`
         `
       : css`
           color: ${theme.font.tertiary};
+          background-color: ${({ theme }) => colord(theme.bg.background2).alpha(0.7).toHex()};
```
mvaivre, line 105: What's that for?

nop33: So the left and right sides of the tabs were not the color of the page background because of where this background prop was. Moving it here maintains the effect that we want in the inactive element without changing the bg color of the whole row.

src/pages/UnlockedWallet/AddressesPage/index.tsx
```diff
@@ -98,5 +99,6 @@ const TabPanel = styled(UnlockedWalletPanel)``
 
 const TabBarPanel = styled(TabPanel)`
   z-index: 1;
+  background-color: ${({ theme }) => colord(theme.bg.background2).alpha(0.7).toHex()};
```
mvaivre: This + the backdrop-filter need to be removed as well

src/pages/UnlockedWallet/OverviewPage/index.tsx
```diff
@@ -71,11 +70,7 @@ const OverviewPage = ({ className }: OverviewPageProps) => {
   )
 }
 
-export default styled(OverviewPage)`
-  background-color: ${({ theme }) => theme.bg.background1};
```
mvaivre, line 75: The overview page was supposed to have the `background1` bg color, but it's now background2. Let's keep your fix for the margin / padding, but revert the background color?

mvaivre:
Shortcut buttons now have a an extra border on hover which isn't supposed to be the case:

image

src/pages/UnlockedWallet/OverviewPage/index.tsx
nop33: ah, damn, I had removed this before our conversation, I'll bring it back.

nop33: Fixed

nop33:
I guess we can merge this since you are working on a PR on top of it.

PR #640, opened 2023-04-26
French i18n v2.0

mvaivre:


nop33:
ok! Why the alphabetic order, if I may ask? How do we track now differences between the different languages? Before we could look always look at the bottom of the file.

nop33 (commented):
ok! Why the alphabetic order, if I may ask? How do we track now differences between the different languages? Before we could look always look at the bottom of the file.

nop33:
Ah, just saw this! I created a tiny script that does a very simple thing:
1. Loads the english translation file and parses it with `JSON.parse`
2. For each other translation file, it checks that every key exists in the english one. If it doesn't, it removes it.
3. For each key of the english file, it checks if the key is in the other language file. If it isn't, it adds it.
4. It writes back the file using `JSON.stringify`, and applies prettier formating.

All we have to do is run a command

```
npm run scripts/syncTranslationFiles
```

If the tool you mentioned can take care of it just as well, I am happy to discard it!

#643

PR #643, opened 2023-04-26
[v2.0] Create script to keep translation files in sync

nop33:
@mvaivre Check this out! It was so simple after all! I purposefully did not commit the results of running the command so that you can test the results yourself.

```
npm run scripts/syncTranslationFiles
```


nop33:
I added the missing keys so that we can go for the rc

nop33:
Also awaiting review/approval and merge 😇

mvaivre (approved):
Super nice

Issue #653, opened 2023-04-26
Not possible to click around select modals to close them


nop33:
![image](https://user-images.githubusercontent.com/1579899/234646028-0ebbed3a-73db-44b1-a2df-c8852a019bab.png)

nop33:
merged

PR #655, opened 2023-04-26
Fix contacts, toggle, add sidebar in new wallet flow

### LINKED ISSUE #652: Fix contacts flex wrapping
image

mvaivre:
Made this one late in the day, so a fresh reviewer brain & tester hands would be appreciated <3

- Fixes #652, #653
- Fixes toggle behaviour when controlled (bug was unnoticed until now). Try in settings > network > set custom.
- Improves contact cards
- Add sidebar in new wallet flow
- Try with reducing passphrase toggle opacity

Also, I just saw that @nop33 you added the sidebar as well in your branch 🙃 Please compare and choose one of the implementation (in term of looks, the goal is to have the same sidebar as in this current branch). Thanks!

### DIFF
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -20,10 +20,11 @@ import { motion } from 'framer-motion'
 import styled, { css } from 'styled-components'
 
 const Card = styled(motion.div)<{ isPlaceholder?: boolean }>`
-  width: 222px;
+  width: 230px;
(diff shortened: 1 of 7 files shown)


nop33 (approved):
image

Just a minor thing, I'll make a commit.

I'll also use my code for adding the sidebar because it's DRYer and it handles the flickering of the theme switcher (when in light theme, there is a flickering because the initial value is "dark").

PR #662, opened 2023-04-27
[v2.0] Fix WalletConnect and passphrase address generation

nop33:
When the dApp requests a signer address in a specific group and the wallet doesn't have one, here's how we handle it:

https://user-images.githubusercontent.com/1579899/234892261-e5df258e-cada-4b2b-beaf-a947f6431f86.mp4


src/modals/WalletConnectModal.tsx
```diff
@@ -161,9 +175,15 @@ const WalletConnectModal = ({ onClose }: WalletConnectModalProps) => {
     onClose()
   }
 
+  const handleGoToAddressesPageClick = () => {
+    rejectConnectionAndCloseModal()
```
polarker, line 179: We might need to await `walletConnectClient.reject` inside `rejectConnectionAndCloseModal`.

nop33: Good call, but I am curious to know why! I personally think the user shouldn't have to wait when they:
- Close the modal clicking the "X" button
- Close the modal clicking outside of it
- Clicking the "generate a new address" link to go to the addresses page to generate an address in the group that the dApp wants.

All those are the times where we call `rejectConnectionAndCloseModal`.

However, I totally think it makes sense to wait for it when clicking "Reject" (which is already the case), because then we move the app to the background and bring the dApp back to the foreground. I think we should do that only when rejection is confirmed.

Nevertheless, I will add the `await` in the `walletConnectClient.reject` call inside `rejectConnectionAndCloseModal` but I think we shouldn't await for `rejectConnectionAndCloseModal` in `handleGoToAddressesPageClick` or when the user manually closes the modal. Let me know your thoughts!

polarker: You had a good point, I agree that there is no point to wait the call for the modal, we probably need to wait to update the state of WC. I don't have time to check the details, so please just take this as a remark. Please feel free to close this.

mvaivre:
And what about adding a "Generate" button directly in the InfoBox? Any technical cons?
This would remove the need for many user inputs!

nop33:
We can totally do that.

nop33:
@mvaivre Done!


https://user-images.githubusercontent.com/1579899/235445775-7fb972ee-314b-4db7-9e4a-79ab3e257a38.mp4

mvaivre:
Just fantastic! So easy now, well done.

mvaivre (approved):

PR #664, opened 2023-04-28
Friday fixes

mvaivre:
Fixes white screen on wallet switch.
@nop33: we really need to adopt another paradigm. What I did here is a quick fix, let's discuss this soon.


nop33 (approved):
I've reviewed the changes and instead of explaining my proposal in the review here, I implemented it in #669 and started a discussion thread there. I am happy to merge this and continue there.

PR #669, opened 2023-05-01
[v2.0] Monday fixes

nop33:


src/components/AddressBadge.tsx
```diff
@@ -37,53 +36,69 @@ type AddressBadgeProps = ComponentPropsWithoutRef<typeof Badge> & {
...
+  withBorders
+}: AddressBadgeProps) => {
   const { t } = useTranslation()
   const address = useAppSelector((s) => selectAddressByHash(s, addressHash))
   const selectContactByAddress = useMemo(makeSelectContactByAddress, [])
   const contact = useAppSelector((s) => selectContactByAddress(s, addressHash))
 
-  return contact ? (
-    <AddressBadgeContainer {...props}>
-      <Label {...props}>
```
nop33, line 52: ok, so, let's discuss our new approach on exporting styled components here.

We both agree that the following starts becoming annoying, for several reasons (not very easy navigation through our IDE to the component, errors with types, dev server requires restart some times)

```ts
export default styled(ComponentName)``
```

We'd rather have a simple default export like so:

```ts
export default ComponentName
```

but we also want to be able to style the root component.

# Naming conventions

I propose the convention `ComponentNameStyled` for the following reasons:
1. I find it more self explanatory than the convention `ComponentNameContainer`. Reading `ComponentNameStyled` I immediately think that this is the component `ComponentName` with some added "styles" to it. Which is essentially what it is. 
2. I have been using the convention `ComponentNameStyled` throughout the app when I want to add additional styles to an existing component. I noticed that you prefer `StyledComponentName`. Even though they are essentially the same, I prefer having the actual name of the component as the prefix. I feel that I can recognise the component faster this way, and think less.

# Technical implementation

Our previous method of exporting the `styled()` component allowed us to have direct access to ALL the typed props of the component, without additional code:

```ts
// ComponentName.tsx
interface ComponentNameProps {
  prop1: string
  className?: string
}

const ComponentName = ({ prop1, className }: ComponentNameProps) => {
  return <div className={className}>{prop1}</div>
}

export default styled(ComponentName)`
  font-size: ${({ prop1 }) => (prop1 === 'x' ? 10 : 20)}px;
`

// Using the component in another component X.tsx
<ComponentName prop1="x" />
```

With the new proposal, however, our `ComponentNameStyled` does not immediately have access to the `ComponentNameProps`. These are the options I can think of:

## Option 1
1. Do not deconstruct the props object in the component definition, but in its body, so that we can deconstruct it in the instance of `ComponentNameStyled`.
2. Type `ComponentNameStyled` with the same type as the `ComponentName`

```ts
interface ComponentNameProps {
  prop1: string
  className?: string
}

const ComponentName = (props: ComponentNameProps) => {
  const { prop1 } = props

  return <ComponentNameStyled {...props}>{prop1}</ComponentNameStyled>
}

export default ComponentName

const ComponentNameStyled = styled.div<ComponentNameProps>`
  font-size: ${({ prop1 }) => (prop1 === 'x' ? 10 : 20)}px;
`
```

### Problems with option 1
I am very concerned regarding the use of `{...props}`. It has led to some weird bugs in the past when some HTML tags would receive our React Component props that happened to also be [valid HTML attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes). Take for example the valid HTML attribute `headers` and imagine we have a component like the following. Notice what gets rendered in the DOM (see below code snippet)

```ts
interface ComponentNameProps {
  prop1: string
  headers: { name: string }[]
  className?: string
}

const ComponentName = (props: ComponentNameProps) => {
  const { headers } = props

  return <ComponentNameStyled {...props}>{headers.map(({ name }) => name)}</ComponentNameStyled>
}

export default ComponentName

const ComponentNameStyled = styled.div<ComponentNameProps>`
  font-size: ${({ prop1 }) => (prop1 === 'x' ? 10 : 20)}px;
`
```

```html
<div headers="[object Object],[object Object]" class="sc-eIoBCF irHphN">onetwo</div>
```

This is just a small example that doesn't impact the behavior of our app, just the rendered DOM, but I remember in the past spending countless hours trying to debug an issue only to realize that it was the use of `{...props}` that rendered unintented things to the DOM. Since then, I try to use `{...props}` very carefully and only in the cases where I know exactly that what I pass to the underlying component will be valid props, like the following example:

```ts
import Button, { ButtonProps } from '@/components/Button'

const DeleteButton = (props: ButtonProps) => <Button Icon={X} role="secondary" {...props} />
```

### Option 2
Another thing that bothers me option 1 is not being able to deconstruct the props at the component definition and having to do it in the body. Alternatively, we could only pass to the `ComponentNameStyled` the properties needed in the styles by picking them out of the `ComponentNameProps`:

```ts
interface ComponentNameProps {
  prop1: string
  prop2: string
  className?: string
}

const ComponentName = ({ prop1, prop2, className }: ComponentNameProps) => (
  <ComponentNameStyled prop1={prop1} className={className}>
    {prop1}
    {prop2}
  </ComponentNameStyled>
)

export default ComponentName

const ComponentNameStyled = styled.div<Pick<ComponentNameProps, 'prop1'>>`
  font-size: ${({ prop1 }) => (prop1 === 'x' ? 10 : 20)}px;
`
```

or we could avoid deconstructing altogether with just a simpler props var naming like so:

### Option 3

```ts
interface ComponentNameProps {
  prop1: string
  className?: string
}

const ComponentName = (p: ComponentNameProps) => {
  return <ComponentNameStyled {...p}>{p.prop1}</ComponentNameStyled>
}

export default ComponentName

const ComponentNameStyled = styled.div<ComponentNameProps>`
  font-size: ${({ prop1 }) => (prop1 === 'x' ? 10 : 20)}px;
`
```

---

So far, these are the options I have thought about and what problems they might have. In practice we might find more limitations, but we'll have to try and see. I personally don't like neither option 1 nor option 2 when comparing to our current convension. From the code style point of you, I think it's the cleanest. But since we are hitting some limitations, I'd be in favor of option 2 or 3.

mvaivre: Thanks you very much for this summary.

- ✅ `ComponentNameStyled`.  I agree, it's faster to spot the component name this way.
- ✅ Option 2. Fully aligned with what you said. It's a bit of a pain to `Pick` all the needed props, but yeah, it's still the best type safe option we have here IMO.

Let's adopt this "new" approach from now on 🤝

src/modals/CenteredModal.tsx
```diff
@@ -191,7 +191,7 @@ const BackButton = styled(Button)`
 export const ModalContent = styled.div`
   display: flex;
   flex-direction: column;
-  padding: var(--spacing-4);
+  padding: var(--spacing-4) var(--spacing-4) 0 var(--spacing-4);
```
mvaivre, line 194: Why this change? Now the Action buttons are really close to the modal bottom and it's less balanced IMO.

mvaivre, line 194: Why this change? Now the Action buttons are really close to the modal bottom and it's less balanced IMO. What I tried to achieve before:
- Padding around action button when they are not floating above content
- Reduced padding when they are, to give more space to content.

mvaivre: Now that I'm talking padding : we could increase lateral padding to be closer to the mockups and give more space to the scrollbar. Here's what it looks like with `var(--spacing-4) var(--spacing-6)` :

image


What do you think?

mvaivre: I've just seen the comment [here](https://github.com/alephium/desktop-wallet/commit/978aa5b71ccf3666beb7174c88dd4358b1808317). Yeah unfortunately I feel like it lacks space compared to the rest of the UI.

nop33: I mentioned the reason in the commit and I tagged you!

src/components/AddressBadge.tsx
nop33: Nice, glad you agree!

src/modals/CenteredModal.tsx
nop33: I understand. If you don't mind tackling this while I work on the WC issue, I'd appreciate it a lot! I'll focus my attention there now.

nop33: LGTM 👍

mvaivre (approved):
LGTM 👾

Issue #675, opened 2023-05-03
Enable copy of address from address badge


nop33:
@JKtranslator proposed today that he was trying to copy his main address from the address badge in the app header and he couldn't. He specifically said (I'm paraphrasing):

> Since you asked me to send you my address so that you send me some testnet tokens, I had to look for the right place to copy it.

nop33:
@VDAODAO also reported that he was trying to copy the address hovering the address hash but he couldn't. I think we should have it enabled everywhere imo. The users can easily get used to the idea and then it's problematic when it's not possible to copy it in some cases:

> WISH - everywhere else in the app, i can copy the adress from the little two boxes on the right of the adress. Here I spent 5' to find that the copy button was below the QR code. I understand this probably comes from ledger, but in terms of coherence it’s just a bit counterintuitive 
>  ![image](https://user-images.githubusercontent.com/1579899/236022438-47da0cec-2729-4bea-90dc-24c9a240f9ef.png)

mvaivre:
Mhh ok. Let's keep the disableCopy prop though. In some places (in selects for instance) I found this to be distractive.

PR #683, opened 2023-05-03
Fix internal testing bugs and recommendations

nop33:
Both @JKtranslator and @VDAODAO reported that they felt confused that they couldn't copy the address in certain places, as they got used to in every other place (see #675). I address this problem with this PR by re-enabling copy.

This PR also solves another small issue. When there is only 1 address in the wallet, clicking the "Default address" badge on the header would constantly set the same address as default, triggering another save in the localstorage. I address this issue by disabling the click event on the select fields when there is only 1 option that is already selected.


src/modals/WalletConnectModal.tsx
```diff
@@ -181,7 +181,7 @@ const WalletConnectModal = ({ onClose }: WalletConnectModalProps) => {
   }
 
   const generateAddressInGroup = () => {
-    const address = generateAddress()
+    const address = generateAddress({ group })
```
nop33, line 184: Somehow we both missed it, just because in the demo it HAPPENED that the next generated address was in group `0`. oopsi

src/utils/transactions.ts
```diff
@@ -18,15 +18,14 @@ along with the library. If not, see <http://www.gnu.org/licenses/>.
...
-  DUST_AMOUNT,
   getDirection,
   isConsolidationTx,
-  MIN_UTXO_SET_AMOUNT,
   TransactionDirection,
   TransactionInfoType
 } from '@alephium/sdk'
 import { AssetOutput, Output } from '@alephium/sdk/api/explorer'
 import { ALPH } from '@alephium/token-list'
+import { DUST_AMOUNT, MIN_UTXO_SET_AMOUNT } from '@alephium/web3'
```
nop33, line 28: If we would have imported the DUST_AMOUNT from the web3 package we wouldn't have had this bug. But maybe it wasn't there when we first started working on this feature. In any case, I imported all constants from web3 now.

src/modals/SendModals/AssetAmountsInput.tsx
```diff
@@ -302,7 +297,7 @@ const AssetSelect = styled(SelectContainer)`
   margin: 0;
 `
 
-const SelectInput = styled.div<InputProps>`
+const SelectInput = styled.button<InputProps>`
```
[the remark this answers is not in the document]
mvaivre: I managed to reproduce it in my branch. I'll rebase on your branch and test again!

mvaivre (approved):
Code looks good 👏 I'll rebase on v2.0 once this is merged for a final round of testing while I'm working on the last send modal change.

PR #688, opened 2023-05-04
Update screenshot

nop33:
image


polarker:
The credit section is also outdated. It was the case when me and Alois were working on it 😅

nop33:
Let's just merge it now and be done with it. New rc and then release on monday.

Issue #682, opened 2023-05-03
Clarify why there's an extra 0.01 ALPH per token added in the tx


nop33:
It's confusing that when sending tokens, there's an additional `0.01` ALPH per token that is added to the transaction. The UI should clarify why this is happening. 

@polarker could you please remind me of the reason so that I can think of how to best present this info to the user?  Was it because every UTXO should have a small amount of ALPH for some technical reason?

thx!

polarker:
The 0.01 ALPH is the minimal requirement per UTXO to avoid utxo spamming. There are usually two outputs in a simple token transaction, one for the recipient and another one for change output. That's why you see 0.02 ALPH in the output.

polarker:
Here is the definition in the web3 SDK: https://github.com/alephium/alephium-web3/blob/master/packages/web3/src/constants.ts#L24

mvaivre:
Yep, agreed, we should add an info icon redirecting to a paragraph in the docs.
Also, I'd make sure that the token(s) to send are displayed above the amount in ALPH in cases like this, WDYT?

nop33:
Good point.

That means we need a paragraph in the docs :)

PR #691, opened 2023-05-04
Last before v2.0

nop33:
~~Waiting for GH to fix its stuff so that I can push my last commit. Thus: WIP~~

### DIFF
--- a/src/utils/transactions.ts
+++ b/src/utils/transactions.ts
@@ -58,11 +58,12 @@ export const getTransactionAssetAmounts = (assetAmounts: AssetAmount[]) => {
     .filter((asset): asset is Required<AssetAmount> => asset.id !== ALPH.id && asset.amount !== undefined)
     .map((asset) => ({ id: asset.id, amount: asset.amount.toString() }))
 
-  const dustAmount = DUST_AMOUNT * BigInt(tokens.length)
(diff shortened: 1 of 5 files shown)


src/utils/transactions.ts
```diff
@@ -58,11 +58,12 @@ export const getTransactionAssetAmounts = (assetAmounts: AssetAmount[]) => {
...
 
-  const dustAmount = DUST_AMOUNT * BigInt(tokens.length)
-  const attoAlphAmount = (alphAmount + dustAmount).toString()
+  const minAlphAmountRequirement = DUST_AMOUNT * BigInt(tokens.length)
+  const minDiff = minAlphAmountRequirement - alphAmount
+  const totalAlphAmount = minDiff > 0 ? alphAmount + minDiff : alphAmount
 
   return {
-    attoAlphAmount,
+    attoAlphAmount: totalAlphAmount.toString(),
```
nop33, line 66: @mvaivre I had to add this last commit. After the discussion here https://github.com/alephium/desktop-wallet/issues/682 I realized that we unnecessarily were forcing the user to send an additional 0.001 ALPH per token, no matter if they decide to send ALPH or not. I realized that the min requirement should be added **only if** the ALPH amount is not enough.

Sorry for breaking the code-freeze, but this was important to fix imo.

mvaivre (approved):

PR #717, opened 2023-05-19
[WIP] Faucet

mvaivre:


nop33:
Ready for review?

src/storage/assets/assetsActions.ts
```diff
@@ -41,3 +44,22 @@ export const syncNetworkTokensInfo = createAsyncThunk('assets/syncNetworkTokensI
...
+
+export const receiveTestnetTokens = createAsyncThunk<undefined, AddressHash, { rejectValue: SnackbarMessage }>(
+  'assets/receiveTestnetTokens',
+  async (destinationAddress: AddressHash, { rejectWithValue }) => {
+    const response = await fetch('https://faucet.testnet.alephium.org/send', {
+      method: 'POST',
+      body: destinationAddress
+    })
+
+    console.log(response)
```
nop33, line 56: Do we wanna keep this for debugging?

nop33 (commented):
This is really cool feature!!

When I clicked "Receive testnet tokens", it took a couple of seconds without any visual feedback until a success snackbar popped up. I think we can improve this with a loading state, or at least a snackbar that says "tokens requested" or sth.

Secondly, after the success message, I don't see any transactions in my transaction history. I tried refreshing and there's still nothing. Then after refreshing 2 minutes later, I could see the tokens. I did not have enough time to read the success message, maybe we can leave it open a bit longer. Moreover, since we are expecting an incoming transaction, we should refresh the data automatically, same way we refresh them with an interval when sending a tx. WDYT?

Lastly, we could improve the error handling (in the future, no need for now) when the user reaches the quota (not sure how many minutes it is). Currently it shows "Encountered error, please try again in a few minutes", but it could say "you reached the limit, try again in X minutes" or sth like that.

mvaivre:
Yes, you saw #718 😆 I'm not happy with the current flow either, it's hard to really see that something happened.

A button spinner + fetching incoming transactions would help a lot.

mvaivre:
So, it took me some time to get in the redux stuff, and this work will eventually be replaced by fetching the mempool data directly, but it's now showing incoming faucet TXs using your system.

src/storage/addresses/addressesSlice.ts
```diff
@@ -210,6 +203,15 @@ const addressesSlice = createSlice({
           }
         })
       })
+
+    builder.addMatcher(isAnyOf(transactionSent, receiveTestnetTokens.fulfilled), (state, action) => {
+      const pendingTransaction = action.payload
+      const fromAddress = state.entities[pendingTransaction.fromAddress] as Address
+      const toAddress = state.entities[pendingTransaction.toAddress] as Address
+
+      ;(fromAddress || toAddress).transactions.push(pendingTransaction.hash)
```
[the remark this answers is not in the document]
nop33: I think this should ensure that `toAddresss.transactions.push(pendingTransaction.hash)` gets called only once and also avoids the weird syntax.

```ts
if (fromAddress) fromAddress.transactions.push(pendingTransaction.hash)
if (toAddress && toAddress !== fromAddress) toAddress.transactions.push(pendingTransaction.hash)
```

src/storage/settings/settingsActions.ts
```diff
@@ -47,3 +51,37 @@ export const walletLockTimeChanged = createAction<Settings['general']['walletLoc
...
+
+    return fulfillWithValue({
+      hash: hash,
+      fromAddress: 'Faucet',
+      toAddress: destinationAddress,
+      amount: '12',
+      timestamp: new Date().getTime(),
+      status: 'pending',
+      type: 'transfer'
+    })
```
nop33, line 85: GJ! 👏 

nit: How about assigning the return value to a variable called `pendingTransaction` first, just for readability's sake?

src/storage/settings/settingsSlice.ts
```diff
@@ -46,15 +47,22 @@ import SettingsStorage from '@/storage/settings/settingsPersistentStorage'
...
+}
 
 const settingsSlice = createSlice({
   name: 'settings',
   initialState,
   reducers: {},
   extraReducers(builder) {
     builder
-      .addCase(localStorageDataMigrated, () => SettingsStorage.load('general') as GeneralSettings)
+      .addCase(localStorageDataMigrated, () => SettingsStorage.load('general') as SettingsState)
```
nop33, line 65: Hm, why this change? The type of the data loaded is `GeneralSettings`. `faucetCallPending` is not stored there.

src/storage/settings/settingsSlice.ts
```diff
@@ -46,15 +47,22 @@ import SettingsStorage from '@/storage/settings/settingsPersistentStorage'
 import { RootState } from '@/storage/store'
 import { GeneralSettings } from '@/types/settings'
 
-const initialState: GeneralSettings = SettingsStorage.load('general') as GeneralSettings
+interface SettingsState extends GeneralSettings {
+  faucetCallPending: boolean
```
nop33, line 51: Tbh, not a big fan of using this slice for storing this. I think it would fit better the `global/globalSlice.tsx`. WDYT?

nop33 (commented):
I noticed that the displayed ALPH amount of the faucet pending incoming tx is `0.00 ALPH`. Maybe we could improve this to display nothing instead since we don't know the amount (just like we do on sweep transactions where we don't know the exact amount)?

src/storage/settings/settingsSlice.ts
mvaivre: Because otherwise the compiler will complain, as `addCase` gets the expected type from `initialState`.
But this will be solved if I move `faucetCallPending ` to the `globalSlice` as you proposed below.

src/storage/settings/settingsSlice.ts
mvaivre: Good idea!

nop33 (approved):
Beautiful! ✨

PR #720, opened 2023-05-23
Custom proxy support

mvaivre:


mvaivre:
@nop33 Not yet asking for a review, but some functional testing with a SOCKS5 proxy. @polto Would you like to try this as well? :)

nop33:
I'll have a look into this tmr

nop33:
@mvaivre any instructions on how to perform the functional tests?

src/modals/SettingsModal/NetworkSettingsSection.tsx
```diff
@@ -169,23 +189,46 @@ const NetworkSettingsSection = () => {
...
+            value={tempNetworkSettings.explorerUrl}
+            onChange={(e) => editNetworkSettings({ explorerUrl: e.target.value })}
+            noMargin
+          />
+          <h2 tabIndex={0} role="label">
+            {t('Custom proxy (SOCKS5)')}
+          </h2>
+          <Input
+            id="proxy-address"
+            label={t`Proxy address`}
```
nop33, line 221:
```suggestion
            label={t('Proxy address')}
```

also for below.

src/types/settings.d.ts
```diff
@@ -43,6 +49,8 @@ export interface Settings {
 
 export type Language = 'en-US' | 'fr-FR' | 'de-DE' | 'vi-VN' | 'pt-PT' | 'ru-RU' | 'bg-BG' | 'es-ES' | 'id-ID' | 'tr-TR'
 
+export type Language = 'en-US' | 'fr-FR' | 'de-DE' | 'vi-VN' | 'pt-PT' | 'ru-RU' | 'bg-BG'
```
nop33, line 52: This seems like unintended.

public/preload.js
```diff
@@ -64,6 +64,7 @@ contextBridge.exposeInMainWorld('electron', {
   app: {
     hide: () => ipcRenderer.invoke('app:hide'),
     show: () => ipcRenderer.invoke('app:show'),
-    getSystemLanguage: async () => ipcRenderer.invoke('app:getSystemLanguage')
+    getSystemLanguage: async () => ipcRenderer.invoke('app:getSystemLanguage'),
+    setProxySettings: (proxySettings) => ipcRenderer.invoke('app:setProxySettings', proxySettings)
```
nop33, line 68: since `setProxySettings` returns a Promise, shouldn't this function be async?

nop33 (commented):
I can't get it to work, following the steps I wrote above... I keep getting `ERR_TIMED_OUT` for 
```
GET https://wallet-v20.testnet.alephium.org/infos/chain-params
```

and being stuck in loading state.

src/types/settings.d.ts
mvaivre: Indeed! Weird, and good catch!

public/preload.js
mvaivre: Mhh, I'm not sure actually. Shouldn't we use the `async` prefix only when using `await` in the function body?
And we don't want to await here I think.

BTW, the "hide" and "show" functions also return a promise, like all the other `invoke` calls.

For consistency sake, we should probably remove all the `async` prefixes in the file.

nop33: I think you're right 👍

public/electron.js
```diff
@@ -290,6 +292,17 @@ app.on('ready', async function () {
     if (preferedLanguages.length > 0) return preferedLanguages[0]
   })
 
+  ipcMain.handle('app:setProxySettings', async (_, proxySettings) => {
+    const { address, port } = proxySettings
+    let proxyRules = !address && !port ? undefined : `socks5://${address}:${port}`
```
nop33, line 297: why `let` and not `const`?

nop33 (approved):
Apart from that little comment, I don't have any other comments regarding the code or the functionality!

After resolving the conflicts a functional test might be needed again, but I guess nothing will change in the code-wise

nop33:
I think the npm audit is fixed in a following PR, don't bother in this one.

mvaivre:
Note for myself: one more thing todo is add a little "experimental" badge next to the feature.

mvaivre:
@nop33 asking for a quick re-review 🙏

nop33 (approved):

nop33:
Functional tests are good ✅  Will merge once tests pass.

PR #721, opened 2023-05-23
Simplify and improve chart animations

mvaivre:


nop33 (commented):
Hm, these changes introduce a lag when opening the Address Details modal... I believe this is due to the removal of 

```ts
onAnimationComplete={(variant) => setShouldMountChart(variant === 'shown')}
```

I think I liked it more before, the UI felt more fluid and responsive.

mvaivre:
> I think I liked it more before, the UI felt more fluid and responsive.

I have the opposite experience 🙃 Do you have your dev tools open?

nop33:
oh, that's interesting! I don't get the lag with devtools closed! nice!

mvaivre:
For me it's actually more fluid with this approach compared to before. Do you feel the same (with dev tools closed)?

nop33 (approved):

PR #722, opened 2023-05-24
Update SDKs

nop33:


mvaivre (approved):

mvaivre:
Should we take this opportunity to add a `verified` property in the metadata object? When upserting the tokens info, this field should reflect wether the metadata are available in the `token-list` package or trough the node only.

Issue #724, opened 2023-05-25
Fetch unknown token metadata from node


nop33:
We're referring to:
- token name
- token symbol
- decimals

The main concern is that the metadata can be whatever the creator of the contract decides, which means there can be scam tokens trying to impersonate other tokens. 

It's probably safe to fetch the decimals so that we can properly display the amounts.

As for the name and symbol, we can decide to either not fetch it or display it with a big fat warning everywhere it's being displayed. 

I am in favor of going forward with just the decimals for now.

nop33:
Merged. Closing.

PR #725, opened 2023-05-30
Fetch unknown tokens metadata from node

nop33:
Related to #724

In this PR we call the `fetchStdTokenMetaData` node endpoint for every unknown token, as long as we are connected to the mainnet or testnet. The metadata that we fetch atm is:
- name
- symbol
- decimals

We still need to add a warning though.


mvaivre:
Should we take this opportunity to add a verified property in the metadata object? When upserting the tokens info, this field should reflect wether the metadata are available in the token-list package or trough the node only.

mvaivre:
It would also be great if we could tweak the default asset order a little bit IMO. 
I'd suggest: known tokens first (alphabetical order), then unknown tokens below. WDYT?

nop33:
Absolutely. I would like to extend the order to also order by worth, but that might be a bit useless for now. How do we plan to fetch prices for bridge tokens? Then, if we cannot order by worth, maybe ordering alphabetically makes more sense.

mvaivre:
It may indeed be a bit too soon to start trying to order by worth. We will one day provide an option to select the ordering strategy. For now I think that alphabetical ordering makes the most sense.

mvaivre:
Final point: let's use this PR to change the unknown token's icon to [this one](https://lucide.dev/icons/help-circle), and add a warning badge (unverified token) when metadata are only available on the node.

mvaivre:
Also, I'm now thinking: should fetchStdTokenMetaData be called on the node directly like it is done now, or should be go though the explorer backend as we do for all the other read-only data? 

cf. @tdroxler

nop33:
We do use the node API however for tx related things (build, send), fetching the chain id to identify if it's mainnet or testnet and for contract related things. Basically we only use POST endpoints (tx, contracts) and one GET endpoint `infos.getInfosChainParams` (on top of `fetchStdTokenMetaData` that this PR adds). Just wanted to add this!

mvaivre:
Thanks @nop33. 

On the explorer I don't use the node client at all for now. If we want to display "unverified" metadata there, this would be the first time I'd be using it. This is why I started asking myself this question :) Interested to know @tdroxler thoughts on this!

tdroxler:
Currently we can't call the explorer-backend because the `stdTokenMetaData` aren't store there. We will need to think if we want to. 
I need to check how those data are fetch from the node, but I'm afraid it comes from contract state, which is mutable data and therefore could be super tricky to store on the explorer-backend

mvaivre:
Update after discussing with Cheng: let's start with calling the node directly, but at one point we will eventually need to store metadata in the explorer backend in order to build a proper token page. So let's keep that in mind :)

src/storage/addresses/addressesSelectors.ts
```diff
@@ -101,11 +103,22 @@ export const makeSelectAddressesAssets = () =>
...
+      const [unverifiedTokens, unknownTokens] = partition(remainingTokens, (token) => token.verified === false)
+
+      return [
+        alphAsset,
+        ...(verifiedTokens as VerifiedAsset[]).sort((a, b) => a.name.toLowerCase().localeCompare(b.name.toLowerCase())),
+        ...(unverifiedTokens as UnverifiedAsset[]).sort((a, b) =>
+          a.name.toLowerCase().localeCompare(b.name.toLowerCase())
+        ),
+        ...unknownTokens.sort((a, b) => a.id.toLowerCase().localeCompare(b.id.toLowerCase()))
+      ]
```
mvaivre, line 121: Nit: may be replaced by a one liner like (to be tested):

```ts
return sortBy(tokenAssets, [(a) => !a.verified, (a) => !symbol, (a) => a.name.toLowerCase()])
```

src/storage/assets/assetsActions.ts
```diff
@@ -41,3 +43,19 @@ export const syncNetworkTokensInfo = createAsyncThunk('assets/syncNetworkTokensI
 
   return metadata
 })
+
+export const syncUnknownTokensInfo = createAsyncThunk(
+  'assets/syncUnknownTokensInfo',
+  async (unknownTokenIds: Asset['id'][]) => {
+    const tokens = await Promise.all(
```
mvaivre, line 50: One thing to keep in mind here, that caused me troubles on the explorer: `Promise.all` is all or nothing. If any of the promises in the map fails, the whole sync will fail. This can be problematic if this ends up in showing incomplete lists of assets.

src/storage/assets/assetsActions.ts
```diff
@@ -41,3 +43,19 @@ export const syncNetworkTokensInfo = createAsyncThunk('assets/syncNetworkTokensI
...
+      unknownTokenIds.map((id) =>
+        client.node.fetchStdTokenMetaData(id).then((data) => ({
+          id,
+          ...data
+        }))
+      )
+    )
+
+    return tokens.map(({ name, symbol, decimals, id }) => ({ name, symbol, decimals, id }))
```
mvaivre, line 59: What's this map for?

src/storage/addresses/addressesSelectors.ts
nop33: wow, thanks for opening my eyes to `sortBy`! I discovered a rly need oneliner:

```ts
sortBy(assets, [(a) => !a.verified, (a) => a.verified === undefined, 'name', 'id'])
```

1. The first function brings the verified assets (`verified === true`) to the top of the list
2. The second function brings the `verified === false` assets right below the verified assets
3. The `'name'` argument sorts items where the `verified` prop is the same by `name`
4. The `'id'` argument sorts items where the `verified` and `name` props are the same by `id`

Fixed ✨

src/storage/assets/assetsActions.ts
nop33: To remove `totalSupply: Number256`.

nop33: Would it be better if I wrote this?

```ts
return tokens.map(({ totalSupply, ...rest }) => ({ ...rest }))
```

mvaivre: Yes, I like it better, way more explicit! 👌

src/storage/addresses/addressesSelectors.ts
mvaivre: You're welcome :) One thing: I don't know if the string sorting is actually case independent?

src/storage/assets/assetsActions.ts
nop33: Thanks for bringing my attention to this!

After reading some doc pages and articles, I think that for this particular case, this is the best way to do it:

```ts
    const tokens = await Promise.allSettled(
      unknownTokenIds.map((id) =>
        client.node.fetchStdTokenMetaData(id).then((data) => ({
          id,
          ...data
        }))
      )
    )

    return (tokens.filter((result) => result.status === 'fulfilled') as PromiseFulfilledResult<TokenMetaData>[]).map(
      ({ value: { totalSupply, ...rest } }) => rest
    )
```

1. `Promise.all` is not optimal because the data returned from each promise is not interdependent, so we don't want to fail the whole thing if any of them fails. Even though `Promise.all` is the most performant solution, we want more granular control, so that we can "save" the promises that fulfill and discard the ones that reject.
2. `for await...of` would be a good solution if we wanted to handle each rejection separately. For this particular case, I don't think we care to react when one of these promises fails. We consider the token as `unknown` directly.
3. `Promise.allSettled` is the best solution in my opinion. It returns an array with a `status` for each item to inform whether it fulfilled or rejected. If it's fulfilled, it includes the return value. It's faster than `for await...of` and it gives more granular control than `Promise.all`.

src/storage/addresses/addressesSelectors.ts
nop33: Good point, but on second thought, I don't think we care a lot about case.

nop33: Fixed! ✅

nop33:
All comments were addressed, ready for review!

mvaivre:
This PR can be updated with the new utility functions to fetch token metadata. There's a distinction between fetching fungible tokens or NFTs. This will prevent the 500 errors we got : https://github.com/alephium/alephium-web3/pull/193#event-9561951802

src/components/AssetLogo.tsx
nop33, line 1: Instead of showing a ❓ in the NFT transaction rows, I chose to display the actual NFT. It looks like this:

image

nop33 (commented):
An overall comment about managing the state:

Our `addressSlice` stores a list of addresses. Each address stores a list of tokens ids and balances. It looks sth like this:

```json
{
  "addresses": [
    {
      "hash": "1Cr776f9jK8aBWeV91GiGDTFPiv7JZBACGoJiNj2QWRZh",
      "tokens": [
        {
          "id": "x7ZvEa7cEwxFqzLvr5A8GNH8Fkn1AfcuhgHyJPV5cw1q",
          "balance": "100",
          "lockedBalance": "0"
        }
      ]
    }
  ]
}
```

Then we have the `assetsInfoSlice` (which should probably be renamed to `tokensMetadataSlice` for clarity and consistency between node and frontend) that stores tokens metadata. It looks sth like this:

```json
{
  "assetsInfo": [
    {
      "id": "x7ZvEa7cEwxFqzLvr5A8GNH8Fkn1AfcuhgHyJPV5cw1q",
      "name": "Fetatoken",
      "symbol": "FETA",
      "decimals": 18,
      "description": "Fetamin's fan club token",
      "logoURI": "xyz",
      "verified": true
    }
  ]
}
```

Lastly, we have the new `nftsSlice` that stores NFT metadata and it looks like this:

```json
{
  "nfts": [
    {
      "id": "x7ZvEa7cEwxFqzLvr5A8GNH8Fkn1AfcuhgHyJPV5cw1q",
      "name": "Alpaca #001",
      "description": "The first alpaca on Alephium",
      "image": "xyz"
    }
  ]
}
```

---

With this structure, to get all info of the tokens of an address (balance, display info, etc), we:
1. Get all the address token IDs from the `addressesSlice`
2. Cross-reference them with the data in `assetsInfo`
3. Merge the info between the 2 slices

Similarly, to get all info of the NFTs of an address, we:
1. Get all the address token IDs from the `addressesSlice`
2. Cross-reference them with the data in `nftsSlice`
3. Merge the info between the 2 slices

src/storage/assets/assetsActions.ts
```diff
@@ -41,3 +45,43 @@ export const syncNetworkTokensInfo = createAsyncThunk('assets/syncNetworkTokensI
...
+        if (type === 'fungible' || type === undefined) {
+          const tokenMetadata = await client.node.fetchFungibleTokenMetaData(id)
+
+          results.tokens.push({
+            id,
+            ...omit(tokenMetadata, ['totalSupply'])
+          })
+        } else if (type === 'non-fungible') {
+          const { tokenUri, collectionAddress } = await client.node.fetchNFTMetaData(id)
+          const nftData = await exponentialBackoffFetchRetry(tokenUri).then((res) => res.json())
```
nop33, line 70: The type of `nftData` is `any` since we can't know what the `tokenUri` will return.

src/types/assets.ts
```diff
@@ -16,6 +16,22 @@ You should have received a copy of the GNU Lesser General Public License
...
+
+export type FungibleTokenBasicMetadata = Omit<FungibleTokenMetaData, 'totalSupply'> & { id: Asset['id'] }
+
+export type NFT = {
+  id: Asset['id']
+  collectionAddress: NFTMetaData['collectionAddress']
+  name: string
+  description: string
+  image: string
+}
```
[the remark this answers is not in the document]
mvaivre: Yes. On the explorer side, I created a separated type:

```ts
export type NFTFile = {
  name: string
  description: string
  image: string
}
```

src/storage/assets/assetsActions.ts
mvaivre: Yes. We need to investigate if the team pushed for a standardisation (I know that Cheng wanted to). On the explorer, I consider the current [format](https://github.com/alephium/desktop-wallet/pull/725#discussion_r1244780430).

mvaivre: Hongchao just confirmed that they encourage using this format, it's here: https://github.com/alephium/alephium-web3/blob/master/packages/web3/std/nft_interface.ral

mvaivre:
I'm enable to start the wallet. I've tried removing the node_modules, installing packages again etc. but I get:

image

src/types/assets.ts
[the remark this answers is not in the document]
nop33: Yeah, I was inspired by that. I wasn't sure about the "File" as part of the name, so I went for something simpler for now, until the need to be more specific arises. 

In your type, however, every property is mandatory...

nop33:
> I'm enable to start the wallet. I've tried removing the node_modules, installing packages again etc. but I get:

That's weird. It might be because of swhiching from the posthog PR back to this... I'll investigate.

nop33:
Hm, can't reproduce...

src/storage/assets/assetsActions.ts
nop33: Gotcha! I would still make the type properties optional so that we don't get any white screens in case of a rogue NFT.

nop33:
Changes discussed on Slack and here were pushed!

src/types/assets.ts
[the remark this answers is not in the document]
mvaivre: Yeah, I'll make then optional.

mvaivre:
Posthog issue has been solved. Now testing the wallet, I still get this error in the console:

```
react_devtools_backend.js:4061 Error: [API Error] - The number of parameters is different from the number specified by the target method
    at convertHttpResponse (utils.js:26:15)
    at async NodeProvider.fetchFungibleTokenMetaData (node-provider.js:45:28)
    at async assetsActions.ts? [sm]:62:33
```

Do you see it too?

mvaivre:
I'm also witnessing some crazy jumps and infinite loading in the asset list:

https://github.com/alephium/desktop-wallet/assets/6715943/1388e2eb-df2a-4b09-9906-40adccc3925e

nop33:
> I'm also witnessing some crazy jumps and infinite loading in the asset list:

I see what you mean. The code is working as expected, but the effect on the UI side is not nice. Until all calls to the "fetch metadata" endpoints return, the asset lists assume that all tokens (including NFTs) are unknown.

I would like to improve the way we are doing this.

nop33:
@mvaivre I fixed all above issues. Here's what I changed:

- The selector to get address tokens now only returns tokens were metadata is known (either through the token-list repo, or through the node APIs)
- There's a new selector now that selects address tokens that do not have any metadata. This is used to know when to dispatch a sync through the node API for the unknown tokens
- To reduce calls to the node API, I keep track of the unknown tokens that have already been checked.

There is still room for improvement, specifically in the loading indications. Currently, we don't show any loading indication when syncing metadata through the node API. It would improve the UX if we did. But I think it should not be the same loading mechanism as when we load address data, because it does not affect the balances. I am open to ideas on how to do it. Feel free to pull my chances and check it out locally!

mvaivre:
Thanks! I've just tested your changes with the same wallet as before, and unfortunately I got the same issues : 

https://github.com/alephium/desktop-wallet/assets/6715943/d64c1c75-cb87-4481-b6d8-eaf8b5bbefca

nop33:
That's strange, I don't get this infinite loading or any jumps in the asset list... I am using the same wallet.

https://github.com/alephium/desktop-wallet/assets/1579899/341b27a4-d623-487f-9a1c-de2ddda7d0c3

nop33:
Here is a review of the async operations that take place to load all wallet data, and their order.

## Stage 1

### Network ID
To initialize the client, we [fetch the network ID from the node](https://github.com/alephium/desktop-wallet/blob/metadata/src/App.tsx#L128). That's 1 request.

## State 2

### Address data
Once we establish a connection with the network (`state.network.status === 'online'`), we send a storm of API requests to the explorer to fetch data for all our addresses (as long as `state.addresses.status === 'uninitialized'`) ([source](https://github.com/alephium/desktop-wallet/blob/metadata/src/App.tsx#L142-L147)). Particularly, we send a [minimum of 5 requests](https://github.com/alephium/desktop-wallet/blob/metadata/src/api/addresses.ts#L29-L33) per address  + [one request for each token](https://github.com/alephium/desktop-wallet/blob/metadata/src/api/addresses.ts#L35-L41) an address might have. So, for a wallet with 9 addresses (as the wallet we are using above), we have 45 requests + 43 requests to fetch token balances. All these requests are sequential. In parallel, however, there's one request per address to fetch historical balances. That brings us to a total of 97 requests. 

### Token-list
Similarly to above, once we establish a connection with the network (`state.network.status === 'online'`), if we haven't yet initialized our token metadata from the token-list repo (`state.assetsInfo.status === 'uninitialized'`) ([source](https://github.com/alephium/desktop-wallet/blob/metadata/src/App.tsx#L157-L159)), we send a request to GH. I won't count this request cause it doesn't go to our services, but I mention it because it's important for the display of the data.

## Stage 3

### Unknown tokens
Now, once all of the above requests have been fulfilled (after all retries), we know which are the known tokens (the ones that we have metadata for from the token-list repo) and which are the unknown ones, so now we can start querying the node to get metadata for the unknown ones ([source](https://github.com/alephium/desktop-wallet/blob/metadata/src/App.tsx#L163-L170)). We send 1 request for each unknown token to guess its type and another to fetch its metadata. In our wallet, that's 115 requests.

---

It's worth noting that the asset list on the Overview page renders in 3 phases:
1. Phase 1: Showing only ALPH, until "Stage 2" from above is completed
2. Phase 2: Showing ALPH and all known tokens
3. Phase 3: Once "Stage 3" from above is completed, the list updates to display the "newly known" tokens (tokens that became known after querying the node).

All my tests confirm everything I wrote above. The network devtools tab as well. I wonder why there is this crazy jumping and infinite loop in your tests :/ ...

WDYT of the data loading strategy?

src/pages/UnlockedWallet/OverviewPage/AssetsList.tsx
```diff
@@ -151,22 +175,93 @@ const TokensList = ({ className, limit, addressHashes, isExpanded, onExpand }: A
...
       </motion.div>
 
-      {!isExpanded && displayedAssets.length > 3 && onExpand && <ExpandRow onClick={onExpand} />}
+      {!isExpanded && knownFungibleTokens.length > 3 && onExpand && <ExpandRow onClick={onExpand} />}
     </>
   )
 }
 
-const NFTsList = ({ className, isExpanded }: AssetsListProps) => {
+const UnknownTokensList = ({ className, addressHashes, isExpanded, onExpand }: AssetsListProps) => {
```
nop33, line 190: I could probably DRY this one out a bit, together with the above component.

mvaivre:
Awesome work 👏 I haven't encountered any loading issue so far.

Some cosmetic details:

---

image

I would add a loading indicator here.

---

image

Let's display the images either with `object-fit` or `background-size` in `cover` mode to avoid deformations. 

---

If an NFT is selected, we could hide the "amount" field as we know that the value is 1.

image

---

On the explorer I used the hash as main title for unknown tokens because it's the most important info if you don't have a name: 

image

We could do that same in the DW, WDYT?

src/pages/UnlockedWallet/OverviewPage/AssetsList.tsx
nop33: done

mvaivre (approved):
BOOM 💥 
👏 👏 👏

PR #728, opened 2023-06-05
Select last used wallet by default, list wallets using alphabetical order

### LINKED ISSUE #663: Welcome page: select last used wallet by default

mvaivre:
Fixes #663

### DIFF
--- a/src/storage/global/globalSlice.ts
+++ b/src/storage/global/globalSlice.ts
@@ -94,7 +94,9 @@ const globalSlice = createSlice({
       .addCase(languageChangeStarted, (state) => toggleLoading(state, true, true))
       .addCase(languageChangeFinished, (state) => toggleLoading(state, false, true))
       .addCase(walletSaved, (state, action) => {
-        state.wallets.push(action.payload.wallet)
+        const { id, name, encrypted, lastUsed } = action.payload.wallet
+
+        state.wallets.push({ id, name, encrypted, lastUsed })
       })
       .addCase(walletLocked, resetState)
       .addCase(activeWalletDeleted, resetState)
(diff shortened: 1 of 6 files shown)


src/storage/global/globalSlice.ts
```diff
@@ -94,9 +94,7 @@ const globalSlice = createSlice({
       .addCase(languageChangeStarted, (state) => toggleLoading(state, true, true))
       .addCase(languageChangeFinished, (state) => toggleLoading(state, false, true))
       .addCase(walletSaved, (state, action) => {
-        const { id, name, encrypted } = action.payload.wallet
-
-        state.wallets.push({ id, name, encrypted })
+        state.wallets.push(action.payload.wallet)
```
nop33, line 97: Before this change, the `wallets` state was an array of wallet objects like this:
```ts
{
  id: string
  name: string
  encrypted: string
}
```

With the change you made, we are storing extra unnecessary data in this array:

```
{
  id: string
  name: string
  encrypted: string

  lastUse: Date

  mnemonic: string
  passphrase?: string
}
```

I think we should simply extract the `lastUse` one and not store the rest (mnemonic, passphrase).

src/types/wallet.ts
```diff
@@ -41,6 +41,7 @@ export type StoredWallet = {
   id: ActiveWallet['id']
   name: ActiveWallet['name']
   encrypted: string
+  lastUse: Date
```
nop33, line 44: It is recommended to store dates in Redux as strings or numbers (timestamps): https://github.com/reduxjs/redux-toolkit/issues/456

See for example how we handle the `lastUsed` prop of the `AddressState` type: https://github.com/alephium/desktop-wallet/blob/next/src/types/addresses.ts#L60

nop33 (approved):
👏

PR #741, opened 2023-06-27
Capture errors in PostHog

nop33:
@mvaivre I purposefully did not capture the actual error message in the spirit of preserving user privacy. We don't know what info the error messages might include. Some could be PII. To respect our decision to not store any user data, I simply capture generic errors that will give us a hint as to where there might be a problem so that we can identify it outselves.

I did a global search for `catch (e) {` and capture an error.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -82,9 +82,10 @@ const App = () => {
       dispatch(localStorageDataMigrated())
     } catch (e) {
       console.error(e)
+      posthog.capture('Error - Local storage data migration failed')
       dispatch(localStorageDataMigrationFailed())
     }
-  }, [dispatch])
+  }, [dispatch, posthog])
 
   useEffect(() => {
     posthog.people.set({
@@ -130,6 +131,7 @@ const App = () => {
       // TODO: Check if connection to explorer also works
       dispatch(apiClientInitSucceeded({ networkId, networkName: network.name }))
     } catch (e) {
+      // Discuss: Do we want to capture client init errors?
       dispatch(apiClientInitFailed({ networkName: network.name, networkStatus: network.status }))
     }
   }, [network.settings.nodeHost, network.settings.explorerApiHost, network.name, network.status, dispatch])
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -82,7 +82,7 @@ const App = () => {
       dispatch(localStorageDataMigrated())
     } catch (e) {
       console.error(e)
-      posthog.capture('Error - Local storage data migration failed')
+      posthog.capture('Error', { message: 'Local storage data migration failed' })
       dispatch(localStorageDataMigrationFailed())
     }
   }, [dispatch, posthog])
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -87,7 +87,7 @@ const App = () => {
   }, [dispatch])
 
   useEffect(() => {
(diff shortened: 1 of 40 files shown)


src/components/PasswordConfirmation.tsx
```diff
@@ -61,6 +61,7 @@ const PasswordConfirmation: FC<PasswordConfirmationProps> = ({
         onCorrectPasswordEntered(password)
       }
     } catch (e) {
+      // Discuss: Do we want to capture password validation errors?
```
nop33, line 64: I would again say no.

src/contexts/global.tsx
```diff
@@ -90,6 +90,7 @@ export const GlobalContextProvider: FC<{ overrideContextValue?: PartialDeep<Glob
       wallet = WalletStorage.load(walletId, password)
     } catch (e) {
       console.error(e)
+      // Discuss: Do we want to capture password validation errors?
```
nop33, line 93: Same here

src/contexts/walletconnect.tsx
```diff
@@ -313,10 +317,13 @@ export const WalletConnectContextProvider: FC = ({ children }) => {
 
         if (!errorMessage.includes('Pairing already exists')) {
           dispatch(walletConnectPairingFailed(errorMessage))
+          posthog.capture('Error - Could not pair with WalletConnect')
+        } else {
+          posthog.capture('Error - Could not pair with WalletConnect: Pairing already exists')
```
nop33, line 322: An issue that I see with this is that we'll spam PostHog a bit... I am not sure if it's our implementation problem or a WC problem, but I very often get these errors in the console, without affecting the normal usage of the wallet.

src/App.tsx
```diff
@@ -129,6 +131,7 @@ const App = () => {
       // TODO: Check if connection to explorer also works
       dispatch(apiClientInitSucceeded({ networkId, networkName: network.name }))
     } catch (e) {
+      // Discuss: Do we want to capture client init errors?
```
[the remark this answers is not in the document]
mvaivre: Agreed.

src/contexts/walletconnect.tsx
mvaivre: So let's start without it. As a first step, let's start with the number of successful WalletConnect sessions only, very interesting metric.

src/components/Buttons/ShortcutButtons.tsx
```diff
@@ -73,31 +73,31 @@ const ShortcutButtons = ({
   const lockWallet = () => {
     dispatch(walletLocked())
 
-    posthog?.capture('Locked wallet', { origin: analyticsOrigin })
+    posthog.capture('Locked wallet', { origin: analyticsOrigin })
```
mvaivre, line 76: Need to check, but isn't there another "Locked wallet" event somewhere, not necessarily attached to the shortcut button? 

What would be interesting here I believe, is to identify which shortcut button is clicked. Ie. "is it the one on the home page"? 

Same for the events below. WDYT?

mvaivre (approved):
Great job!

src/components/Buttons/ShortcutButtons.tsx
nop33: This is already the case :) 

image

PR #755, opened 2023-07-11
Add More Fiat Currency options on Alephium Wallet 

ykris45:
Hello, i add three fiat currencies in to Alph Wallet.

1. Turkiye Lira
2. Russian Ruble
3. Vietnamese Dong
Thanks You So Much!

### DIFF
--- a/src/types/settings.d.ts
+++ b/src/types/settings.d.ts
@@ -53,7 +53,7 @@ export type ThemeType = 'light' | 'dark'
 
 export type ThemeSettings = ThemeType | 'system'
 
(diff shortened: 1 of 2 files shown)


nop33:
Hm, what about only adding currencies for countries that we support the language? Does that make sense?

ykris45:
> Hm, what about only adding currencies for countries that we support the language? Does that make sense?

yes it makes sense. did I re-edit and only added 3 currencies

ykris45:
@nop33 can you help to fix my unit test errors. i formated with prettier but its doesnt works.

ykris45:
@nop33 done, 3 currencies

nop33 (approved):
Thanks for adding these 3! FYI, to enable then, you also need to add them in this file: https://github.com/alephium/desktop-wallet/blob/master/src/utils/settings.ts#L45-L50

I'll take care of that for now! Cheers!

PR #766, opened 2023-07-19
Fix linux

nop33:


src/App.tsx
```diff
@@ -104,17 +104,29 @@ const App = () => {
...
+    network.name,
+    posthog.__loaded,
+    posthog.people,
+    settings.devTools,
+    settings.fiatCurrency,
+    settings.language,
+    settings.passwordRequirement,
+    settings.theme,
+    settings.walletLockTimeInMinutes,
+    wallets.length
```
mvaivre, line 128: Hahaha when the deps array is longer that the actual useEffect 😆

src/contexts/analytics.tsx
```diff
@@ -23,7 +23,9 @@ import AnalyticsStorage from '@/storage/analytics/analyticsPersistentStorage'
 import SettingsStorage from '@/storage/settings/settingsPersistentStorage'
 import { GeneralSettings } from '@/types/settings'
 
-const { VITE_PUBLIC_POSTHOG_KEY, VITE_PUBLIC_POSTHOG_HOST, VITE_VERSION } = import.meta.env
+const VITE_PUBLIC_POSTHOG_KEY = 'phc_FLKGQDmMQSdSb3qjaTwHWwm9plmz7couyVJFG9GOMr7'
+const VITE_PUBLIC_POSTHOG_HOST = 'https://eu.posthog.com'
+const VITE_VERSION = import.meta.env
```
mvaivre, line 28: 1) If we don't use env variables, let's simplify the names by removing the VITE_ prefix?
2) Line 28 looks fishy

nop33: you're totally right

nop33: fixed

mvaivre (approved):
LGTM, well done

Issue #770, opened 2023-07-26
Show the transaction hash in the transaction modal


mvaivre:
According to a user, some services ask for a screenshot as confirmation, and the hash should be visible. 

> Simple swap and Stealth EX always want to have the hash number when their end of the exchange stops working. they always ask for a un cropped screenshot of the wallet app with the hash number to verify and they don't believe it when it takes you to the explorer.

Issue #761, opened 2023-07-14
Title is not aligned in Vietnamese


nit27:
## **Describe the bug**
At the welcome page, the title is not aligned: **Chào mừng bạn quay trở lại.**

## **To Reproduce**
1. Open Alephium.app
2. See error at *welcome page*

## **Expected behavior**
Decrease the font size of this text "**Chào mừng bạn quay trở lại.**"  in order to make them in a single line.

## **Screenshots**
<img width="1312" alt="image" src="https://github.com/alephium/desktop-wallet/assets/95131051/fb5f3cea-11e0-456b-8c10-3d75a019316a">

compare to English version:
<img width="1312" alt="image" src="https://github.com/alephium/desktop-wallet/assets/95131051/11b8a045-59b1-4e01-ba50-d9cc90fdaff5">

## **Environment**
 - OS: MacOS
 - Version: 13.4.1 Ventura

PR #772, opened 2023-08-03
2.1.3 - part 1

### LINKED ISSUE #771: Update default localhost settings to match those of alephium-stack
Node: http://127.0.0.1:22973
Explorer: http://127.0.0.1:9090
Frontend: http://localhost:23000

mvaivre:
Fixes #771, #761, #770


nop33 (approved):
One small comment: I would avoid creating branches named after the version cause it might be confusing when the time for releasing comes (the tag will have the same name).

mvaivre:
> the tag will have the same name

PR #777, opened 2023-08-09
Improve loading experience (like, a lot!)

nop33:


src/api/fetchRetry.ts
nop33, line 1: I had to move this to a new file because of the dependency in a web worker

src/storage/addresses/addressesSlice.ts
```diff
@@ -140,9 +147,7 @@ const addressesSlice = createSlice({
           return {
             id: hash,
             changes: {
-              ...details,
-              tokens,
-              transactions: uniq([...address.transactions, ...transactionHashes]),
+              transactions: uniq(address.transactions.concat(transactions.map((tx) => tx.hash))),
```
mvaivre, line 150: Nit: we usually use spreading to do this, which is easier to read IMO. 

```ts
uniq([...address.transactions, ...map(transactions, 'hash')])
``` 

or 
```ts
uniq([...address.transactions, ...transactions.map((tx) => tx.hash)])
``` 

(because I'm sad about the typing of lodash's property shorthand `map` which isn't strict...)

mvaivre:
Incredible jump in performances. This feels extremely good.

src/storage/addresses/addressesSlice.ts
nop33: Fixed!

package.json
```diff
@@ -33,10 +33,10 @@
     "electron-updater": "^5.3.0"
   },
   "devDependencies": {
-    "@alephium/sdk": "0.7.2",
-    "@alephium/token-list": "^0.0.8",
+    "@alephium/sdk": "0.7.3",
```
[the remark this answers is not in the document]
nop33: done

mvaivre (approved):

PR #778, opened 2023-08-14
Use latest price as last data point in chart

### LINKED ISSUE #774: Total wallet worth growth percentage is misleading compared to chart
This is from our latest release 2.1.2. There seems to be a problem with the chart or the percentage calculation. It should be green since the current worth is higher than the past one.

image

nop33:
Closes #774

The problem was that the last item of the response of the `getHistoricalPrice` endpoint was slightly different than the latest price from the `getPrice` endpoint. The fix removes today's price received from the 1st endpoint and adds the one from the 2nd.

### DIFF
--- a/src/storage/assets/priceApiSlice.ts
+++ b/src/storage/assets/priceApiSlice.ts
@@ -51,9 +51,18 @@ export const priceApi = createApi({
         const { prices } = response
         const today = dayjs().format(CHART_DATE_FORMAT)
 
-        return prices
-          .filter(([date]) => dayjs(date).format(CHART_DATE_FORMAT) !== today)
-          .map(([date, price]) => ({ date: dayjs(date).format(CHART_DATE_FORMAT), price }))
+        return prices.reduce((acc, [date, price]) => {
+          const itemDate = dayjs(date).format(CHART_DATE_FORMAT)
+          const isDuplicatedItem = !!acc.find(({ date }) => dayjs(date).format(CHART_DATE_FORMAT) === itemDate)
+
+          if (!isDuplicatedItem && itemDate !== today)
+            acc.push({
+              date: itemDate,
+              price
+            })
+
+          return acc
+        }, [] as HistoricalPriceResult[])
       }
     })
   })
--- a/src/storage/assets/priceApiSlice.ts
+++ b/src/storage/assets/priceApiSlice.ts
@@ -47,8 +47,14 @@ export const priceApi = createApi({
     }),
     getHistoricalPrice: builder.query<HistoricalPriceResult[], HistoricalPriceQueryParams>({
       query: ({ currency, days }) => `/coins/alephium/market_chart?vs_currency=${currency.toLowerCase()}&days=${days}`,
-      transformResponse: (response: { prices: number[][] }) =>
-        response.prices.map((p) => ({ date: dayjs(p[0]).format(CHART_DATE_FORMAT), price: p[1] }))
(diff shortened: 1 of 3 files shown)


mvaivre:
Thanks!

I'm noticing some artefacts in the graph, as if the interpolation between the last 2 points is broken:

image

src/storage/assets/priceApiSlice.ts
```diff
@@ -47,8 +47,14 @@ export const priceApi = createApi({
     }),
     getHistoricalPrice: builder.query<HistoricalPriceResult[], HistoricalPriceQueryParams>({
       query: ({ currency, days }) => `/coins/alephium/market_chart?vs_currency=${currency.toLowerCase()}&days=${days}`,
-      transformResponse: (response: { prices: number[][] }) =>
-        response.prices.map((p) => ({ date: dayjs(p[0]).format(CHART_DATE_FORMAT), price: p[1] }))
+      transformResponse: (response: { prices: number[][] }) => {
+        const { prices } = response
```
mvaivre, line 51: The very last fetched value has a "lag", which means that  the last 2 `dates` can actually be the same day.

We should probably systematically remove the last value here, and push the latest price in the chart as you've done it.

src/storage/assets/priceApiSlice.ts
```diff
@@ -47,8 +47,14 @@ export const priceApi = createApi({
...
+        const today = dayjs().format(CHART_DATE_FORMAT)
+
+        return prices
+          .filter(([date]) => dayjs(date).format(CHART_DATE_FORMAT) !== today)
```
mvaivre, line 55: Let's just `pop()`

src/storage/assets/priceApiSlice.ts
nop33: > We should systematically remove the last value here: it makes sense for historical data to stop the day before today IMO. In the chart, we'll add the latest price as you've done in this PR.

That's exactly what's happening here.

What I implemented (and what we actually want to achieve) is to replace the last 2 elements (that seem to always be today's date) with the actual latest price. Since I can't find any documentation that says that the last 2 elements are always going to end up being today's date, I used `filter` instead of `pop`. In the chart component, I then `push` the today's date with the latest price.

src/storage/assets/priceApiSlice.ts
nop33: I think `filter` is a safer implementation that `pop` x2. WDYT?

nop33:
> I'm noticing some artefacts in the graph, as if the interpolation between the last 2 points was broken.
> It is caused by the fact that there can be 2 times the same date (day before today) overlayed in the chart.

Hm, strange, I'll test with some other wallets, but in all my tests I ensured that each day is unique in the set of data...

src/storage/assets/priceApiSlice.ts
mvaivre: The last 2 items were yesterday's date during my test, hence the issue I witnessed.

nop33: You tested today and the last 2 elements were yesterday's date?! Interesting case!

nop33:
@mvaivre I've improved the algorithm that cleans the last chart data by ensuring there are no duplicates and that the "today"'s date is always skipped so that we can manually add the latest price.

mvaivre (approved):
Thanks for the fix, LGTM.

PR #787, opened 2023-08-18
Fix tokens loading

nop33:
The bug was that we were not resetting `page` and `pageTotalResults` after checking the first address, meaning that we would only display the token balances of the first address.


mvaivre (approved):
😅

PR #794, opened 2023-09-12
Walletconnect: auto connect

Lbqds:
The branch for the @alephium/web3 is here: https://github.com/alephium/alephium-web3/pull/270

When testing with `npm run start:electron`, please be mindful **NOT** to use CTRL + C to shutdown the client. Electron [won't persist localStorage in this case](https://github.com/electron/electron/issues/1731#issuecomment-221922779). And I tried `process.on('exit', () => app.quit())`, but it still does not work.

### DIFF
--- a/src/contexts/walletconnect.tsx
+++ b/src/contexts/walletconnect.tsx
@@ -29,7 +29,7 @@ import {
 import { node } from '@alephium/web3'
 import SignClient from '@walletconnect/sign-client'
 import { EngineTypes, SignClientTypes } from '@walletconnect/types'
-import { getSdkError } from '@walletconnect/utils'
+import { getChainsFromNamespaces, getSdkError } from '@walletconnect/utils'
 import { partition } from 'lodash'
 import { usePostHog } from 'posthog-js/react'
 import { createContext, useCallback, useContext, useEffect, useState } from 'react'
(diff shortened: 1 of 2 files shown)


src/contexts/walletconnect.tsx
```diff
@@ -115,6 +115,32 @@ export const WalletConnectContextProvider: FC = ({ children }) => {
...
+  const resetSession = useCallback(() => {
+    setRequiredChainInfo(undefined)
+    setProposalEvent(undefined)
+    setWcSessionState('uninitialized')
+    setSessionTopic(undefined)
+  }, [])
+
+  const tryRestoreLastSession = useCallback(
+    async (client: SignClient) => {
+      if (!client.session.length) return resetSession()
```
nop33, line 127: Note to myself: When checking for active sessions, we should probably first cross-reference them with the active pairings. See: https://github.com/alephium/mobile-wallet/blob/master/src/utils/walletConnect.ts#L50-L58

nop33:
Thanks for this PR @Lbqds!

As per request from the community I've implemented in my PR #802 the feature of handling multiple WalletConnect sessions. This includes initializing the app state with the current active WalletConnect sessions, derived from the client. I believe this covers the use case of this PR as well, so I believe it's not needed anymore.

Let me know if you think differently!

Lbqds:
Great work👍, feel free to close it if it's not needed!

nop33:
I think I'll keep it open until the new desktop wallet version is released, then we can test together to verify that the requirements of this PR are met 😉

PR #802, opened 2023-10-23
Manage multiple WalletConnect sessions

nop33:


src/contexts/walletconnect.tsx
```diff
@@ -52,43 +55,27 @@ import {
...
   dappTxData?: DappTxData
-  onSessionRequestError: (event: RequestEvent, error: ReturnType<typeof getSdkError>) => Promise<void>
-  onSessionRequestSuccess: (event: RequestEvent, result: node.SignResult) => Promise<void>
-  onSessionDelete: () => void
-  connectToWalletConnect: (uri: string) => void
-  requiredChainInfo?: ChainInfo
-  wcSessionState: WalletConnectSessionState
-  sessionTopic?: string
-  onProposalApprove: (topic: string) => void
-  connectedDAppMetadata?: ProposalEvent['params']['proposer']['metadata']
```
nop33, line 76: As seen here, the WC context is now much more minimal than before, similar to the mobile wallet.

src/modals/SendModals/SendModal.tsx
```diff
@@ -147,24 +152,25 @@ function SendModal<PT extends { fromAddress: Address }, T extends PT>({
         }
       } catch (e) {
         // TODO: When API error codes are available, replace this substring check with a proper error code check
-        const { error } = e as APIError
-        if (error?.detail && (error.detail.includes('consolidating') || error.detail.includes('consolidate'))) {
```
nop33, line 151: Seems like something changed in the API and the errors are not returned in this structure anymore.

nop33:
ping @mvaivre

src/modals/SendModals/SendModal.tsx
mvaivre: Has this be discussed with BE team? cc @polarker

src/contexts/walletconnect.tsx
mvaivre: Much better!

src/modals/SendModals/SendModal.tsx
mvaivre: After a quick sync with BE, it appears that the idea is still to improve the error system and introduce error code to the endpoints. Once that’s done, we will be able to handle error based on error code.

In the meantime, let's try to keep the error messages rather static. @h0ngcha0 do you know if the new structure is used everywhere? Any other error messages have been changed? 

If you plan any other change in the future, thanks in advance for letting us know a little bit in advance so that we can update the way we handle errors 🙏 :)

src/contexts/walletconnect.tsx
```diff
@@ -305,126 +295,424 @@ export const WalletConnectContextProvider: FC = ({ children }) => {
...
-        return await walletConnectClient.pair({ uri })
+        const pairings = walletConnectClient.core.pairing.pairings
+        const existingPairing = pairings.values.find(({ topic }) => topic === pairingTopic)
+
+        if (existingPairing) {
+          console.log('⏳ TRYING TO CONNECT WITH EXISTING PAIRING:', pairingTopic)
+          if (!existingPairing.active) {
+            console.log('⏳ EXISTING PAIRING IS INACTIVE, ACTIVATING IT...')
+            // `activate` doesn't trigger the onSessionProposal as the `pair` does (even if we call `pair` or `connect`)
+            // despite what the docs say (https://specs.walletconnect.com/2.0/specs/clients/sign/session-events#session_proposal)
```
mvaivre, line 328: 😓

src/contexts/walletconnect.tsx
mvaivre: Incredible work on this file. It must have required a ton of patience 👏

src/index.tsx
```diff
@@ -37,6 +37,7 @@ import { store } from '@/storage/store'
 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 // const root = createRoot(document.getElementById('root')!)
 
+// eslint-disable-next-line react/no-deprecated
```
mvaivre, line 40: What is preventing us from using the non-deprecated way of rendering the dom?

src/modals/SendModals/SendModal.tsx
mvaivre: @nop33 Thought: once stabilised, the error mgmt logic could be extracted to the future shared library :)

src/modals/WalletConnect/WalletConnectSessionProposalModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  const generateAddressInGroup = () => {
+    const address = generateAddress({ group })
+    saveNewAddresses([{ ...address, isDefault: false, color: getRandomLabelColor() }])
+
+    posthog.capture('New address created through WalletConnect modal')
+  }
+
+  const rejectAndCloseModal = async () => {
```
mvaivre, line 90: This method is async but nothing inside is awaited. Should `rejectProposal` be awaited?

src/modals/WalletConnect/WalletConnectSessionProposalModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const showNetworkWarning =
+    requiredChainInfo?.networkId && !isNetworkValid(requiredChainInfo.networkId, currentNetworkId)
+
+  useEffect(() => {
+    setSignerAddress(
+      addressesInGroup.length > 0 ? addressesInGroup.find((a) => a.isDefault) ?? addressesInGroup[0] : undefined
+    )
+  }, [addressesInGroup])
+
+  const handleSwitchNetworkPress = async () => {
```
mvaivre, line 77: This method is async but nothing inside is awaited

mvaivre (commented):
Part 1 👏

src/index.tsx
nop33: The scrollbar lib :)

image

src/modals/WalletConnect/WalletConnectSessionProposalModal.tsx
nop33: I'll address this as a commit in the monorepo, if you don't mind!

src/modals/WalletConnect/WalletConnectSessionProposalModal.tsx
nop33: Same here 😇

src/index.tsx
mvaivre: arhhhhaarrrhrhhhrrr right.

src/modals/WalletConnect/WalletConnectSessionProposalModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+    <CenteredModal
+      title={
+        metadata.description ? t('Connect to {{ dAppUrl }}', { dAppUrl: cleanUrl(metadata.url) }) : t('Connect to dApp')
+      }
+      subtitle={metadata.description || metadata.url}
+      onClose={rejectAndCloseModal}
+      Icon={() =>
+        metadata?.icons &&
+        metadata.icons.length > 0 &&
+        metadata.icons[0] && <AssetLogo assetImageUrl={metadata.icons[0]} assetId={metadata.url} size={50} />
```
mvaivre, line 107: Nit: If `metadata.icons.length > 0` is `true` then `metadata.icons[0]` must exist?

src/modals/WalletConnect/WalletConnectSessionProposalModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+                values={{
+                  dAppUrl: cleanUrl(metadata.url),
+                  group
+                }}
+                components={{
+                  1: <Highlight />,
+                  3: <Highlight />
+                }}
+              >
+                {'Connect to <1>{{ dAppUrl }}</1> with one of your addresses in group <3>{{ group }}</3>:'}
```
mvaivre, line 169: Question: can't the same component (1) be reused in 2 spots?

src/utils/misc.ts
```diff
@@ -85,3 +85,5 @@ export function removeItemFromArray<T>(array: T[], index: number) {
   newArray.splice(index, 1)
   return newArray
 }
+
+export const cleanUrl = (url: string) => url.replace('https://', '')
```
mvaivre, line 89: Nit: maybe the name could be a bit more explicit? Like `removeUrlScheme`? Or `removeUrlHttps` :P 

image

src/utils/walletConnect.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { parseChain, PROVIDER_NAMESPACE } from '@alephium/walletconnect-provider'
+import { SignClient } from '@walletconnect/sign-client/dist/types/client'
+
+import { networkPresets } from '@/storage/settings/settingsPersistentStorage'
+import { NetworkPreset } from '@/types/network'
+import { NetworkSettings } from '@/types/settings'
+import { SessionProposalEvent } from '@/types/walletConnect'
+
+export const isNetworkValid = (networkId: string, currentNetworkId: NetworkSettings['networkId']) =>
```
mvaivre, line 27: These kind of methods could be ported to the `shared` lib and also tested there. WDYT?

Just saying that out loud so that we identify the parts we could share between projects ;)

src/modals/WalletConnect/WalletConnectSessionProposalModal.tsx
nop33: But there is no guarantee that it's not an empty string...

src/modals/WalletConnect/WalletConnectSessionProposalModal.tsx
nop33: Hm, good question, haven't tried it.

src/utils/misc.ts
nop33: I guess we could make the name more explicit, but I like that the current name is small. No objection though.

src/utils/walletConnect.ts
nop33: Absolutely man.

src/modals/WalletConnect/WalletConnectSessionProposalModal.tsx
mvaivre: Yes but if it's an empty string, then AssetLogo would show a "?" which would be ok?

It's a nit anyway.

src/utils/misc.ts
mvaivre: I'm just expecting a more profound cleaning when I see a name like that :D But no objection either.

nop33:
The code changes of this PR have already been integrated into the monorepo. So I propose we merge it, for archive purposes. Do you agree @mvaivre ?

